PR #71, opened 2020-07-02
Update to Prettier 2.0

southpolesteve:
List of breaking changes in Prettier 2.0: https://prettier.io/blog/2020/03/21/2.0.0.html#breaking-changes

Trailing commas and parens around arrow function arguments seem to be the major changes affecting us


southpolesteve:
Will tackle this some other day. Not urgent.

PR #81, opened 2020-07-09
Make explicit any a linting error

southpolesteve:
Explicit `any` types is currently a warning, but this turns it into an error. We should avoid using any. Here are some good primers:

https://itnext.io/avoiding-any-in-typescript-advanced-types-and-their-usage-691b02ac345a
https://43081j.com/2019/02/typescript-avoid-using-any

In many cases where `any` is required,  `unknown` is a better type. If `any` is truly needed, we can add an eslint line exception with a comment justifying the exception.

### DIFF
--- a/src/Common/MongoProxyClient.ts
+++ b/src/Common/MongoProxyClient.ts
@@ -16,6 +16,7 @@ import { MessageHandler } from "./MessageHandler";
 import { MessageTypes } from "../Contracts/ExplorerContracts";
 import { NotificationConsoleUtils } from "../Utils/NotificationConsoleUtils";
 import { ResourceProviderClient } from "../ResourceProvider/ResourceProviderClient";
+import { MinimalQueryIterator } from "./IteratorUtilities";
 
 const defaultHeaders = {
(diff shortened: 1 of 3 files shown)


src/Common/MongoProxyClient.ts
```diff
@@ -448,12 +452,13 @@ export async function _createMongoCollectionWithARM(
   }
 
   try {
-    await new ResourceProviderClient(armEndpoint).putAsync(
+    return new ResourceProviderClient<DataModels.CreateCollectionWithRpResponse>(armEndpoint).putAsync(
```
southpolesteve, line 455: @shmotlan This seems like a bug? The upstream code here expects a return value, but it was returning undefined.

shmotlan: yes it seems like a bug as its not returning anything. do you still need to keep the `await` before `new ResourceProviderClient`?

southpolesteve: No. You can omit `await` when returning a promise.

shmotlan: ok lgtm

PR #82, opened 2020-07-09
Add hideInputs query param in notebookViewer and other minor updates

gitanuj:
- `hideInputs` hides input cells when rendering notebook
- Use fav icons for `galleryViewer` and `notebookViewer`
- Use notebook title as document title in `notebookViewer`

### DIFF
--- a/src/NotebookViewer/NotebookViewer.tsx
+++ b/src/NotebookViewer/NotebookViewer.tsx
@@ -17,24 +17,26 @@ const onInit = async () => {
   const galleryViewerProps = GalleryUtils.getGalleryViewerProps(window.location.search);
   const notebookViewerProps = GalleryUtils.getNotebookViewerProps(window.location.search);
   const backNavigationText = galleryViewerProps.selectedTab && GalleryUtils.getTabTitle(galleryViewerProps.selectedTab);
+  const hideInputs = notebookViewerProps.hideInputs;
 
   const notebookUrl = decodeURIComponent(notebookViewerProps.notebookUrl);
-  render(notebookUrl, backNavigationText);
+  render(notebookUrl, backNavigationText, hideInputs);
 
   const galleryItemId = notebookViewerProps.galleryItemId;
   if (galleryItemId) {
     const junoClient = new JunoClient();
     const notebook = await junoClient.getNotebook(galleryItemId);
-    render(notebookUrl, backNavigationText, notebook.data);
+    render(notebookUrl, backNavigationText, hideInputs, notebook.data);
   }
 };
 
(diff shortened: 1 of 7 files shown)


src/NotebookViewer/NotebookViewer.tsx
```diff
@@ -10,35 +10,42 @@ import {
...
   const notebookUrl = decodeURIComponent(notebookViewerProps.notebookUrl);
-  render(notebookUrl, backNavigationText);
+  render(notebookUrl, backNavigationText, hideInputs);
 
   const galleryItemId = notebookViewerProps.galleryItemId;
   if (galleryItemId) {
     const junoClient = new JunoClient();
     const notebook = await junoClient.getNotebook(galleryItemId);
-    render(notebookUrl, backNavigationText, notebook.data);
+    render(notebookUrl, backNavigationText, hideInputs, notebook.data);
```
languy, line 30: nit: this could be a 2nd call to `render()`? Isn't it better to put the first `render()` inside an `else` of this case (iow if `galleryItemId` is `undefined`)?

src/NotebookViewer/NotebookViewer.tsx
```diff
@@ -10,35 +10,42 @@ import {
...
   }
 };
 
-const render = (notebookUrl: string, backNavigationText: string, galleryItem?: IGalleryItem) => {
+const render = (notebookUrl: string, backNavigationText: string, hideInputs: boolean, galleryItem?: IGalleryItem) => {
```
languy, line 34: nit: personally, I would make the `hideInputs` optional. We only need it in some specific case, but normal behavior is off.

src/NotebookViewer/NotebookViewer.tsx
gitanuj: Yes this would be a 2nd call to render. Since React anyway takes care of diff-ing the DOM I thought it's better to call render as soon as we can. In our case first render can be done immediately after the page loads, but the second render needs to wait for `junoClient.getNotebook` to finish. This makes the loading experience more snappy.

src/NotebookViewer/NotebookViewer.tsx
gitanuj: True. `hideInputs` is already optional for `NotebookViewerComponentProps`. For this particular scenario though we're always getting `hideInputs` from query params using `GalleryUtils.getNotebookViewerProps`

src/NotebookViewer/NotebookViewer.tsx
languy: Good point. I agree!

PR #85, opened 2020-07-11
Fix indexing off not working for SQL and graph free tier accounts

### LINKED ISSUE #74: Indexing "OFF" not respected for AAD logged in user with free tier account
From @languy: It appears that the indexing setting is (incorrectly) ignored for Graph and SQL API when the user is signed in with AAD. Under these conditions, the default settings are being used (automatic indexing), so choosing indexing OFF will produce the same result as indexing set to Automatic.

victor-meng:
If the user is logged in with AAD, the indexing policy that the user selects is not used when creating a new collection/graph with resource provider (github issue: https://github.com/Azure/cosmos-explorer/issues/74).

The fix is to pass the indexing policy down to the create collection/graph call in RP. Tested the fix using a free tier account and confirmed that the fix worked for both SQL and graph.

Closes #74


southpolesteve:
FYI If you put `closed #issuenum` in the body of the PR it will close the linked issue automatically when this merges https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #90, opened 2020-07-15
Added infinite progress bar when gallery read-only notebook viewer loads

srinathnarayanan:
Reused the infinite progress bar from the magic commands of Notebooks. It is displayed when notebook content has not been loaded yet.
gallery progress bar gif


src/Explorer/Controls/InfiniteProgressBar/InfiniteProgressBarComponent.less
```diff
@@ -0,0 +1,57 @@
+.progress {
```
languy, line 1: This is great, but can't we use this?
https://developer.microsoft.com/en-us/fluentui#/controls/web/progressindicator

srinathnarayanan: The progressindicator looks better! added that instead

PR #93, opened 2020-07-15
Add more files to strict mode

southpolesteve:


src/Explorer/Controls/ErrorDisplayComponent/ErrorDisplayComponent.ts
```diff
@@ -23,8 +23,5 @@ interface ErrorDisplayParams {
 }
 
 class ErrorDisplayViewModel {
-  private params: ErrorDisplayParams;
-  public constructor(params: ErrorDisplayParams) {
-    this.params = params;
-  }
+  public constructor(public params: ErrorDisplayParams) {}
```
gitanuj, line 26: should this be private?

southpolesteve: yep!

southpolesteve: Actually I lied. This is bug in the current implementation. Knockout templates don't know the difference between public and private and private isn't a real thing in vanilla JS

PR #105, opened 2020-07-21
Support cosmos readers in Cosmos Explorer

vignesh-mr:
Cosmos DB readers cannot use the same API that is used by contributors to fetch Cosmos DB keys, so this change falls back to this other API to fetch read only keys for readers.

### DIFF
--- a/src/Common/Constants.ts
+++ b/src/Common/Constants.ts
@@ -351,6 +351,7 @@ export class HttpStatusCodes {
   public static readonly Created: number = 201;
   public static readonly Accepted: number = 202;
   public static readonly NoContent: number = 204;
+  public static readonly NotModified: number = 304;
   public static readonly Unauthorized: number = 401;
   public static readonly Forbidden: number = 403;
   public static readonly NotFound: number = 404;
--- a/src/Platform/Hosted/Main.ts
+++ b/src/Platform/Hosted/Main.ts
@@ -568,8 +568,9 @@ export default class Main {
 
     this._explorer.hideConnectExplorerForm();
 
+    const masterKey = Main._getMasterKey(keys);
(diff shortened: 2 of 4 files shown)


src/Common/Constants.ts
```diff
@@ -351,6 +351,7 @@ export class HttpStatusCodes {
   public static readonly Created: number = 201;
   public static readonly Accepted: number = 202;
   public static readonly NoContent: number = 204;
+  public static readonly NotModified: number = 304;
```
languy, line 354: It feels like we should use some package for these definitions. Like [this](https://www.npmjs.com/package/http-status-codes)...

src/Platform/Hosted/Main.ts
```diff
@@ -568,21 +568,31 @@ export default class Main {
...
+    const masterKey = Main._getMasterKey(response[1]);
     const authorizationToken = response[2];
-    Main._setExplorerReady(this._explorer, keys.primaryMasterKey, account, authorizationToken);
+    Main._setExplorerReady(this._explorer, masterKey, account, authorizationToken);
     this._getAadAccessDeferred.resolve(this._explorer);
   }
 
+  private static _getMasterKey(keys: AccountKeys): string {
+    return (
+      keys?.primaryMasterKey ??
```
languy, line 589: wow, I never I'd marvel to see optional chaining and nullish coalescing in a single statement one day... :)

vignesh-mr: That's what I love about TS! :D

src/Common/Constants.ts
vignesh-mr: Yep, that'd definitely help given the comprehensive list. I'd do a separate pass just for that cause it looks like we're using these Constants everywhere.

PR #108, opened 2020-07-22
Added support for taking screenshot during Notebook publish to Gallery

srinathnarayanan:
- Options now available include Screenshot and using first display output for cover Image of published Notebook
- Screenshot is taken using html2canvas package
- Converted to base 64 and uploaded to metadata
- Actions for referencing the NotebookRenderer div from redux store were added
- For Using first display output
  - Notebook object is passed instead of string, to publish pane
  - The first cell with output present is found from Notebook object
  - The dom is also parsed to get corresponding div element to take screenshot of the first output

screenshot


src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -111,9 +187,79 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
+        return <TextField {...this.thumbnailUrlProps} />;
+      case PublishNotebookPaneComponent.CustomImage:
+        return (
+          <input
+            id="selectImageFile"
+            type="file"
+            accept="image/*"
+            onChange={event => {
+              const file = event.target.files[0];
+              if (file.size / 1024 ** 2 > PublishNotebookPaneComponent.maxImageSizeInMib) {
```
languy, line 202: is the double star `**` intended?

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -111,9 +187,79 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
+
+  private findFirstOutput(): HTMLElement {
+    const indexOfFirstCodeCellWithDisplay = this.findFirstCodeCellWithDisplay();
+    const cellOutputDomElements = this.props.notebookParentDomElement.querySelectorAll<HTMLElement>(
+      ".nteract-cell-outputs"
+    );
+    return cellOutputDomElements[indexOfFirstCodeCellWithDisplay];
+  }
+
+  private findFirstCodeCellWithDisplay(): number {
```
languy, line 235: Can we move this to `Notebookutils.ts` and add a unit test?

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -111,9 +187,79 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
+      }
+    }
+
+    throw new Error("Output does not exist for any of the cells.");
+  }
+
   public render(): JSX.Element {
     return (
-      <div className="publishNotebookPanelContent">
+      <div className="publishNotebookPanelContent" ref={this.publishPaneRef}>
```
languy, line 262: Do we need this `publishPaneRef`?

src/Explorer/Tabs/NotebookV2Tab.ts
```diff
@@ -449,7 +449,11 @@ export default class NotebookTabV2 extends TabsBase implements ViewModels.Tab {
 
   private publishToGallery = () => {
     const notebookContent = this.notebookComponentAdapter.getContent();
-    this.container.publishNotebook(notebookContent.name, notebookContent.content);
+    this.container.publishNotebook(
+      notebookContent.name,
+      notebookContent.content,
+      this.notebookComponentAdapter.getNotebookParentDom()
```
languy, line 455: Rename to `getNotebookParentDomElement`

src/Explorer/Notebook/NotebookRenderer/NotebookRenderer.tsx
```diff
@@ -78,13 +85,22 @@ class BaseNotebookRenderer extends React.Component<NotebookRendererProps> {
 
   componentDidMount() {
     loadTransform(this.props as any);
+    this.props.updateNotebookParentDomElt(this.notebookRendererRef.current);
+  }
+
+  componentDidUpdate() {
+    this.props.updateNotebookParentDomElt(this.notebookRendererRef.current);
```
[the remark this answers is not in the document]
languy: Thinking about this more, I believe the one in `componentDidMount()` is sufficient: we are caching the parent of the component which should be stable. When `props` or state changes, React re-renders and may modify the DOM then call `componentDidUpdate`. But the top-level element should stay the same and the changes should just happen underneath it. Long-term, we should figure out a mechanism that doesn't cache the element as it is dangerous to rely on something that React can modify at any time.

src/Explorer/Explorer.ts
```diff
@@ -2347,9 +2348,9 @@ export default class Explorer {
     return Promise.resolve(false);
   }
 
-  public publishNotebook(name: string, content: string): void {
+  public publishNotebook(name: string, content: string | ImmutableNotebook, parentDomRef: HTMLElement): void {
     if (this.notebookManager) {
-      this.notebookManager.openPublishNotebookPane(name, content);
+      this.notebookManager.openPublishNotebookPane(name, content, parentDomRef);
```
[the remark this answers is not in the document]
languy: Check if `parentDomRef` is defined before opening the `publishNotebookPane`?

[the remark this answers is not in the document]
srinathnarayanan: I am checking for that inside the pane, the pane has other options like url and file upload and can still be opened and used even if parentHtmlElement is not yet set.

src/Explorer/Notebook/NotebookRenderer/NotebookRenderer.tsx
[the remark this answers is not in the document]
srinathnarayanan: I needed to to add the componentDidUpdate for the case when the notebook is opened, edited and then published. The ref was becoming undefined (I guess componenDidUnmount was kicking in)

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
srinathnarayanan: can make it 1024 * 1024 instead

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
srinathnarayanan: nope, will remove it

src/Explorer/Explorer.ts
```diff
@@ -79,6 +79,7 @@ import { UploadItemsPane } from "./Panes/UploadItemsPane";
 import { UploadItemsPaneAdapter } from "./Panes/UploadItemsPaneAdapter";
 import { ReactAdapter } from "../Bindings/ReactBindingHandler";
 import { toRawContentUri, fromContentUri } from "../Utils/GitHubUtils";
+import { ImmutableNotebook } from "@nteract/commutable";
```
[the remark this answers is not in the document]
southpolesteve: This needs to be checks for what it does to the JS size. Our JS payload for explorer.js is already massive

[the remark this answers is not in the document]
gitanuj: Btw I created `NotebookManager` which contains all notebook related stuff and is dynamically loaded by explorer. You can put all notebook specific logic in that if needed (to avoid increasing explorer's bundle size).

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -26,20 +31,25 @@ interface PublishNotebookPaneState {
 
 export class PublishNotebookPaneComponent extends React.Component<PublishNotebookPaneProps, PublishNotebookPaneState> {
   private static readonly maxImageSizeInMib = 1.5;
-  private static readonly ImageTypes = ["URL", "Custom Image"];
+  private static readonly Url = "URL";
+  private static readonly CustomImage = "Custom Image";
+  private static readonly TakeScreenshot = "Take Screenshot";
+  private static readonly UseFirstDisplayOutput = "Use First Display Output";
```
gitanuj, line 37: May be create a string enum for image type?

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -61,6 +71,39 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
+      };
+
+      target.scrollIntoView();
+      Html2Canvas(target, {
+        useCORS: true,
+        allowTaint: true,
+        scale: 1,
+        logging: true
+      })
+        .then(function(canvas) {
```
gitanuj, line 87: Can we use Promise awaits instead? We've been following that pattern everywhere.

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -61,6 +71,39 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
+        this.props.onChangeImageSrc(canvasUrl);
+        this.setState({ imageSrc: canvasUrl });
+      };
+
+      target.scrollIntoView();
+      Html2Canvas(target, {
+        useCORS: true,
+        allowTaint: true,
+        scale: 1,
+        logging: true
```
gitanuj, line 85: Do we want this logging for Prod builds too? Is there a way to plumb this logging to our `TelemetryProcessor`?

src/Explorer/Panes/PublishNotebookPaneComponent.tsx
```diff
@@ -61,6 +71,39 @@ export class PublishNotebookPaneComponent extends React.Component<PublishNoteboo
...
 
+    this.takeScreenshot = (target: HTMLElement, onError: (error: Error) => void): void => {
+      const updateImageSrcWithScreenshot = (canvasUrl: string): void => {
+        this.props.onChangeImageSrc(canvasUrl);
+        this.setState({ imageSrc: canvasUrl });
+      };
+
+      target.scrollIntoView();
+      Html2Canvas(target, {
+        useCORS: true,
```
gitanuj, line 82: Is there anything cross origin involved?

srinathnarayanan: this is for taking screenshots, there maybe a  tag in a markdown which will need this to be captured

srinathnarayanan:
> Need to compare the total build size before and after importing from @nteract in Explorer.ts

Removed nteract dependency from Explorer.ts

src/Explorer/Notebook/NotebookComponent/reducers.ts
```diff
@@ -4,6 +4,7 @@ import { Areas } from "../../../Common/Constants";
 import TelemetryProcessor from "../../../Shared/Telemetry/TelemetryProcessor";
 import * as cdbActions from "./actions";
 import { CdbRecord } from "./types";
+import { type } from "jquery";
```
southpolesteve: Whats this for?

src/Explorer/Explorer.ts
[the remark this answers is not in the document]
srinathnarayanan: I've removed this import

src/Explorer/Notebook/NotebookComponent/reducers.ts
srinathnarayanan: removed it, imported by mistake

src/Explorer/Notebook/NotebookComponent/reducers.ts
```diff
@@ -82,6 +82,19 @@ export const cdbReducer = (state: CdbRecord, action: Action) => {
...
+
+    case cdbActions.UPDATE_NOTEBOOK_PARENT_DOM_ELTS: {
+      const typedAction = action as cdbActions.UpdateNotebookParentDomEltAction;
+      var parentEltsMap = state.get("currentNotebookParentElements");
+      const contentRef = typedAction.payload.contentRef;
+      const parentElt = typedAction.payload.parentElt;
+      if (parentElt) {
+        parentEltsMap.set(contentRef, parentElt);
+      } else {
+        parentEltsMap.delete(contentRef);
```
languy, line 94: Can you test that if you close a notebook tab this gets reset? Otherwise, we should call the dispatch from the notebookTab when it closes?

src/Explorer/Notebook/NotebookUtil.test.ts
```diff
@@ -1,12 +1,73 @@
...
 import * as GitHubUtils from "../../Utils/GitHubUtils";
+import {
+  ImmutableNotebook,
+  MediaBundle,
+  CodeCellParams,
+  MarkdownCellParams,
+  makeCodeCell,
+  makeMarkdownCell,
+  makeNotebookRecord
+} from "@nteract/commutable";
```
languy, line 11: This works. FYI, nteract also has a `fixture` package that has pre-created notebooks in various shapes (small, big, object, json, etc).

src/Explorer/Notebook/NotebookUtil.ts
```diff
@@ -100,4 +100,30 @@ export class NotebookUtil {
...
+              return false;
+            });
+            if (displayOutput) {
+              return codeCellCount;
+            }
+          }
+        }
+      }
+    }
+    throw new Error("Output does not exist for any of the cells.");
```
languy, line 127: do we need to throw or just return -1? It doesn't seem like an error condition, but more like a normal not found condition.

src/Explorer/Notebook/NotebookUtil.ts
```diff
@@ -100,4 +100,30 @@ export class NotebookUtil {
...
+    let codeCellCount = -1;
+    for (let i = 0; i < notebookObject.cellOrder.size; i++) {
+      const cellId = notebookObject.cellOrder.get(i);
+      if (cellId) {
+        const cell = notebookObject.cellMap.get(cellId);
+        if (cell && cell.cell_type === "code") {
+          codeCellCount++;
+          const codeCell = cell as ImmutableCodeCell;
+          if (codeCell.outputs) {
+            const displayOutput = codeCell.outputs.find((output: ImmutableOutput) => {
```
languy, line 114: nit: just return the expression. No need to specify `output` type as it can be inferred.
`codeCell.outputs.find(output => output.output_type === "display_data" || output.output_type === "execute_result");

src/Explorer/Notebook/NotebookUtil.ts
```diff
@@ -100,4 +100,30 @@ export class NotebookUtil {
...
     return `${basePath}${newName}`;
   }
+
+  public static findFirstCodeCellWithDisplay(notebookObject: ImmutableNotebook): number {
+    let codeCellCount = -1;
+    for (let i = 0; i < notebookObject.cellOrder.size; i++) {
+      const cellId = notebookObject.cellOrder.get(i);
+      if (cellId) {
+        const cell = notebookObject.cellMap.get(cellId);
+        if (cell && cell.cell_type === "code") {
```
languy, line 110: This works. You might want to check the nteract's `selector` package to see if they have an iterator that saves you from doing all this access/check-if-exist logic.
Also, you can use the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#:~:text=%20%20%20%20Desktop%20%20%20,74%20%20%20IE%20No%20support%20No%20) to remove some of the `if(blah)` conditions.

src/Explorer/Notebook/NotebookUtil.ts
```diff
@@ -100,4 +100,30 @@ export class NotebookUtil {
...
+          const codeCell = cell as ImmutableCodeCell;
+          if (codeCell.outputs) {
+            const displayOutput = codeCell.outputs.find((output: ImmutableOutput) => {
+              if (output.output_type === "display_data" || output.output_type === "execute_result") {
+                return true;
+              }
+              return false;
+            });
+            if (displayOutput) {
+              return codeCellCount;
```
languy, line 121: Personally, I would prefer returning a 0-based index and -1 if not found to conform with usual `find` methods out there. If not, let's comment at the top of the function.

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -1,156 +1,178 @@
...
+
+  public open(
+    name: string,
+    author: string,
+    notebookContent: string | ImmutableNotebook,
+    parentDomElement: HTMLElement
+  ): void {
+    this.name = name;
+    this.author = author;
+    if (typeof notebookContent === "string") {
```
languy, line 69: I'm a little bit uneasy with the content being either string of object, because now every caller has to check the type, but I can't think of a better solution (returning always string by stringify'ing and parsing again seems inefficient indeed).

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: I catch the error to display the message on the generic pane. it doesn't bubble up

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: I tried that, compile errors were thrown since notebookContent is set to unknown type in Explorer.ts since we did not want to have any nteract dependencies

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: it is a 0 based index. i do the increment early. Maybe I can push it down. I throw an error instead of -1 for displaying it on the generic pane.

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
srinathnarayanan: yes I considered the same, but seemed inefficient. also, publish is the only action that calls the getContent() in the notebookComponentBootstrapper, where this notebookContent is defined as string | ImmutableNotebook. so we are good.

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: will add to next PR

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: will add some of these changes to next PR

PR #110, opened 2020-07-23
Move Delete Container call to use ARM when logged in with AAD

southpolesteve:
This PR moves a single call: `deleteContainer` to use ARM when logged in with AAD and the data plane SDK in all other cases. Most importantly, I hope it proves a useful pattern and foundation for moving our other calls to ARM. Specifics:

- Adds generated ARM clients for all Cosmos ARM operations
- Adds code for generating new ARM clients
- Adds a general purpose `armRequest` function that wraps fetch. I see we have something similar already, but it is stateful and I didn't want to try to refactor it at this time.
- Extracts `deleteContainer` from the DataAccessUtility into its own file. Adds a test.
- Refactors NotificationConsoleUtils to be a bit easier to use with this new pattern.


utils/armClientGenerator/generator.ts
```diff
@@ -16,6 +19,14 @@ function refToType(path: string | undefined) {
   return "unknown";
 }
 
+function camelize(str: string) {
+  return str
+    .replace(/(?:^\w|[A-Z]|\b\w)/g, function(word: any, index: any) {
```
languy, line 24: `(word, index) => {}`

southpolesteve:
Taking a second look at my own code, I don't love these as classes. Its forcing a bunch of state when none of these are stateful. Like `resourceGroup` will never change. Open to other opinions, but tempted to go back to bare functions even though it is slightly more verbose.

PR #111, opened 2020-07-24
Add react based AddDatabaseComponent

gitanuj:
Uses `GenericRightPaneComponent` and `SmartUiComponent`

### DIFF
--- a/src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.ts
+++ b/src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.ts
@@ -282,7 +282,8 @@ export class CommandBarComponentButtonFactory {
       iconSrc: AddDatabaseIcon,
       iconAlt: label,
       onCommandClick: () => {
-        container.addDatabasePane.open();
+        // container.addDatabasePane.open();
+        container.addDatabasePaneAdapter.open();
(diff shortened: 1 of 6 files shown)


src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.ts
```diff
@@ -282,7 +282,8 @@ export class CommandBarComponentButtonFactory {
       iconSrc: AddDatabaseIcon,
       iconAlt: label,
       onCommandClick: () => {
-        container.addDatabasePane.open();
+        // container.addDatabasePane.open();
```
languy, line 285: Remove?

src/Explorer/Panes/AddDatabaseComponent.tsx
```diff
@@ -0,0 +1,121 @@
...
+          }
+        });
+      }
+    }
+
+    return <SmartUiComponent {...props} />;
+  }
+
+  private onChange = (newValues: Map<string, InputType>): void =>  {
+    newValues.forEach((value: InputType, key: string) => {
```
languy, line 90: It's the simplest I could find to return the values, but if you can think of a more optimal way, feel free to modify `SmartUiComponent`.

PR #113, opened 2020-07-24
Fix handling numeric partition keys

languy:
Rather than being sent as a gremlin query, the `g.V()` query is intercepted by GraphExplorer and processed as a sql query: this allows paging and the code path involves processing the response and issuing more (gremlin) queries to display the node's neighbors.

This change fixes a bug that incorrectly errors when processing the `g.V()` results and encountering numeric partition key value.


southpolesteve:
@languy can this be merged?

PR #117, opened 2020-07-27
Added support for acknowledging code of conduct for using public Notebook Gallery

srinathnarayanan:
This PR
- Adds corresponding front end code for this JUNO tools change : https://msdata.visualstudio.com/CosmosDB/_git/CosmosDB-portal/pullrequest/401396
- Adds a Code of Conduct Component, with links to microsoft's code of conduct for using public material and privacy statement
- Adds a feature flag (enableCodeOfConduct) to hide this feature
- Ensures that Acknowledging the code of conduct sets a flag in the user metadata
- Adds an Info component to display "Help" box with links to code of conduct and privacy statement

code of conduct

code of conduct-2


src/Juno/JunoClient.ts
```diff
@@ -208,6 +208,24 @@ export class JunoClient {
...
 
+  public async isCodeOfConductAccepted(): Promise<IJunoResponse<boolean>> {
+    const input = `${this.getNotebooksAccountUrl()}/gallery/isCodeOfConductAccepted`;
+    const response = await window.fetch(input, {
+      method: "PATCH",
+      headers: JunoClient.getHeaders()
+    });
+
+    let data: boolean;
+    if (response.status === HttpStatusCodes.OK) {
```
southpolesteve, line 219: You can check `response.ok` here

src/Juno/JunoClient.ts
```diff
@@ -208,6 +208,24 @@ export class JunoClient {
     };
   }
 
+  public async isCodeOfConductAccepted(): Promise<IJunoResponse<boolean>> {
+    const input = `${this.getNotebooksAccountUrl()}/gallery/isCodeOfConductAccepted`;
+    const response = await window.fetch(input, {
```
southpolesteve, line 213:
```suggestion
    const url = `${this.getNotebooksAccountUrl()}/gallery/isCodeOfConductAccepted`;
    const response = await window.fetch(url, {

```

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -58,12 +62,30 @@ export class PublishNotebookPaneAdapter implements ReactAdapter {
...
     notebookContent: string | ImmutableNotebook,
-    parentDomElement: HTMLElement
-  ): void {
+    parentDomElement: HTMLElement,
+    isCodeOfConductEnabled: boolean
+  ): Promise<void> {
+    if (isCodeOfConductEnabled) {
+      try {
+        const response = await this.junoClient.isCodeOfConductAccepted();
+        if (response.status !== HttpStatusCodes.OK && response.status !== HttpStatusCodes.NoContent) {
```
southpolesteve, line 75:
```suggestion
        if (response.ok) {

```

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -86,6 +89,7 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
   private publicNotebooks: IGalleryItem[];
   private favoriteNotebooks: IGalleryItem[];
   private publishedNotebooks: IGalleryItem[];
+  private acceptedCodeOfConduct: boolean;
```
languy, line 92: `isCodeOfConductAccepted`

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -134,9 +139,17 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
...
+      tabs.push(
+        this.createPublicGalleryTab(
+          GalleryTab.PublicGallery,
+          this.state.publicNotebooks,
+          this.state.acceptedCodeOfConduct
+        )
+      );
       tabs.push(this.createTab(GalleryTab.Favorites, this.state.favoriteNotebooks));
-      tabs.push(this.createTab(GalleryTab.Published, this.state.publishedNotebooks));
+      if (this.state.acceptedCodeOfConduct !== false) {
```
languy, line 150: Remove the `!== false` part (it's a boolean, right?)

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -167,13 +180,37 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
     );
   }
 
+  private createPublicGalleryTab(
+    tab: GalleryTab,
+    data: IGalleryItem[],
+    acceptedCodeOfConduct: boolean
```
languy, line 186: `isCodeOfConductAccepted`

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -167,13 +180,37 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
...
+
   private createTab(tab: GalleryTab, data: IGalleryItem[]): GalleryTabInfo {
     return {
       tab,
       content: this.createTabContent(data)
     };
   }
 
+  private createPublicGalleryTabContent(data: IGalleryItem[], acceptedCodeOfConduct: boolean): JSX.Element {
+    return acceptedCodeOfConduct === false ? (
```
languy, line 202: Replace `acceptedCodeOfConduct === false` with `!acceptedCodeOfConduct`

src/Explorer/Controls/NotebookGallery/CodeOfConductComponent.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+describe("CodeOfConductComponent", () => {
+  it("renders", () => {
+    const props: CodeOfConductComponentProps = {
+      junoClient: undefined,
+      onAcceptCodeOfConduct: undefined
+    };
+
+    const wrapper = shallow(<CodeOfConductComponent {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
```
languy, line 14: It would be good if you can simulate a click on the submit button and verify that it makes a remote call with checkbox on.

src/Explorer/Explorer.ts
```diff
@@ -194,6 +194,7 @@ export default class Explorer {
 
   // features
   public isGalleryPublishEnabled: ko.Computed<boolean>;
+  public isCodeOfConductEnabled: ko.Computed<boolean>;
```
languy, line 197: If possible, I would scope this to gallery publish: e.g `isPublishCodeOfConductEnabled`. So it's more self-explanatory as to which feature this applies to.

src/Explorer/Notebook/NotebookUtil.ts
```diff
@@ -102,25 +102,19 @@ export class NotebookUtil {
   }
 
   public static findFirstCodeCellWithDisplay(notebookObject: ImmutableNotebook): number {
-    let codeCellCount = -1;
+    let codeCellCount = 0;
```
languy, line 105: Add a comment as to why we're now skipping the first cell. Btw, since the newly injected cell isn't a code cell, I would keep starting from -1. It won't hurt and we're not adding more dependency to this special cell. If later on, the mechanism changes we won't have to go back here and fix.

languy: It's really an index rather than a count. Rename to `cellIndex`?

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -26,6 +28,7 @@ export class PublishNotebookPaneAdapter implements ReactAdapter {
   private imageSrc: string;
   private notebookObject: ImmutableNotebook;
   private parentDomElement: HTMLElement;
+  private acceptedCodeOfConduct: boolean;
```
languy, line 31: `isCodeOfConductAccepted`

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -104,11 +126,9 @@ export class PublishNotebookPaneAdapter implements ReactAdapter {
         this.imageSrc,
         this.content
       );
-      if (!response.data) {
-        throw new Error(`Received HTTP ${response.status} when publishing ${name} to gallery`);
+      if (response.data) {
+        NotificationConsoleUtils.logConsoleMessage(ConsoleDataType.Info, `Published ${name} to gallery`);
       }
```
languy, line 131: Should an error be displayed if there is no `response.data`?

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -158,10 +178,22 @@ export class PublishNotebookPaneAdapter implements ReactAdapter {
...
+          }}
+        />
+      </div>
+    ) : (
+      <PublishNotebookPaneComponent {...publishNotebookPaneProps} />
+    );
   };
 
-  private reset = (): void => {
+  private reset = async (): Promise<void> => {
```
languy, line 196: I'm not seeing any async call inside. Or am I missing something? If it's to confirm to some signature in the caller, add a comment.

src/Juno/JunoClient.ts
```diff
@@ -162,6 +171,61 @@ export class JunoClient {
     return this.getNotebooks(`${this.getNotebooksUrl()}/gallery/public`);
   }
 
+  // will be renamed once feature.enableCodeOfConduct flag is removed
+  public async fetchPublicNotebooks(): Promise<IJunoResponse<IPublicGalleryData>> {
+    const input = `${this.getNotebooksAccountUrl()}/gallery/public`;
+    const response = await window.fetch(input, {
+      method: "PATCH",
```
languy, line 178: Is `PATCH` intended here?

src/Juno/JunoClient.ts
```diff
@@ -310,6 +374,8 @@ export class JunoClient {
     let data: IGalleryItem;
     if (response.status === HttpStatusCodes.OK) {
       data = await response.json();
+    } else {
+      throw new Error(`HTTP status ${response.status} thrown. ${(await response.json()).Message}`);
```
languy, line 378: Maybe replace `thrown` by `received`.

src/Juno/JunoClient.ts
srinathnarayanan: this is the general pattern followed here. Will make these changes as part of a clean up PR.

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
srinathnarayanan: this "response" is of type IJunoResponse, which only has the statusCode

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
srinathnarayanan: So the reason I check for false is, if its is undefined to start with when galleryViewerComponent is loaded and i check for (!this.state.acceptedCodeOfConduct), then the codeOfConduct component is always shown first and then removed to make way for the gallery (in case the juno call to read codeOfConduct flag returns true). So for a few milliseconds, the code of conduct component will pop up even if it has been formerly accepted. The experience looked bad, so I resorted to this type of check.

src/Explorer/Explorer.ts
srinathnarayanan: this code of conduct needs to be accepted for both publish and for viewing the public gallery (since download and usage of public data is possible in that case)

src/Explorer/Notebook/NotebookUtil.ts
srinathnarayanan: this change was independent of the linkInjection change. I reworked the method to use base index as 0, just to make it look better.will rename to codecellIndex

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
srinathnarayanan: an error will be thrown from junoClient.publishNotebook() in case publish fails, the catch block will catch and display it.

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
srinathnarayanan: reverted it

src/Juno/JunoClient.ts
srinathnarayanan: yes, I followed the pattern where all authenticated requests are sent as PATCH requests. Juno also accepts a patch request on this URL

src/Juno/JunoClient.ts
srinathnarayanan: the only place where this is called is covered by a try catch block , with a notification consle logError message in the catch block.

src/Explorer/Explorer.ts
languy: Ok then maybe do `isGalleryCodeOfConductEnabled`. Otherwise it would appear it's the CodeOfConduct for DataExplorer as a whole.

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
languy: I see. Then I would opt for something more explicit:
`this.state.acceptedCodeOfConduct !== undefined && !this.state.acceptedCodeOfConduct`
Or add a comment about why you're using this construct.

srinathnarayanan: will add a comment

src/Explorer/Explorer.ts
srinathnarayanan: will leave it as such for now, since this is a feature flag that will be removed shortly.

src/Explorer/Controls/NotebookGallery/CodeOfConductComponent.test.tsx
```diff
@@ -0,0 +1,33 @@
+import { shallow } from "enzyme";
+import * as sinon from "sinon";
+import React from "react";
+import { CodeOfConductComponent, CodeOfConductComponentProps } from "./CodeOfConductComponent";
+import { IJunoResponse, JunoClient } from "../../../Juno/JunoClient";
+import { HttpStatusCodes } from "../../../Common/Constants";
+
+sinon.stub(JunoClient.prototype, "acceptCodeOfConduct").returns({
```
vignesh-mr, line 8: Might want to set this up in a separate sinon sandbox so this does not affect other tests

src/Explorer/Controls/NotebookGallery/CodeOfConductComponent.tsx
```diff
@@ -0,0 +1,108 @@
...
+              label: {
+                margin: 0,
+                padding: "2 0 2 0"
+              },
+              text: {
+                fontSize: 12
+              }
+            }}
+            label="I have read and accepted the code of conduct and privacy statement"
+            onChange={() => this.setState({ readCodeOfConduct: !this.state.readCodeOfConduct })}
```
vignesh-mr, line 90: Might want to track the lambda through a var so it doesn't create one with every re-render. Same below.

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -134,9 +139,20 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
...
+        this.createPublicGalleryTab(
+          GalleryTab.PublicGallery,
+          this.state.publicNotebooks,
+          this.state.isCodeOfConductAccepted
+        )
+      );
       tabs.push(this.createTab(GalleryTab.Favorites, this.state.favoriteNotebooks));
-      tabs.push(this.createTab(GalleryTab.Published, this.state.publishedNotebooks));
+
+      // explicitely checking if isCodeOfConductAccepted is not false, as it is initially undefined.
```
vignesh-mr, line 151: explicitly*

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.less
```diff
@@ -0,0 +1,26 @@
...
+    padding-left: 5px;
+    padding-right: 5px;
+}
+
+.infoLabel, .infoLabelMain {
+    padding-left: 5px  
+}
+
+.infoLabel {
+    font-weight: 400
```
vignesh-mr, line 17: reuse constants as much as possible

srinathnarayanan: I can't find any constants for font weight as of now.

PR #118, opened 2020-07-28
Updating error message on mongo collection create

vchske:
1) Updated mongo collection create pane to display a better error when a shard key is ill formed.

2) Updated an error message to use double quotes since no string interpolation is used. I thought it was included in a previous PR but I guess the change didn't get staged.

### DIFF
--- a/src/Common/ErrorParserUtility.ts
+++ b/src/Common/ErrorParserUtility.ts
@@ -7,6 +7,10 @@ export function replaceKnownError(err: string): string {
     err.indexOf("SharedOffer is Disabled for your account") >= 0
   ) {
     return "Database throughput is not supported for internal subscriptions.";
+  } else if (
+    err.indexOf("Partition key paths must contain only valid") >= 0
+  ) {
+    return "Partition key paths must contain only valid characters and not contain a trailing slash or wildcard character.";
   }
 
   return err;
--- a/src/Common/ErrorParserUtility.ts
+++ b/src/Common/ErrorParserUtility.ts
@@ -7,9 +7,7 @@ export function replaceKnownError(err: string): string {
     err.indexOf("SharedOffer is Disabled for your account") >= 0
   ) {
     return "Database throughput is not supported for internal subscriptions.";
-  } else if (
-    err.indexOf("Partition key paths must contain only valid") >= 0
-  ) {
+  } else if (err.indexOf("Partition key paths must contain only valid") >= 0) {
     return "Partition key paths must contain only valid characters and not contain a trailing slash or wildcard character.";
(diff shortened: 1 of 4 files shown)


src/Common/ErrorParserUtility.ts
```diff
@@ -1,67 +1,71 @@
...
-    // Single-partition error flavor
-    const regExp = /^Message: (.*)ActivityId: (.*), Request URI: (.*)/g;
-    const regString = regExp.exec(singleLineMessage);
-    const innerMessageString = regString[1];
-    innerMessage = JSON.parse(innerMessageString);
-  }
-
-  return innerMessage.errors ? innerMessage.errors : innerMessage.Errors;
-}
+﻿import * as DataModels from "../Contracts/DataModels";
```
vchske, line 1: Not at all sure why the diff looks strange for this file. Locally it looks like only 4 lines were inserted. Any ideas? Just curious.

southpolesteve: Might need to merge in master and run format. I made a bunch of changes in the last few days.

vchske: Ran format.

src/Common/ErrorParserUtility.ts
```diff
@@ -1,67 +1,69 @@
...
-    window.dataExplorer.subscriptionType() === ViewModels.SubscriptionType.Internal &&
-    err.indexOf("SharedOffer is Disabled for your account") >= 0
-  ) {
-    return "Database throughput is not supported for internal subscriptions.";
-  }
-
-  return err;
-}
-
-export function parse(err: any): DataModels.ErrorDataModel[] {
```
vchske, line 15: I did not remove ErrorParserUtility from eslintignore because of several uses of `any` for errors. I don't think we're using any unified error object (this fact spurred the investigation into this fix in the first place). I think this will entail a more thorough fix throughout several files.

src/Utils/PricingUtils.ts
```diff
@@ -7,10 +7,11 @@ import { AutopilotTier } from "../Contracts/DataModels";
...
  */
-export function normalizeNumber(number: any): number {
-  const normalizedNumber: number = number === null ? 0 : isNaN(number) ? 0 : parseInt(number);
-
-  return normalizedNumber;
+export function normalizeNumber(number: null | undefined | string | number): number {
+  if (!number) {
+    return 0;
+  }
+  return Math.floor(Number(number));
```
vchske, line 14: Judging by the unit tests, the floor is expected. Let me know if you think otherwise.

PR #120, opened 2020-07-29
Added session based view updation for gallery notebooks

srinathnarayanan:


src/Shared/StorageUtility.ts
```diff
@@ -44,7 +44,11 @@ export class SessionStorageUtility {
   }
 
   public static getEntryNumber(key: StorageKey): number {
-    return StringUtility.toNumber(localStorage.getItem(StorageKey[key]));
+    return StringUtility.toNumber(sessionStorage.getItem(StorageKey[key]));
```
southpolesteve, line 47: Is there anywhere relying on this being `localStorage`?

srinathnarayanan: nope, this isn't being called anywhere. safe to change

PR #133, opened 2020-08-06
Add generated clients to strict compile

southpolesteve:


src/Utils/arm/request.ts
```diff
@@ -30,11 +30,15 @@ interface Options {
 // TODO: This is very similar to what is happening in ResourceProviderClient.ts. Should probably merge them.
 export async function armRequest<T>({ host, path, apiVersion, method, body: requestBody }: Options): Promise<T> {
```
languy, line 31: This is a bit subjective, but can we keep the convention of capitalizing the module filenames?

southpolesteve: I'm in favor but I would consider it out of scope for this PR. There are linting rules for it. 

I consider this to be idiomatic to JS, but I don't necessarily think we have to comply 100%:
- variables and method names are camelCased
- Class names are capitalized
- filenames are snake-case. I believe this is mostly because npm does not allow upper case names. snake-case makes it easy to extract a file/folder into an npm module.

southpolesteve:
Closing as stale. Will revisit later.

PR #134, opened 2020-08-06
Move readCollection, readCollections, and readDatabases to ARM

victor-meng:
Move readCollection, readCollections, and readDatabases to ARM


src/Common/dataAccess/deleteCollection.test.ts
```diff
@@ -1,23 +1,44 @@
...
+
+  it("should call ARM if logged in with AAD", async () => {
+    window.authType = AuthType.AAD;
     await deleteCollection("database", "collection");
     expect(armRequest).toHaveBeenCalled();
   });
-  // TODO: Test non-AAD case
+
+  it("should call SDK if not logged in with non-AAD method", async () => {
+    window.authType = AuthType.MasterKey;
```
southpolesteve, line 29: Thanks!

src/Common/dataAccess/readCollection.ts
```diff
@@ -0,0 +1,24 @@
...
+import { client } from "../CosmosClient";
+import { logConsoleProgress, logConsoleError } from "../../Utils/NotificationConsoleUtils";
+import { logError } from "../Logger";
+import { sendNotificationForError } from "./sendNotificationForError";
+
+export async function readCollection(databaseId: string, collectionId: string): Promise<DataModels.Collection> {
+  let collection: DataModels.Collection;
+  const clearMessage = logConsoleProgress(`Querying container ${collectionId}`);
+  try {
+    const response = await client()
```
ato9000, line 11: Where is the RP support for readCollection?

src/Common/dataAccess/readCollections.ts
```diff
@@ -0,0 +1,42 @@
...
+import { logError } from "../Logger";
+import { sendNotificationForError } from "./sendNotificationForError";
+import { userContext } from "../../UserContext";
+
+export async function readCollections(databaseId: string): Promise<DataModels.Collection[]> {
+  let collections: DataModels.Collection[];
+  const clearMessage = logConsoleProgress(`Querying containers for database ${databaseId}`);
+  try {
+    if (window.authType === AuthType.AAD) {
+      const rpResponse = await listSqlContainers(
```
ato9000, line 15: Is this for SQL only or does listSqlContainers takes care of all APIs?

src/Common/dataAccess/readDatabases.ts
```diff
@@ -0,0 +1,40 @@
...
+import { logError } from "../Logger";
+import { sendNotificationForError } from "./sendNotificationForError";
+import { userContext } from "../../UserContext";
+
+export async function readDatabases(): Promise<DataModels.Database[]> {
+  let databases: DataModels.Database[];
+  const clearMessage = logConsoleProgress(`Querying databases`);
+  try {
+    if (window.authType === AuthType.AAD) {
+      const rpResponse = await listSqlDatabases(
```
ato9000, line 15: Same question: is this SQL only or does listSqlDatabases handles all APIs?

src/Common/dataAccess/readCollection.ts
victor-meng: As discussed offline, readCollection is only called during the resource token workflow so RP support is not required.

src/Common/dataAccess/readCollections.ts
victor-meng: I've added support for all APIs in the latest commit.

PR #139, opened 2020-08-10
Add init message when warming up notebook workspace

vignesh-mr:
Add init message when warming up notebook workspace

### DIFF
--- a/src/Explorer/Explorer.ts
+++ b/src/Explorer/Explorer.ts
@@ -1723,6 +1723,7 @@ export default class Explorer {
       return;
     }
 
+    let notificationId;
     try {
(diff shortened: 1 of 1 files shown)


src/Explorer/Explorer.ts
```diff
@@ -1734,10 +1735,17 @@ export default class Explorer {
         notebookWorkspace.properties.status &&
         notebookWorkspace.properties.status.toLowerCase() === "stopped"
       ) {
+        notificationId = NotificationConsoleUtils.logConsoleMessage(
```
southpolesteve, line 1738: There is a new set of methods you can use for this that is less verbose:

`const clearMessage = NotificationConsoleUtils.logConsoleProgress("message")`

... later

`clearMessage()`

PR #140, opened 2020-08-10
Add subscriptionId and authType to telemetry

southpolesteve:
Need to add this info to telemetry to better understand connection string vs AAD usage


src/Shared/Telemetry/TelemetryProcessor.ts
```diff
@@ -115,6 +116,9 @@ export default class TelemetryProcessor {
 
   private static getData(data?: any): any {
     return {
+      // TODO: Need to `any` here since the window imports Explorer which can't be in strict mode yet
+      authType: (window as any).authType,
+      subscriptionId: userContext.subscriptionId,
```
ato9000, line 121: No need to do anything now, but you will eventually need to introduce an "account context" where subscrptionId belongs. It will be necessary once we support more than one account.

PR #143, opened 2020-08-11
Fix deleteDatabase and deleteCollection with ARM

victor-meng:
Currently deleteDatabase and deleteCollection always use the SQL URI for deleting database and collection. It should determine the correct URI to use based on the default experience.


src/Common/dataAccess/deleteDatabase.ts
```diff
@@ -33,3 +32,23 @@ export async function deleteDatabase(databaseId: string): Promise<void> {
   clearMessage();
   await refreshCachedResources();
 }
+
+function deleteDatabaseWithARM(databaseId: string): Promise<void> {
+  const subscriptionId: string = userContext.subscriptionId;
```
southpolesteve, line 37: None of these need the type assertion, the compiler can infer it. Same in the other files:

```suggestion
  const subscriptionId = userContext.subscriptionId;
```

victor-meng: Done

PR #146, opened 2020-08-12
Refactor GenericRightPaneComponent to accept a ReactComponent as its content

victor-meng:
GenericRightPaneComponent now accepts a ReactComponent as its content instead of a JSX.Element passed via props.content.

Also updated UploadItemsPaneAdapter, CopyNotebookPaneAdapter, and PublishNotebookPaneAdapter to reflect the changes in GenericRightPaneComponent.

### DIFF
--- a/src/Explorer/Panes/CopyNotebookPaneComponent.tsx
+++ b/src/Explorer/Panes/CopyNotebookPaneComponent.tsx
@@ -1,0 +1,119 @@
+import * as GitHubUtils from "../../Utils/GitHubUtils";
+import * as React from "react";
+import { IPinnedRepo } from "../../Juno/JunoClient";
+import { ResourceTreeAdapter } from "../Tree/ResourceTreeAdapter";
+import {
+  Stack,
+  Label,
+  Text,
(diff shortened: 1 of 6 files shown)


src/Explorer/Panes/CopyNotebookPaneComponent.tsx
```diff
@@ -0,0 +1,119 @@
...
+      ariaLabel: "Location",
+      placeholder: "Select an option",
+      onRenderTitle: this.onRenderDropDownTitle,
+      onRenderOption: this.onRenderDropDownOption,
+      options: this.getDropDownOptions(),
+      onChange: this.props.onDropDownChange
+    };
+
+    return (
+      <div className="paneMainContent">
```
gitanuj, line 47: Can we also put this in `GenericPaneComponent` so that all the panes using it don't have to do it themselves?

victor-meng: Look likes the other panes have slightly different layering. I will probably do this in another PR.

PR #148, opened 2020-08-13
Remove ExplorerOptions

southpolesteve:
This removes the last two pieces of state passed to `new Explorer()`
- `isEmulator`. Pretty obvious what this does. checking `configContext.platform` is now preferred
- Remove `NotificationsClient`. This was threaded through a lot of stuff, but only really used when an offer was in a state of massive throughput change. But even then, it doesn't appear to actually get used. This state is now detectable from the offer header. No need to use the notifications API. This banner appears even after removing the notifications client code:

Screenshot 2020-08-12 214711


southpolesteve:
It looks like I let this one sit too long and now has some significant conflicts. Will take another shot at it later.

Issue #103, opened 2020-07-20
Move All Control Plane calls to Cosmos Resource Provider


southpolesteve:
Anywhere we are using the JS SDK to perform control plane operations, we need to switch to Cosmos Resource Provider when logged in as an AAD user. I generated a preliminary list of all the operations that need to be checked:

- [x] Update Collection
- [ ] Update Offer
- [x] Delete Collection
- [x] Delete Database
- [x] Read Collections
- [x] Read Databases
- [x] Read Collection
- [ ] Read Collection Quota Info
- [ ] Read Offer
- [ ] Read Offers
- [x] Read Databases
- [x] Get or Create Database and Collection
- [x] Create Database
- [x] Refresh Cached Offers
- [x] Refresh Cached Resources
- [x] Read Subcription
- [x] Read Subscription Defaults

Each of these will also need possibly support a fallback to the JS SDK for connection string users.

southpolesteve:
Read Subscription and Read Subscription Defaults are not used anymore https://github.com/Azure/cosmos-explorer/pull/107

victor-meng:
Calls that are difficult to move to RP:

- Mongo:
	- readDatabases and readCollections: The databases and collections that RP returns do not have the `ExtendedResourceProperties` such as `_rid` and `_self`. These properties are used in several places in data explorer so we need to do some refactoring before we can move these two calls to RP.
- Table:
	- all database level operations (readDatabases, deleteDatabase, etc): RP does not support any database level operations for table API.

southpolesteve:
Closing. We ended up tracking this work outside of GitHub. There are a few remaining calls waiting on full RP support but the vast majority of logged in AAD calls go through RP

PR #155, opened 2020-08-18
Switch back to SDK for deleteDatabase and use RP for readCollections for table API

victor-meng:
`deleteDatabase`: RP does not support any operation on the database level for table API so we have to switch back to using SDK.
`readCollections`: I moved this call back to SDK but it actually works using RP as well. Therefore, moving the table call back to RP.


southpolesteve:
@victor-meng Can you also keep a list of difficult calls to move to ARM in #103 with a small note of what makes it hard? Feel free to edit my issue body

PR #161, opened 2020-08-21
Refactored Settings Tab

srinathnarayanan:
This PR 
- Adds a Settings v2 tab, which hosts a SettingsComponent
- Heirarchy of new components : 
  - Settings Component
    -- ScaleComponent
        -- ThroughputInputReactComponent
        -- ThroughputInputReactComponentAutoPilotV3
    -- SubSettingsComponent
    -- IndexingPolicyComponent
    -- ConflictResolutionComponent
- A ToolTipLabel component was added to render labels with tool tips 
- All Components are react based
- Hidden behind enablesettingsv2 feature flag
- Moved all Inputs to Fluent UI and added Tabs for the 4 sub components
settingsv2


src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2669 @@
...
+
+  private setAutoPilotSelected = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    this.setState({ isAutoPilotSelected: e.currentTarget.value === "true" });
+  };
+
+  private setAutoPilotTier = (selectedAutoPilotTier: DataModels.AutopilotTier): void => {
+    this.setState({ selectedAutoPilotTier: selectedAutoPilotTier });
+  };
+
+  public render(): JSX.Element {
```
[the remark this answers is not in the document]
srinathnarayanan: Will breakdown into separate components

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2669 @@
...
+          <div className="warningErrorContainer scaleWarningContainer">
+            <>
+              {this.shouldShowNotificationStatusPrompt() && (
+                <div className="warningErrorContent">
+                  <span>
+                    <img src={InfoColor} alt="Info" />
+                  </span>
+                  <span
+                    className="warningErrorDetailsLinkContainer"
+                    dangerouslySetInnerHTML={{ __html: this.state.notificationStatusInfo }}
```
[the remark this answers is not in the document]
srinathnarayanan: I was going to rework it in later PRs, but can do it as part of this PR itself.

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,42 @@
+export interface StatefulValue<T> {
```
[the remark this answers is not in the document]
srinathnarayanan: I wanted to keep it generic in case we wanted enums and other objects later to reuse this pattern

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
```diff
@@ -0,0 +1,191 @@
...
+}
+
+export class ThroughputInputComponent extends React.Component<ThroughputInputProps, ThroughputInputState> {
+  private defaultStep = 100;
+
+  public constructor(props: ThroughputInputProps) {
+    super(props);
+    this.state = {
+      spendAckChecked: this.props.spendAckChecked,
+      step: this.props.step ? this.props.step : this.defaultStep
```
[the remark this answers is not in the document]
srinathnarayanan: currently, in the places where we pass the step param, it is set to 100. We shouldn;t be having a step of 0. So will go witht he '??' pattern

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
```diff
@@ -0,0 +1,191 @@
...
+    };
+  }
+
+  private onThroughputChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    let newThroughput = parseInt(e.currentTarget.value);
+    newThroughput = isNaN(newThroughput) ? 0 : newThroughput;
```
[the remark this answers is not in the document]
srinathnarayanan: I check for the 0 and make the text input blank if it is 0. The blank input corresponds to undefined input and a save operation will not go through.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2669 @@
...
+  Rupm = "rupm",
+  ConflictResolutionPolicyMode = "conflictResolutionPolicyMode",
+  ConflictResolutionPolicyPath = "conflictResolutionPolicyPath",
+  ConflictResolutionPolicyProcedure = "conflictResolutionPolicyProcedure",
+  AnalyticalStorageTtlSelection = "analyticalStorageTtlSelection",
+  AnalyticalStorageTtlSeconds = "analyticalStorageTtlSeconds",
+  ChangeFeedPolicyToggled = "changeFeedPolicyToggled"
+}
+
+interface UpdateStatefulValueParams {
```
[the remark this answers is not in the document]
srinathnarayanan: StatefulValue is used to track the baseline/default and current values of a state object. It is modelled after ViewModels.Editable used in the ko version. If the baseline and default are different, only then is a save allowed. When they are different, css is also changed to indicate the same. Will add a comment.

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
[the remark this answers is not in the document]
languy: Then let's define this "magic" value as a constant to document.

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,42 @@
...
+  baseline: T;
+  current: T;
+  isValid: boolean;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
srinathnarayanan: Will move most of it to unknown. There are couple of place where I still have to use any, since it deal with indexingPolicy returned from the container which has 'any' type.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2669 @@
...
+      key: StatefulValueNames.Throughput,
+      value: this.state.throughput.baseline,
+      updateCurrent: true
+    });
+    this.updateStatefulValue({
+      key: StatefulValueNames.TimeToLive,
+      value: this.state.timeToLive.baseline,
+      updateCurrent: true
+    });
+    this.updateStatefulValue({
```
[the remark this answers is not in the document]
srinathnarayanan: so StatefulValue is a new interface I've defined modelled around ViewModels.Editable, used in ko version of the SettingsTab. These StatefulValue<> objects are part of the state of the SettingsTab, and are modified when controls are changed in the SettingsTab.
For example, for modifying throughput:
1) SettingsComponentState has a "throughput" field of type StatefulValue<number>
2) baseline and current values for state.throughput are set to 'undefined' in the constructor
3) when saving, this.state.throughput.current is used
4) when resetting, this.state.throughput.current is set to this.state.throughput.baseline (in the updateStatefulValue function)
5) When modifying, this.state.thoughput.current is set to the new value from the input (using updateStatefulValue)

updateStatefulValue function was introduced to reuse code since the same functionality is replicated for the different state fields that need both baseline and current values tracked.

src/Explorer/Controls/Settings/SettingsUtils.tsx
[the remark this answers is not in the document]
southpolesteve: In that case we should fix the indexingPolicy type to be correct.

src/Explorer/Controls/Settings/SettingsUtils.tsx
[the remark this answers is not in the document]
srinathnarayanan: nvm, I added this.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2588 @@
...
+                </div>
+              </div>
+            )}
+          </div>
+        )}
+      </>
+    );
+  };
+
+  private getConflictResolutionModeComponent = (): JSX.Element => {
```
languy, line 1952: Now that we have these inputs nicely isolated to separate functions. Can we create separate `ReactComponents` for each of them?
If they are "dumb" components (ie without elaborate biz logic, etc) you can make them a react function component (see `FeaturePanelComponent` for example). A react function component is essentially one single big function that renders some props. You can also add state to it and side effects. Avoid nested functions (that are harder to unit test).

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2588 @@
...
+            />
+          </div>
+        </div>
+        {this.state.analyticalStorageTtlSelection.current === TtlType.On && (
+          <>
+            <input
+              type="number"
+              required
+              min="1"
+              max="2147483647"
```
languy, line 2395: This number seems to be already be defined in `EntityPropertyValidationCommons`.

src/Explorer/Controls/Settings/SettingsComponent.tsx
languy: Once separate in components, how hard would it be to convert to fluent-ui? Since we're in this code, we might as well go all the way in fixing/refactor.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2588 @@
...
+              </>
+            )}
+          </div>
+        )}
+      </>
+    );
+  };
+
+  public render(): JSX.Element {
+    return (
```
languy, line 2570: Very nice! `render()` looks much more manageable!

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,331 @@
...
+  baseline: StatefulValuesType;
+  current: StatefulValuesType;
+  isValid: boolean;
+}
+
+const getStringValue = (value: StatefulValuesType): string => {
+  const type = typeof value;
+  switch (type) {
+    case "string":
+    case "undefined":
```
languy, line 35: In the "undefined" type, it would return "undefined". Is it what we want?

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,331 @@
...
+    default:
+      return JSON.stringify(value);
+  }
+};
+
+export function isDirty(value: StatefulValue<StatefulValuesType>): boolean {
+  const current = getStringValue(value.current);
+  const baseline = getStringValue(value.baseline);
+
+  if (current !== baseline) {
```
languy, line 49: `return current !== balseline;`

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,331 @@
...
+    return true;
+  }
+
+  return false;
+}
+
+export function getAutoPilotV3SpendElement(
```
languy, line 56: These functions below are a typical good use case for a `React.FunctionComponent`. Make this a separate component. We can then unit test with `maxAutoPiloThroughputSet` set or not set, etc.
A react component should be reusable, so it might make more sense to group them into one or only a couple of components rather than converting every single one of them to a component.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,2588 @@
...
+  AnalyticalStorageTtlSelection = "analyticalStorageTtlSelection",
+  AnalyticalStorageTtlSeconds = "analyticalStorageTtlSeconds",
+  ChangeFeedPolicyToggled = "changeFeedPolicyToggled",
+  AutoPilotThroughput = "autoPilotThroughput"
+}
+
+interface UpdateStatefulValueParams {
+  key: keyof SettingsComponentState;
+  value: StatefulValuesType;
+  updateCurrent: boolean;
```
languy, line 89: I'm not seeing a place where we use `updateCurrent: false`. Is it needed?

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,331 @@
...
+  getCurrencySign,
+  getAutoscalePricePerRu,
+  getMultimasterMultiplier,
+  computeRUUsagePriceHourly,
+  getPricePerRu,
+  calculateEstimateNumber
+} from "../../../Utils/PricingUtils";
+
+/*
+Class used to track baseline and current values of a state object. If both the values differ,
```
languy, line 19: This seems like a generally useful utility that we could reuse elsewhere. Can we move it to its own file?
A class would be better suited, so we can call `isDirty()` on it (and possibly pass the `updateCurrent`) parameter when updating the value. I would make it generic so we can pass our own type, rather than be limited to `StatefulValuesType`. We can pass `StatefulValuesType` in this case.

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponentAutoPilotV3.tsx
```diff
@@ -0,0 +1,249 @@
...
+    if (this.props.overrideWithProvisionedThroughputSettings) {
+      return "";
+    } else {
+      return this.props.maxAutoPilotThroughput.current === 0
+        ? ""
+        : this.props.maxAutoPilotThroughput.current.toString();
+    }
+  };
+
+  public render(): JSX.Element {
```
languy, line 100: Break this function into sub-functions that are easier to unit test. Like in Settings, the visible/non visible (as in `flag && <></>`) code paths are probably natural boundaries to break it down. If you think that any of the sub render part is reusable elsewhere, make it a separate component.

src/Explorer/Controls/ThroughputInput/ThroughputInputReact.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    expect(wrapper.exists("#autopilotInput")).toEqual(false);
+  });
+
+  it("autopilot input visible", () => {
+    baseProps.isAutoPilotSelected = true;
+    const wrapper = shallow(<ThroughputInputAutoPilotV3Component {...baseProps} />);
+    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.exists("#autopilotInput")).toEqual(true);
+    expect(wrapper.exists("#throughputInput")).toEqual(false);
+  });
```
languy, line 111: There seems to be more visible/non-visible conditions to test.

languy (approved):

src/Explorer/Controls/Settings/SettingsUtils.tsx
srinathnarayanan: yes, it would still lead to an equivalent comparison.

src/Explorer/Controls/Settings/SettingsComponent.tsx
srinathnarayanan: good catch. there is no scenario in which we would update the baseline and not update current value. WIll remove this flag.

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
```diff
@@ -0,0 +1,372 @@
...
+  getAutoscalePricePerRu,
+  getMultimasterMultiplier,
+  computeRUUsagePriceHourly,
+  getPricePerRu,
+  calculateEstimateNumber
+} from "../../../Utils/PricingUtils";
+import { StatefulValue } from "../StatefulValue";
+import { ITextFieldStyles, ICheckboxStyles, IStackProps, IStackTokens } from "office-ui-fabric-react";
+
+export function getAutoPilotV3SpendElement(
```
languy, line 19: nit: use fat arrow syntax.

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
```diff
@@ -0,0 +1,372 @@
...
+    Enable change feed log retention policy to retain last 10 minutes of history for items in the container by default.
+    To support this, the request unit (RU) charge for this container will be multiplied by a factor of two for writes.
+    Reads are unaffected.
+  </span>
+);
+
+const dirtyTextFieldColor = "#9b4f96";
+
+export const getTextFieldStyles = (statefulValue?: StatefulValue<unknown>): Partial<ITextFieldStyles> => {
+  return {
```
languy, line 332: nit: to return the object, you can just do: ` export const blah = () => ({ field1:"foo", field2: "bar" });`

src/Explorer/Controls/Settings/SettingsSubComponents/ConflictResolutionComponent.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <div className="tab">
+            <Label
+              tabIndex={0}
+              role="radio"
+              className={`settingsV2Label ${this.props.conflictResolutionPolicyMode.isDirty() ? "dirty" : ""} ${
+                this.props.conflictResolutionPolicyMode.current === DataModels.ConflictResolutionMode.LastWriterWins
+                  ? "selectedRadio"
+                  : "unselectedRadio"
+              }`}
+              onKeyPress={this.onConflictResolutionLWWKeyPress}
```
languy, line 77: Can we use office-ui [ChoiceGroup](https://developer.microsoft.com/en-us/fluentui#/controls/web/choicegroup)? Or our [RadioSwitchComponent](https://github.com/Azure/cosmos-explorer/tree/master/src/Explorer/Controls/RadioSwitchComponent)? Or AccessibleElement. It would great if we can avoid creating another component. It is missing aria-label.

src/Explorer/Controls/Settings/SettingsSubComponents/ConflictResolutionComponent.tsx
```diff
@@ -0,0 +1,150 @@
...
+  ) => void;
+  conflictResolutionPolicyProcedure: StatefulValue<string>;
+  onConflictResolutionPolicyProcedureChange: (
+    event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+    newValue?: string
+  ) => void;
+}
+
+export class ConflictResolutionComponent extends React.Component<ConflictResolutionComponentProps> {
+  constructor(props: ConflictResolutionComponentProps) {
```
languy, line 35: Remove constructor.

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      return;
+    },
+    indexingPolicyContent: new StatefulValue<DataModels.IndexingPolicy>({
+      automatic: false,
+      indexingMode: "",
+      includedPaths: [],
+      excludedPaths: []
+    }),
+    setIndexingPolicyElementFocussed: () => {
+      return;
```
languy, line 20: remove

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+  setIndexingPolicyContent: (newIndexingPolicy: DataModels.IndexingPolicy) => void;
+  setIndexingPolicyValidity: (isValid: boolean) => void;
+  logIndexingPolicySuccessMessage: () => void;
+}
+
+export class IndexingPolicyComponent extends React.Component<IndexingPolicyComponentProps> {
+  private indexingPolicyDiv = React.createRef<HTMLDivElement>();
+  private indexingPolicyEditor: monaco.editor.IStandaloneCodeEditor;
+
+  constructor(props: IndexingPolicyComponentProps) {
```
languy, line 20: Remove.

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+      this.props.setIndexingPolicyContent(newIndexingPolicyContent);
+      this.props.setIndexingPolicyValidity(true);
+    } catch (e) {
+      this.props.setIndexingPolicyValidity(true);
+    }
+  };
+
+  public render(): JSX.Element {
+    return (
+      <>
```
languy, line 77: Remove

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+describe("SubSettingsComponent", () => {
+  container.isPreferredApiDocumentDB = ko.computed(() => true);
+
+  const baseProps: SubSettingsComponentProps = {
+    collection: collection,
+    container: container,
+
+    timeToLive: new StatefulValue<TtlType>(TtlType.On),
+    onTtlChange: () => {
+      return;
```
languy, line 18: Remove

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    props.timeToLive.current = TtlType.Off;
+    const wrapper = shallow(<SubSettingsComponent {...props} />);
+    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.exists("#timeToLive")).toEqual(true);
+    expect(wrapper.exists("#timeToLiveSeconds")).toEqual(false);
+  });
+
+  it("analyticalTimeToLive hidden", () => {
+    const props = { ...baseProps };
+    props.isAnalyticalStorageEnabled = false;
```
languy, line 72: FYI, you can do both previous statements in one (but be careful with the order):
`const props = { ...baseProps, isAnalyticalStorageEnabled: false };`

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
```diff
@@ -0,0 +1,443 @@
...
+    this.props.onChangeFeedPolicyChange(ChangeFeedPolicyState.On);
+  };
+
+  private getTtlComponent = (): JSX.Element => {
+    return (
+      <Stack {...titleAndInputStackProps}>
+        <Text>Time to Live</Text>
+        <div className="tabs" id="timeToLive" aria-label="Time to Live" role="radiogroup">
+          <div className="tab">
+            <Label
```
languy, line 187: Same comment about reusing existing multi-choice component if you can.
Otherwise, if it's similar to the previous one, consider creating a new multi-choice component. Add aria-label. Inside that component, consider setting the correct tabIndex (0, 1, 2, etc).

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export enum SettingsV2TabTypes {
+  ScaleTab,
+  ConflictResolutionTab,
+  SubSettingsTab,
+  IndexingPolicyTab
+}
+
+export function hasDatabaseSharedThroughput(collection: ViewModels.Collection): boolean {
```
languy, line 33: nit: use fat arrow syntax

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,122 @@
...
+export function getMinRUs(collection: ViewModels.Collection, container: Explorer): number {
+  const isTryCosmosDBSubscription = (container && container.isTryCosmosDBSubscription()) || false;
+  if (isTryCosmosDBSubscription) {
+    return SharedConstants.CollectionCreation.DefaultCollectionRUs400;
+  }
+
+  const offerContent = collection && collection.offer && collection.offer() && collection.offer().content;
+
+  if (offerContent && offerContent.offerAutopilotSettings) {
+    return 400;
```
languy, line 76: Define this magic number as a constant (probably already defined?)

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const baseRU = SharedConstants.CollectionCreation.DefaultCollectionRUs400;
+
+  const quotaInKb = collection.quotaInfo().collectionSize;
+  const quotaInGb = PricingUtils.usageInGB(quotaInKb);
+
+  const perPartitionGBQuota: number = Math.max(10, quotaInGb / numPartitions);
+  const baseRUbyPartitions: number = ((numPartitions * perPartitionGBQuota) / 10) * 100;
+
+  return Math.max(baseRU, baseRUbyPartitions);
```
languy, line 106: We should unit test this (if not already).

src/Explorer/Controls/StatefulValue.tsx
```diff
@@ -0,0 +1,38 @@
...
+    this.current = baselineValue;
+    this.baseline = baselineValue;
+    this.isValid = true;
+  }
+
+  public isDirty(): boolean {
+    const currentStringValue = this.getStringValue(this.current);
+    const baselineStringValue = this.getStringValue(this.baseline);
+
+    return currentStringValue !== baselineStringValue;
```
languy, line 22: Add unit test?

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
```diff
@@ -0,0 +1,207 @@
+import * as DataModels from "../../../Contracts/DataModels";
```
languy, line 1: Can we keep the filename the same as the class name (ie remove "React")?

src/Explorer/Controls/ThroughputInput/ThroughputInputReact.test.tsx
```diff
@@ -0,0 +1,122 @@
+import { shallow } from "enzyme";
```
languy, line 1: Keep the filename the same as the component that is being tested (ie `ThroughputInputComponent.test.tsx`)

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponentAutoPilotV3.tsx
```diff
@@ -0,0 +1,212 @@
+import React from "react";
```
languy, line 1: Keep filename consistent with class name.

languy (approved):

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.test.tsx
srinathnarayanan: having it defined as "setIndexingPolicyElementFocussedL () => {}" gave linting errors , and "() => void" definition isn't possible.

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.test.tsx
srinathnarayanan: same comment as at the other place.

src/Explorer/Controls/Settings/SettingsUtils.tsx
srinathnarayanan: have a unit test for this.

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
srinathnarayanan: Problem is that the knockout class is of the same name and same file name. WIll move this into SettingsSubComponents instead and rename.

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
srinathnarayanan: The radio buttons dont give the intended styling effect that is currently present AccessibleELement doesn't play well with the tabbing either. will create a new component which can have N choices and accept its effects (onClick, onPress) as an array.

srinathnarayanan: Just synced with Jessica, we decided to move all these elements to vertical radio buttons to be consistent.

src/Explorer/Controls/Settings/SettingsComponent.tsx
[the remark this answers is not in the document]
swviswan: I think it is not needed. We have off the box solutions in React. Let us sync up offline.

---
In reply to: [475792525](https://github.com/Azure/cosmos-explorer/pull/161#discussion_r475792525) [](ancestors = 475792525)

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,1403 @@
...
+  ttlOnFocused: boolean;
+  userCanChangeProvisioningTypes: boolean;
+  selectedAutoPilotTier: DataModels.AutopilotTier;
+  isAutoPilotSelected: boolean;
+  autoPilotThroughput: StatefulValue<number>;
+  wasAutopilotOriginallySet: boolean;
+  selectedTab: SettingsV2TabTypes;
+}
+
+export class SettingsComponent extends React.Component<SettingsComponentProps, SettingsComponentState> {
```
swviswan, line 114: >SettingsComponent [](start = 13, length = 17)

I believe this can be made more cleaner. Let us sync up offline.

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.test.tsx
languy: There's a typo in your comment, I'm assuming you meant `:` instead of `L`.
What linting error are you getting? if it's [typescript no-empty-function](https://eslint.org/docs/rules/no-empty-function), do this: 
```
setIndexingPolicyElementFocussed: () => {
  // do nothing
}
```

src/Explorer/Controls/ThroughputInput/ThroughputInputReactComponent.tsx
languy: Right. We had the same issue before, and we did the same thing (add "React") in the filename so it's ok.

src/Explorer/Controls/Settings/SettingsComponent.tsx
srinathnarayanan: synced offline. removed stateful values.

languy (approved):

src/Explorer/Controls/Settings/SettingsRenderUtils.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+  getThroughputApplyDelayedMessage,
+  getThroughputApplyShortDelayMessage,
+  getThroughputApplyLongDelayMessage,
+  getToolTipContainer,
+  conflictResolutionCustomToolTip,
+  changeFeedPolicyToolTip
+} from "./SettingsRenderUtils";
+import { AutopilotTier } from "../../../Contracts/DataModels";
+
+class SettingsRenderUtilsTestComponent extends React.Component {
```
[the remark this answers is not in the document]
srinathnarayanan: haha thanks

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
```diff
@@ -0,0 +1,399 @@
...
+    <span>
+      Estimated monthly cost ({currency}):{" "}
+      <b>
+        {currencySign}
+        {calculateEstimateNumber(monthlyPrice / 10)}
+        {` - `}
+        {currencySign}
+        {calculateEstimateNumber(monthlyPrice)}{" "}
+      </b>
+      ({regions} {regions === 1 ? "region" : "regions"}, {throughput / 10} - {throughput} RU/s, {currencySign}
```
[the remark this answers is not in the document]
srinathnarayanan: changed it.

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
```diff
@@ -0,0 +1,399 @@
...
+);
+
+export const updateThroughputDelayedApplyWarningMessage: JSX.Element = (
+  <span>
+    You are about to request an increase in throughput beyond the pre-allocated capacity. This operation will take some
+    time to complete.
+  </span>
+);
+
+const getCurrentThroughput = (
```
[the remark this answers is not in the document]
srinathnarayanan: it is tested as part of SettingsRenderUtils.test

src/Explorer/Controls/Settings/SettingsSubComponents/IndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  private onEditorContentChange = (): void => {
+    const indexingPolicyEditorModel = this.indexingPolicyEditor.getModel();
+    try {
+      const newIndexingPolicyContent = JSON.parse(indexingPolicyEditorModel.getValue()) as DataModels.IndexingPolicy;
+      this.props.setIndexingPolicyContent(newIndexingPolicyContent);
+      this.props.setIndexingPolicyValidity(true);
+    } catch (e) {
+      this.props.setIndexingPolicyValidity(true);
```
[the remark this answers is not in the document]
srinathnarayanan: no, it should be false, thanks for catching it.

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,138 @@
...
+  if (isTryCosmosDBSubscription) {
+    return Constants.TryCosmosExperience.maxRU;
+  }
+
+  const numPartitionsFromOffer: number =
+    collection?.offer && collection.offer()?.content?.collectionThroughputInfo?.numPhysicalPartitions;
+
+  const numPartitionsFromQuotaInfo: number = collection?.quotaInfo().numPartitions;
+
+  const numPartitions = numPartitionsFromOffer || numPartitionsFromQuotaInfo || 1;
```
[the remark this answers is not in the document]
srinathnarayanan: yes that is intentional

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  return SharedConstants.CollectionCreation.MaxRUPerPartition * numPartitions;
+};
+
+export const getMinRUs = (collection: ViewModels.Collection, container: Explorer): number => {
+  const isTryCosmosDBSubscription = container?.isTryCosmosDBSubscription() || false;
```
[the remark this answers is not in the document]
srinathnarayanan: removed it.

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return SharedConstants.CollectionCreation.DefaultCollectionRUs400;
+  }
+
+  const collectionThroughputInfo: DataModels.OfferThroughputInfo = offerContent?.collectionThroughputInfo;
+
+  if (collectionThroughputInfo?.minimumRUForCollection > 0) {
+    return collectionThroughputInfo.minimumRUForCollection;
+  }
+
+  const numPartitions = collectionThroughputInfo?.numPhysicalPartitions || collection.quotaInfo().numPartitions;
```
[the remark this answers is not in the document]
srinathnarayanan: number of partitions should never be 0, so this shouldn't be a problem.

src/Explorer/Tree/Collection.ts
```diff
@@ -559,33 +561,51 @@ export default class Collection implements ViewModels.Collection {
...
+      collectionName: this.id(),
+      defaultExperience: this.container.defaultExperience(),
+      dataExplorerArea: Constants.Areas.Tab,
+      tabTitle: tabTitle
+    };
+
+    const settingsTabOptions: ViewModels.TabOptions = {
+      tabKind: undefined,
+      title: !this.offer() ? "Settings" : "Scale & Settings",
+      tabPath: "",
```
swviswan, line 576: >tabPath [](start = 6, length = 7)

nit: why blank line?

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,1285 @@
...
+            hasAutoPilotV2FeatureFlag={this.hasAutoPilotV2FeatureFlag}
+            isFixedContainer={this.isFixedContainer}
+            autoPilotTiersList={this.autoPilotTiersList}
+            onThroughputChange={this.onThroughputChange}
+            throughput={this.state.throughput}
+            throughputBaseline={this.state.throughputBaseline}
+            autoPilotThroughput={this.state.autoPilotThroughput}
+            autoPilotThroughputBaseline={this.state.autoPilotThroughputBaseline}
+            selectedAutoPilotTier={this.state.selectedAutoPilotTier}
+            isAutoPilotSelected={this.state.isAutoPilotSelected}
```
swviswan, line 1172: Instead of passing so many props, can you pass on scale prop object which has these values?

src/Explorer/Tree/Collection.ts
srinathnarayanan: will remove it.

src/Explorer/Controls/Settings/SettingsComponent.tsx
srinathnarayanan: I dont think that will streamline it much. a ScaleComponentProps interface already exists and these props being passed are a part of that.

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -0,0 +1,1285 @@
...
+  isVisible: () => boolean;
+  isEnabled: () => boolean;
+  isSelected?: () => boolean;
+}
+
+export interface SettingsComponentProps {
+  settingsTab: SettingsTab;
+}
+
+export interface SettingsComponentState {
```
swviswan, line 68: >SettingsComponentState [](start = 17, length = 22)

As discussed, let us have 4 values; scaleUpdated, indexingPolicyUpdated, ConflictresolnUpdated etc which this component will use to determine in enabling or disabling save /discard buttons.  And the associated logic for scaleUpdated is all within the scale component.

vignesh-mr (approved):

src/Explorer/Controls/Settings/SettingsUtils.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    expect(parseConflictResolutionMode("custom")).toEqual(DataModels.ConflictResolutionMode.Custom);
+    expect(parseConflictResolutionMode("lastwriterwins")).toEqual(DataModels.ConflictResolutionMode.LastWriterWins);
+  });
+
+  it("parseConflictResolutionProcedure", () => {
+    expect(parseConflictResolutionProcedure("/dbs/db/colls/coll/sprocs/conflictResSproc")).toEqual("conflictResSproc");
+    expect(parseConflictResolutionProcedure("conflictResSproc")).toEqual("conflictResSproc");
+  });
+
+  describe("isDirty", () => {
```
southpolesteve, line 68: Jest.each would be a good fit for these tests https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#jest-each

Here is an example https://github.com/Azure/cosmos-explorer/blob/401ee7af2cb61c1f012cfe21d637f416e9ce843d/src/Utils/MessageValidation.test.ts

southpolesteve (approved):

PR #164, opened 2020-08-22
Add analytical store schema POC

ausfeldt:


src/Explorer/Tree/Database.ts
```diff
@@ -388,4 +396,39 @@ export default class Database implements ViewModels.Database {
   private _getOfferForDatabase(offers: DataModels.Offer[], database: DataModels.Database): DataModels.Offer {
     return _.find(offers, (offer: DataModels.Offer) => offer.resource === database._self);
   }
+
+  private addSchema(collection: DataModels.Collection): void {
+    if (collection.analyticalStorageTtl == undefined || !this.container.isSchemaEnabled()) {
```
languy, line 401: Is there a particular reason you use `== undefined` instead of `=== undefined`?

src/Explorer/Tree/Database.ts
```diff
@@ -388,4 +396,39 @@ export default class Database implements ViewModels.Database {
...
+          userContext.databaseAccount.name,
+          this.id(),
+          collection.id
+        );
+
+        if (response.status >= 404) {
+          clearInterval(checkForSchema);
+        }
+
+        if (response.data != null) {
```
languy, line 425: Is there a particular reason to use `!= null` rather than `!== null`?

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -245,6 +246,11 @@ export class ResourceTreeAdapter implements ReactAdapter {
       isSelected: () => this.isDataNodeSelected(collection.rid, "Collection", ViewModels.CollectionTabKind.Settings)
     });
 
+    let schemaNode: TreeNode = this.buildSchemaNode(collection);
```
languy, line 249: `const schemaNode`. And with `const` we usually drop the type and let the compiler infer unless you want to be verbose.

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
     };
   }
 
+  private buildSchemaNode(collection: ViewModels.Collection): TreeNode {
+    if (collection.analyticalStorageTtl() == undefined) {
+      return null;
```
languy, line 362: Use `undefined` when possible. We are trying to remove all `null`'s from the code (because it's redundant with `undefined` and we have to check for both. Steve will turn on a lint rule for it).

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+    }
+
+    if (
+      collection.schema == undefined ||
```
languy, line 366: I think you can do:
`if (!collection.schema || !collection.schema.fields) { return undefined; }`

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+        collection.selectedSubnodeKind(ViewModels.CollectionTabKind.Schema);
+        this.container.tabsManager.refreshActiveTab(
+          (tab: TabsBase) => tab.collection && tab.collection.rid === collection.rid
+        );
+      }
+    };
+  }
+
+  private getSchemaNodes(fields: DataModels.IDataField[]): TreeNode[] {
+    let schema: any = {};
```
languy, line 387: `const`.

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+
+    //unflatten
+    fields.forEach((field: DataModels.IDataField, fieldIndex: number) => {
```
languy, line 390: nit: no need to specify type for `field` and `number` unless you want to.

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+      const path: string[] = field.path.split(".");
+      const fieldProperties = [field.dataType.name, `HasNulls: ${field.hasNulls}`];
+      let current: any = {};
+      path.forEach((name: string, pathIndex: number) => {
+        if (pathIndex == 0) {
+          if (schema[name] == undefined) {
+            if (pathIndex == path.length - 1) {
```
languy, line 397: `===`

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+  private getSchemaNodes(fields: DataModels.IDataField[]): TreeNode[] {
+    let schema: any = {};
+
+    //unflatten
+    fields.forEach((field: DataModels.IDataField, fieldIndex: number) => {
+      const path: string[] = field.path.split(".");
+      const fieldProperties = [field.dataType.name, `HasNulls: ${field.hasNulls}`];
+      let current: any = {};
+      path.forEach((name: string, pathIndex: number) => {
+        if (pathIndex == 0) {
```
languy, line 395: `===`

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+            } else {
+              current[name] = {};
+            }
+          }
+          current = current[name];
+        }
+      });
+    });
+
+    function traverse(obj: any): TreeNode[] {
```
languy, line 417: Prefer the fat arrow syntax: `const traverse = (obj: any): TreeNode[] => {}`

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+      let children: TreeNode[] = [];
```
languy, line 418: `const` if children isn't going to be reassigned.

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -351,6 +357,80 @@ export class ResourceTreeAdapter implements ReactAdapter {
...
+
+      if (obj !== null && !Array.isArray(obj) && typeof obj == "object") {
```
languy, line 420: `=== "object"`

src/Explorer/Tree/ResourceTreeAdapter.tsx
languy: Consider moving this out of the class or convert to a static function to enable unit testing.

languy (approved):
Consider adding some unit tests if you think this POC code is going to stick.

src/Contracts/DataModels.ts
```diff
@@ -92,6 +92,38 @@ export interface ResourceRequest {
   id: string;
 }
 
+export interface IType {
```
languy, line 95: If possible, move these definitions in the file where it is being used. We are trying to break this monolithic pattern of putting everything in `DataModels` and `ViewModels`.

languy (approved):

src/Explorer/Tree/Database.ts
ausfeldt: nope, c# muscle memory .

src/Explorer/Tree/Database.ts
ausfeldt: nope, c# muscle memory.

src/Contracts/DataModels.ts
ausfeldt: I initially tried putting everything in JunoClient but then found I needed them in other places. I put them into DataModels as that seem to be the established pattern :p

src/Explorer/Tree/ResourceTreeAdapter.tsx
ausfeldt: Integration test added to ResourceTreeAdapter.test.tsx which covers all within getSchemaNodes including traverse. This is just a POC and depending on what sticks and what does not stick, I refactor afterward when the dust settles to make it unit test friendly.

languy: SGTM 👍

southpolesteve (approved):

PR #166, opened 2020-08-25
Move createDatabase to RP

victor-meng:
- Refactored and moved all code paths for creating database in different APIs into createDatabase.ts.
- Before creating a database with RP, we need to check whether there already exists a database with the same id by using the getDatabase call. This is because RP bundles the create call with the update call which means if we try to create a database that already exists, RP will update the existing database instead of creating a new one.
- If there's no existing database with the same id, then RP returns 404 for the get call, so we need to add a try catch block to intercept the error and check if the error code is `"NotFound"`

### DIFF
--- a/src/Common/dataAccess/createDatabase.ts
+++ b/src/Common/dataAccess/createDatabase.ts
@@ -1,4 +1,3 @@
-import * as Constants from "../Constants";
 import * as DataModels from "../../Contracts/DataModels";
 import { AuthType } from "../../AuthType";
 import { DatabaseResponse } from "@azure/cosmos";
@@ -152,7 +151,7 @@ async function createCassandraKeyspace(params: DataModels.CreateDatabaseParams):
       userContext.databaseAccount.name,
       params.databaseId
     );
-    if (getResponse && getResponse.properties && getResponse.properties.resource) {
+    if (getResponse?.properties?.resource) {
       throw new Error(`Create database failed: database with id ${params.databaseId} already exists`);
     }
   } catch (error) {
@@ -222,11 +221,9 @@ async function createDatabaseWithSDK(params: DataModels.CreateDatabaseParams): P
   // TODO: replace when SDK support autopilot
   if (params.databaseLevelThroughput) {
     if (params.autoPilotMaxThroughput) {
-      databaseOptions.initialHeaders = {
-        [Constants.HttpHeaders.autoPilotThroughputSDK]: JSON.stringify({ maxThroughput: params.autoPilotMaxThroughput })
-      };
+      createBody.maxThroughput = params.autoPilotMaxThroughput;
     } else {
-      createBody.throughput = params.autoPilotMaxThroughput;
+      createBody.throughput = params.offerThroughput;
     }
   }
 
@@ -241,7 +238,9 @@ function constructRpOptions(params: DataModels.CreateDatabaseParams): CreateUpda
 
   if (params.autoPilotMaxThroughput) {
     return {
-      [Constants.HttpHeaders.autoPilotThroughput]: { maxThroughput: params.autoPilotMaxThroughput }
+      autoscaleSettings: {
+        maxThroughput: params.autoPilotMaxThroughput
+      }
     };
   }
 
--- a/src/Common/dataAccess/createDatabase.ts
+++ b/src/Common/dataAccess/createDatabase.ts
@@ -1,0 +1,251 @@
+import * as Constants from "../Constants";
+import * as DataModels from "../../Contracts/DataModels";
+import { AuthType } from "../../AuthType";
+import { DatabaseResponse } from "@azure/cosmos";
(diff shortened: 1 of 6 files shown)


src/Common/dataAccess/createDatabase.ts
```diff
@@ -0,0 +1,251 @@
...
+    params.databaseId,
+    rpPayload
+  );
+  return createResponse && (createResponse.properties.resource as DataModels.Database);
+}
+
+async function createDatabaseWithSDK(params: DataModels.CreateDatabaseParams): Promise<DataModels.Database> {
+  const createBody: DatabaseRequest = { id: params.databaseId };
+  const databaseOptions: RequestOptions = {};
+  // TODO: replace when SDK support autopilot
```
southpolesteve, line 222: This should work now. maxThroughput is available first class in the SDK. Ping @zfoster if you hit any trouble

src/Common/dataAccess/createDatabase.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+function constructRpOptions(params: DataModels.CreateDatabaseParams): CreateUpdateOptions {
+  if (!params.databaseLevelThroughput) {
+    return {};
+  }
+
+  if (params.autoPilotMaxThroughput) {
+    return {
+      [Constants.HttpHeaders.autoPilotThroughput]: { maxThroughput: params.autoPilotMaxThroughput }
```
southpolesteve, line 244: Same as above, I think you don't need to set this header anymore.

src/Common/dataAccess/createDatabase.ts
```diff
@@ -0,0 +1,251 @@
...
+  createUpdateGremlinDatabase,
+  getGremlinDatabase
+} from "../../Utils/arm/generatedClients/2020-04-01/gremlinResources";
+import { logConsoleProgress, logConsoleError, logConsoleInfo } from "../../Utils/NotificationConsoleUtils";
+import { logError } from "../Logger";
+import { refreshCachedOffers, refreshCachedResources } from "../DataAccessUtilityBase";
+import { sendNotificationForError } from "./sendNotificationForError";
+import { userContext } from "../../UserContext";
+
+export async function createDatabase(params: DataModels.CreateDatabaseParams): Promise<DataModels.Database> {
```
languy, line 32: Can we switch from the old-school `function` syntax to the fat arrow syntax:
`export const createDatabase = async () => { }`;

languy (approved):

src/Common/dataAccess/createDatabase.ts
victor-meng: This is actually a bug. I'm setting the RP options for creating database so I shouldn't be using the header string. Fixed in my latest commit.

victor-meng:
> After this change is `AddDatabaseUtility` used anymore? If not, should delete it.

@southpolesteve Some of the functions in there are being used but we should be able to clean them up. I can do this in another PR. I think we should be able to remove `AddCollectionUtility` too after we move create collection to RP. I'll probably clean up both utilities in the same PR.

src/Common/dataAccess/createDatabase.ts
```diff
@@ -0,0 +1,251 @@
...
+  getMongoDBDatabase
+} from "../../Utils/arm/generatedClients/2020-04-01/mongoDBResources";
+import {
+  createUpdateGremlinDatabase,
+  getGremlinDatabase
+} from "../../Utils/arm/generatedClients/2020-04-01/gremlinResources";
+import { logConsoleProgress, logConsoleError, logConsoleInfo } from "../../Utils/NotificationConsoleUtils";
+import { logError } from "../Logger";
+import { refreshCachedOffers, refreshCachedResources } from "../DataAccessUtilityBase";
+import { sendNotificationForError } from "./sendNotificationForError";
```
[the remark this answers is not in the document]
victor-meng: I will update all of them in a separate PR. Steve suggested that we could add a lint rule for this.

src/Common/dataAccess/createDatabase.ts
victor-meng: I'll make the switch in the same PR where I update the module names.

src/Common/dataAccess/createDatabase.ts
```diff
@@ -0,0 +1,250 @@
...
+      throw new Error(`Create database failed: database with id ${params.databaseId} already exists`);
+    }
+  } catch (error) {
+    if (error.code !== "NotFound") {
+      throw error;
+    }
+  }
+
+  const options: CreateUpdateOptions = constructRpOptions(params);
+  const rpPayload: SqlDatabaseCreateUpdateParameters = {
```
ato9000, line 92: nit: if all APIs use the same parameters, consider renaming this interface to DatabaseCreateUpdateParameters

victor-meng: This is actually a bug. Different APIs should have different parameter types. I'll update this in my next PR. Thanks for catching it!

PR #168, opened 2020-08-25
Start kernel when opening notebook tab

languy:
Code completion is not working unless a kernel is running, so we automatically start the kernel by calling `restartKernel` (which doesn't require any kernel parameter).

### DIFF
--- a/src/Explorer/Notebook/NotebookComponent/NotebookComponentBootstrapper.tsx
+++ b/src/Explorer/Notebook/NotebookComponent/NotebookComponentBootstrapper.tsx
@@ -98,7 +98,7 @@ export class NotebookComponentBootstrapper {
       actions.fetchContentFulfilled({
         filepath: undefined,
         model: NotebookComponentBootstrapper.wrapModelIntoContent(name, undefined, content),
-        kernelRef: createKernelRef(),
+        kernelRef: undefined, // must be undefined or it will be auto-started by the epic
         contentRef: this.contentRef
       })
     );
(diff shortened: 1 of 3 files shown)


languy:
> I remember @victor-meng added support for lazy launching kernels intentionally before, do we know why we did that before? This will undo whatever that was achieving. I think one idea was to support opening notebooks without needing a container in the future.

Yes, it is partly correct. 

The primary purpose for implementing lazy launching kernels was to prevent nteract from auto-starting a kernel, so that we could control when to start it. It was specifically intended for the read-only notebook scenario which, presumably, would never be "executable" (at least for now). So this is still the case with this change: the standalone notebook viewer as well as the notebook gallery viewer are not starting any kernel. The mechanism is kind of subtle (but documented [here](https://github.com/Azure/cosmos-explorer/blob/431bab9d72e5219bec0236319a3b6316c87ba6c9/src/Explorer/Notebook/NotebookComponent/epics.ts#L99)): if no `kernelRef` is specified for a given `contentRef`, don't start any kernel.

As you mention, another benefit in the "normal" portal use case is that it saves container resources by not starting a new kernel for every single open notebook. In the short-term, in our current implementation a container is still created regardless, so spawning a new process for the kernel may just have low impact: CPU usage is low (if not executing), it might use some memory. 
The immediate goal for this change seems more important: fix the regression and enable again code completion which requires the kernel to be running (it's the kernel that provides the options in the dropdown via the websocket channel).

A possible improvement would be to keep the lazy-start on execution, but also lazy-start upon editing (ie when the user starts typing code or some other event that hints the user will be editing code). The challenge is to ensure the kernel starts quickly in order to respond to the `complete_request` command on time. In the current implementation, if the websocket isn't established prior to the request, it will simply fail. Or we write a bunch of code to hold it until it's ready. This optimization would probably degrade the UX in some cases, so it seems premature at this time. By then, we may go for a more elaborate language service that doesn't require a running kernel. Lazy-starting the container poses the same challenge.

For now, since code-completion is important for a good editing UX, this seems like the simplest path.

PR #171, opened 2020-08-27
Screen reader changes

DanielSPham:
Removing alt text for decorative images.

### DIFF
--- a/src/Explorer/SplashScreen/SplashScreenComponent.tsx
+++ b/src/Explorer/SplashScreen/SplashScreenComponent.tsx
@@ -45,7 +45,7 @@ export class SplashScreenComponent extends React.Component<SplashScreenComponent
                 tabIndex={0}
                 role="button"
               >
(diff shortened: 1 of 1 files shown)


src/Explorer/SplashScreen/SplashScreenComponent.tsx
```diff
@@ -45,7 +45,7 @@ export class SplashScreenComponent extends React.Component<SplashScreenComponent
                 tabIndex={0}
                 role="button"
               >
-                <img src={item.iconSrc} alt={item.title} />
+                <img src={item.iconSrc} alt="" />
```
languy, line 48: Can you explain the purpose of the change? I thought this was required for those who can't see the image.

DanielSPham: Screen reader is reading this twice, once for the image alt and once for the button content. Since the image is acting as a button icon, we should remove the alt text.

PR #177, opened 2020-09-02
Support showing only notebook output scenario

languy:
Hide cell prompts when hiding cell inputs. This enables us to render a notebook's outputs only.
Imported the Prompt code from nteract in order to conditionally render it in the stateful component.


src/Explorer/Notebook/NotebookRenderer/OriginalPrompt.tsx
```diff
@@ -0,0 +1,66 @@
+// This is the original read-only [] prompt
```
gitanuj, line 1: Can we use Prompt directly from Nteract? (instead of copying it)

src/Explorer/Controls/NotebookViewer/NotebookViewerComponent.tsx
```diff
@@ -30,6 +30,7 @@ export interface NotebookViewerComponentProps {
   isFavorite?: boolean;
   backNavigationText: string;
   hideInputs?: boolean;
+  hidePrompt?: boolean;
```
gitanuj, line 33: nit:
```suggestion
  hidePrompts?: boolean;
```

src/Explorer/Notebook/NotebookRenderer/NotebookReadOnlyRenderer.tsx
```diff
@@ -3,6 +3,8 @@ import "./base.css";
 import "./default.css";
 
 import { CodeCell, RawCell, Cells, MarkdownCell } from "@nteract/stateful-components";
+import Prompt, { PassedPromptProps } from "./OriginalPrompt";
+// import Prompt, { PassedPromptProps } from "@nteract/stateful-components/src/inputs/prompt";
```
gitanuj, line 7: can be removed

src/Explorer/Notebook/NotebookRenderer/OriginalPrompt.tsx
languy: Good catch! I was having trouble importing because I was using the wrong src location (see your comment below). It now works with the correct imported package.

src/Explorer/Notebook/NotebookRenderer/NotebookReadOnlyRenderer.tsx
languy: That was the wrong package location.

src/Explorer/Controls/NotebookViewer/NotebookViewerComponent.tsx
languy: Fixed.

srinathnarayanan (approved):

southpolesteve (approved):

PR #178, opened 2020-09-02
Use edge instead of chrome for emulator e2e test

victor-meng:
The emulator e2e test is failing with the following error:

> Can't run because you've entered an invalid browser name.
> 
> Browser: 'chrome' was not found on your system or is not supported by Cypress.
> 
> Cypress supports the following browsers:
> - chrome
> - chromium
> - edge
> - electron
> - firefox (Cypress support in beta)
> 
> You can also use a custom browser: https://on.cypress.io/customize-browsers
> 
> Available browsers found on your system are:
> - firefox
> - edge
> - electron

This fix will change the browser from chrome to edge for the emulator e2e test temporarily while we investigate why chrome is missing from our system.


zfoster (approved):
just ran into this myself

southpolesteve (approved):

PR #182, opened 2020-09-04
Telemetry Adjustments

southpolesteve:
- Enables telemetry for fetch calls
- Removes "LoadingStatus" event from telemetry which was causing a lot of noise
- Add linting to telemetry files

### DIFF
--- a/src/Shared/Telemetry/TelemetryProcessor.ts
+++ b/src/Shared/Telemetry/TelemetryProcessor.ts
@@ -8,118 +8,118 @@ import { userContext } from "../../UserContext";
 /**
  * Class that persists telemetry data to the portal tables.
  */
-// TODO: Move to a separate Diagnostics folder
-// TODO: Log telemetry for StorageExplorer case/other clients as well
-export default class TelemetryProcessor {
-  public static trace(action: Action, actionModifier: string = ActionModifiers.Mark, data?: any): void {
-    sendMessage({
-      type: MessageTypes.TelemetryInfo,
-      data: {
-        action: Action[action],
-        actionModifier: actionModifier,
-        data: JSON.stringify(data)
-      }
-    });
 
-    appInsights.trackEvent({ name: Action[action] }, TelemetryProcessor.getData(data));
-  }
-
-  public static traceStart(action: Action, data?: any): number {
-    const timestamp: number = Date.now();
-    sendMessage({
-      type: MessageTypes.TelemetryInfo,
-      data: {
-        action: Action[action],
-        actionModifier: ActionModifiers.Start,
-        timestamp: timestamp,
-        data: JSON.stringify(data)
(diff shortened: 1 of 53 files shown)


notlaforge (approved):

src/Shared/Telemetry/TelemetryProcessor.ts
```diff
@@ -8,123 +8,124 @@ import { userContext } from "../../UserContext";
...
-        actionModifier: ActionModifiers.Start,
-        timestamp: timestamp,
-        data: JSON.stringify(data)
-      }
-    });
-
-    appInsights.startTrackEvent(Action[action]);
-    return timestamp;
-  }
+export function trace(action: Action, actionModifier: string = ActionModifiers.Mark, data?: unknown): void {
```
languy, line 12: Can we use the fat arrow syntax?
```suggestion
export const trace = (action: Action, actionModifier: string = ActionModifiers.Mark, data?: unknown): void => {
```

src/Shared/Telemetry/TelemetryProcessor.ts
```diff
@@ -8,123 +8,124 @@ import { userContext } from "../../UserContext";
...
+      timestamp: timestamp || Date.now(),
+      data: JSON.stringify(data)
+    }
+  });
+
+  appInsights.stopTrackEvent(Action[action], getData(data));
+}
+
+export function traceOpen(action: Action, data?: unknown, timestamp?: number): number {
+  const validTimestamp = timestamp || Date.now();
```
languy, line 84: I was going to suggest nullish coalescing here `timestamp && Date.now()`, but a timestamp of 0 is probably not valid, so this is intentional?

languy (approved):

src/Shared/appInsights.ts
```diff
@@ -1,7 +1,8 @@
 import { ApplicationInsights } from "@microsoft/applicationinsights-web";
 const appInsights = new ApplicationInsights({
   config: {
-    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49"
+    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49",
+    disableFetchTracking: false
```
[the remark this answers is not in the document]
southpolesteve: Correct. It is disabled by default in app insights. I have no idea why.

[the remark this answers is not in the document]
southpolesteve: Correct. Its disable by default. No idea why they would do that.

src/Shared/Telemetry/TelemetryProcessor.ts
southpolesteve: I think so. This was existing code, just formatting changes when the class got removed.

PR #184, opened 2020-09-04
Fixed contrast ratio issues

DanielSPham:
Changed border / text colors to match fabric ui and increase contrast ratio.

Before
before

After
after


southpolesteve (approved):

jessica-tsui:
@DanielSPham thanks for adding me as an approver. Where did you get the color code: 605e5c from? I checked text fields styles in fluent and the border color seems to be #323130 when it's not in focus, and #0078D4 when it's in focus. Here is the link for reference: https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield

DanielSPham:
@jessica-tsui, it's #605e5c for both semantic colors inputBorder / focusBorder for the theme that's currently used by the data explorer.

languy (approved):

PR #186, opened 2020-09-09
Add temp workaround for ARM missing  parition key information

southpolesteve:


languy (approved):

southpolesteve:
This is a bit of a hack, so going to hold of until RP returns the correct result. Can bring back if we see more CRIs

PR #191, opened 2020-09-10
Refactor MemoryTracker to use SWR

southpolesteve:


src/Explorer/Menus/CommandBar/MemoryTrackerComponent.tsx
```diff
@@ -1,50 +1,71 @@
...
+      userContext.subscriptionId,
+      userContext.resourceGroup,
+      userContext.databaseAccount.name,
+      "default"
+    ],
+    (_key, subscriptionId, resourceGroup, accountName, workspace) =>
+      listConnectionInfo(subscriptionId, resourceGroup, accountName, workspace)
+  );
+  const { data } = useSWR<any>(connectionInfo ? ["memoryUsage", connectionInfo] : null, fetchMemoryInfo, {
+    refreshInterval: 2000
```
gitanuj, line 48: Love how declarative this is! No scheduled tasks for refreshing this data.

gitanuj (approved):
:shipit:

src/Explorer/Menus/CommandBar/MemoryTrackerComponent.tsx
victor-meng, line 48: I think the current interval is 5000. 2000 seems a bit too short.

src/Explorer/Menus/CommandBar/MemoryTrackerComponent.tsx
```diff
@@ -1,50 +1,71 @@
...
+const fetchMemoryInfo = async (_key: unknown, connectionInfo: NotebookWorkspaceConnectionInfoResult) => {
+  const response = await fetch(`${connectionInfo.notebookServerEndpoint}/api/metrics/memory`, {
+    method: "GET",
+    headers: {
+      Authorization: `Token ${connectionInfo.authToken}`,
+      "content-type": "application/json"
     }
+  });
+  if (!response.ok) {
+    throw new Error(await response.text());
```
victor-meng, line 26: We should also try to restart the notebook server if the call fails. See `onConnectionLost` in NotebookContainerClient.

src/Explorer/Menus/CommandBar/MemoryTrackerComponent.tsx
```diff
@@ -1,50 +1,71 @@
-import * as React from "react";
-import { Observable, Subscription } from "knockout";
-import { MemoryUsageInfo } from "../../../Contracts/DataModels";
+import React, { FunctionComponent } from "react";
+import useSWR from "swr";
```
swviswan, line 2: >useSWR from "swr"; [](start = 7, length = 18)

While I dont think I have specific concerns with this PR, as you know, i would prefer we all talk about and agree on the framework we are adopting. We have a meeting setup early next week to discuss adopting redux/any other framework. If we want to change the philosophy and use swr where every component gets its data instead of data being passed as props, i want the same pattern to be used all across our code base. I think capacity calculator already uses redux. 

I don't want multiple patterns/frameworks dispersed in the code. Hope you agree.

southpolesteve: Yep I agree. This is intended to be a jumping off point for such a discussion. I do think it can be helpful to have actual usage to talk about pros/cons. Long term I certainly am against multiple approaches, but I am OK with trying out multiple approaches in production on a shorter time frame.

southpolesteve:
Going to try out SWR on a different component.

PR #192, opened 2020-09-11
Add telemetry for OpenTerminal success/failure

gitanuj:


southpolesteve (approved):

src/Common/MessageHandler.ts
```diff
@@ -48,16 +48,29 @@ export function sendCachedDataMessage<TResponseDataModel>(
...
+        dataExplorerWindow.document.referrer
+      );
+    }
   }
 }
 
+function getDataExplorerWindow(): Window {
+  // Traverse up the window to find a window with `dataExplorerPlatform` property
+  let dataExplorerWindow: Window = window;
+  while (!dataExplorerWindow.hasOwnProperty("dataExplorerPlatform") && dataExplorerWindow.parent) {
```
vignesh-mr, line 67: Will this break if we rename/remove this property at some point? I wonder if it's easier to just look for the dataExplorer global var instead.

vignesh-mr (approved):

src/Common/MessageHandler.ts
languy: nit: `const getDataExplorerWindow = (): Window => {`

languy: So this is synchronously waiting for `dataExplorerPlatform`: can we add some sleep and timeout with error message (ideally to yield the hand so the ui thread can process other events and avoid possible 100% cpu-type situation)?
Or consider refactoring to async polling. Something like:
`setTimeout(() => pollForProperty(), 200)` 
and 

```
const pollForProperty = () => {
if (!dataExplorerWindow.hasOwnProperty("dataExplorerPlatform") && dataExplorerWindow.parent) {
   if (_we_ve_been_waiting_for_too_long_) {
      // Display error and end execution.
     return;
   }
  // otherwise schedule next poll
setTimeout(() => pollForProperty(), 200);
return;
}
// We've good! continue execution
onDataExplorerWindowReady();
}
```

languy (approved):

src/Common/MessageHandler.ts
gitanuj: @languy There is not waiting for `dataExplorerPlatform`. This code just traverses up the window hierarchy to get the first window which contains `dataExplorerPlatform` property, so we don't need any timeouts/polling.

gitanuj: @vignesh-mr I'll change it to dot syntax instead so that we can at least get a compilation error

PR #195, opened 2020-09-11
Adds cassandra e2e container CRUD test

zfoster:
- [x] Need to fix the error to take a screenshot to a named file based on the test name, suggestion from @languy 

Adds Cassandra version of the SQL Container crud spec in puppeteer/jest


test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
+import "expect-puppeteer";
+import crypto from 'crypto'
```
languy, line 2: nit: `"`. We seem to use `"` to enclose string in Data Explorer. I wonder if our lint and format build targets can apply to these files as well.

test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      await frame.waitFor('input[class="textfontclr"]');
+      await frame.type('input[class="textfontclr"]', tableId); 
+
+      // click submit
+      await frame.waitFor('#cassandraaddcollectionpane > div > form > div.paneFooter > div > input');
+      await frame.click('#cassandraaddcollectionpane > div > form > div.paneFooter > div > input');
+
+      // open database menu
+      await frame.waitFor(`div[data-test="${keyspaceId}"]`);
+      await frame.waitForSelector('div[class="splashScreen"] > div[class="title"]', { visible: true });
```
languy, line 42: FYI, I've seen future designs where splash screen might have it's own tab. Consider relying on command bar or resource tree which might be more reliable (but don't get our of your way is splash screen is easier, we'll see this change coming).

test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      await frame.waitFor('input[id="keyspace-id"]', { visible: true });
+      await frame.type('input[id="keyspace-id"]', keyspaceId);      
+
+      // type table id
+      await frame.waitFor('input[class="textfontclr"]');
+      await frame.type('input[class="textfontclr"]', tableId); 
+
+      // click submit
+      await frame.waitFor('#cassandraaddcollectionpane > div > form > div.paneFooter > div > input');
+      await frame.click('#cassandraaddcollectionpane > div > form > div.paneFooter > div > input');
```
languy, line 38: Is there a way to select that input without relying on the intermediary elements/classes? (which will break if someone changes the layout).

test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      await frame.waitForSelector('body > div.ms-Layer.ms-Layer--fixed');
+      const elements = await frame.$$('span[class="treeComponentMenuItemLabel"]')
+      await elements[0].click()
+
+      // confirm delete container
+      await frame.type('input[data-test="confirmCollectionId"]', tableId.trim());
+
+      // click delete
+      await frame.click('input[data-test="deleteCollection"]');
+      await frame.waitFor(5000);
```
languy, line 63: Let's define all these as constants. Reuse the same constant if possible for the all the `waitFor`. We make it one dial in case we want to fine tune and speed up the test (by decreasing the constant).
Here I noticed that we wait for selector, do we actually need the `waitFor` or can we reduce to 1s (presumably for the ui to refresh to some state?). I'm just concerned that once we have many of these tests it will take for ever...

languy (approved):

test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      await frame.waitFor(`span[title="${tableId}"]`);
+      
+      // delete container
+
+      // click context menu for container
+      await frame.waitFor(`div[data-test="${tableId}"] > div > button`);
+      await frame.click(`div[data-test="${tableId}"] > div > button`);
+
+      // click delete container
+      await frame.waitForSelector('body > div.ms-Layer.ms-Layer--fixed');
```
languy, line 54: I would be very aggressive in tagging what we need for these tests and we don't accidentally break them when refactoring things.

test/cassandra/container.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      await frame.waitFor(`div[data-test="${keyspaceId}"]`);
+      await frame.waitForSelector('div[class="splashScreen"] > div[class="title"]', { visible: true });
+
+      await frame.click(`div[data-test="${keyspaceId}"]`);
+      await frame.waitFor(`span[title="${tableId}"]`);
+      
+      // delete container
+
+      // click context menu for container
+      await frame.waitFor(`div[data-test="${tableId}"] > div > button`);
```
languy, line 50: Can we set an `id` to the button and remove the `div` from the selector so we make it less easy to break?

test/cassandra/container.spec.ts
zfoster: Yeah, these tests fail lint and are formatted but it's never changed the quotes or semicolons. Somewhat frustrating, maybe I can fix the lint rules

test/cassandra/container.spec.ts
zfoster: Yeah, I think so. It's somewhat difficult to get puppeteer to find these elements without the tree as it doesn't reload the DOM tree during the tests. I was able to remove a lot of these "drilling in" selectors in the SQL spec so will give it a go for Cassandra too

test/cassandra/container.spec.ts
zfoster: These `waitFor` calls end up making the tests much less brittle but can be reduced. The tests try to click elements before they seem to show up without them. In this particular case, when we click `Delete` to confirm deleting a container, there is significant lag for the UI to show the loading state (and we are waiting for the "ready" state). So the "ready" state is already present, and the test moves on before we ever hit the loading state. I can reduce this I think.

As for speed, I'll get these running in parallel when I'm working on the last one to hopefully avoid any issues with CI taking forever.

test/cassandra/container.spec.ts
zfoster: Yeah, said above but I'll refactor these to use `data-test` at the relevant level. had some issues with puppeteer not being able to find them but I suspect there's a way to fix

test/cassandra/container.spec.ts
zfoster: The best solution is probably to wait for the loading state, then the ready state!

southpolesteve (approved):

zfoster:
@languy I had trouble removing those dependencies on layouts in some of these selectors. puppeteer just cant seem to find certain elements _even_ when they have a data-test or an id. I'm not sure why yet, so going to revisit. Made some of those delays constants and will be removing some of them when I merge mongo since a few are related to @victor-meng's work

zfoster:
want to merge this since the last test found some bugs!

PR #196, opened 2020-09-12
Fix file downloads from notebooks container

gitanuj:
- Fix `NotebookContentClient.readFileContent` to return content string for `text`, `json`, and `base64` files
- Fix how we're constructing `Blob` from a `string`

Verified that the following scenarios now work:
- Downloading `.csv`, `.xlsx`, `.txt`, `.ipynb` files from container works
- `Copy To...` works for `.csv`, `.txt`, `.ipynb` files from container to github

TODO
- `Copy To...` doesn't work for `.xlsx` from container to github correctly. It copies the file but it is corrupted.

### DIFF
--- a/src/Explorer/Notebook/NotebookContentClient.ts
+++ b/src/Explorer/Notebook/NotebookContentClient.ts
@@ -194,18 +194,24 @@ export class NotebookContentClient {
       });
   }
 
-  public readFileContent(filePath: string): Promise<string> {
-    const fileType = NotebookUtil.isNotebookFile(filePath) ? "notebook" : "file";
-    return this.contentProvider
-      .get(this.getServerConfig(), filePath, { type: fileType, format: "text", content: 1 })
-      .toPromise()
-      .then(xhr => {
-        const content = (xhr.response as any).content;
-        if (!content) {
-          throw new Error("No content read");
-        }
+  public async readFileContent(filePath: string): Promise<string> {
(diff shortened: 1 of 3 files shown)


swviswan (dismissed):
:shipit:

swviswan (approved):
:shipit:

vchske (approved):

src/Explorer/Notebook/NotebookContentClient.ts
```diff
@@ -194,18 +194,24 @@ export class NotebookContentClient {
...
-    const fileType = NotebookUtil.isNotebookFile(filePath) ? "notebook" : "file";
-    return this.contentProvider
-      .get(this.getServerConfig(), filePath, { type: fileType, format: "text", content: 1 })
-      .toPromise()
-      .then(xhr => {
-        const content = (xhr.response as any).content;
-        if (!content) {
-          throw new Error("No content read");
-        }
+  public async readFileContent(filePath: string): Promise<string> {
```
southpolesteve, line 197: Not blocking, but why can't you using `fetch` and `response.blob()` for all of this? Can't Juno return a binary response?

southpolesteve (approved):

src/Explorer/Notebook/NotebookContentClient.ts
gitanuj: It's Nteract which is talking to Jupyter server and unfortunately we can't tell it to get a binary response currently. It'll be a bigger change to allow Nteract do that which we can tackle later on. This unblocks users right now who want to download binary files.

PR #200, opened 2020-09-14
Add telemetry for OpenTerminal

gitanuj:
- Add telemetry for `OpenTerminal`
- Add support for iframes within Data Explorer to also send telemetry messages to Portal window (for kusto logs)

This change broke fullscreen DE because the `getDataExplorerWindow` function got stuck in a busy while loop and had to be reverted. It happened because in Javascript if a window does not have a parent, its parent property is a reference to itself. Updated the method to work in that case as well. That behavior can't be tested locally so I added unit tests to test all possible theoretical scenarios.

### DIFF
--- a/src/Common/MessageHandler.ts
+++ b/src/Common/MessageHandler.ts
@@ -48,15 +48,37 @@ export function sendCachedDataMessage<TResponseDataModel>(
 
 export function sendMessage(data: any): void {
   if (canSendMessage()) {
-    window.parent.postMessage(
-      {
-        signature: "pcIframe",
-        data: data
-      },
-      window.document.referrer
-    );
+    const dataExplorerWindow = getDataExplorerWindow(window);
+    if (dataExplorerWindow) {
+      dataExplorerWindow.parent.postMessage(
+        {
+          signature: "pcIframe",
+          data: data
+        },
+        dataExplorerWindow.document.referrer
+      );
+    }
   }
 }
+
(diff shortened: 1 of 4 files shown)


languy (approved):
Great!

src/Common/MessageHandler.ts
```diff
@@ -48,16 +48,38 @@ export function sendCachedDataMessage<TResponseDataModel>(
...
 }
 
+// Only exported for unit tests
+export const getDataExplorerWindow = (currentWindow: Window): Window | undefined => {
+  // Start with the current window and traverse up the parent hierarchy to find a window
+  // with `dataExplorerPlatform` property
+  let dataExplorerWindow: Window | undefined = currentWindow;
+  // TODO: Need to `any` here since the window imports Explorer which can't be in strict mode yet
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  while (dataExplorerWindow && (dataExplorerWindow as any).dataExplorerPlatform == undefined) {
```
vignesh-mr, line 71: dataExplorerWindow as ViewModels.Explorer

vignesh-mr (approved):

src/Common/MessageHandler.ts
gitanuj: We can't do that because this file is in strict mode and Explorer.ts isn't (also have this as a TODO comment).

PR #206, opened 2020-09-15
Fixed aria label for autopilot throughput

DanielSPham:
The ariaLabel parameter is used for another text field in this component and only one text field is shown at a time. Hard coding aria label to match the hard coded label for this text field.


southpolesteve:
@srinathnarayanan make sure to pick up this change in any of your work

southpolesteve (approved):

ato9000 (approved):

PR #207, opened 2020-09-15
Adds mongo e2e spec

zfoster:
Similar to the SQL container spec, this adds Mongo
Adds helper function `login` to share between specs
Reduces delays resulting in 10+ seconds saved per test (~2.5 mins total time)
Uses constant delay times for rendering


package.json
```diff
@@ -193,8 +193,8 @@
     "compile": "tsc",
     "compile:contracts": "tsc -p ./tsconfig.contracts.json",
     "compile:strict": "tsc -p ./tsconfig.strict.json",
-    "format": "prettier --write \"{src,cypress}/**/*.{ts,tsx,html}\" \"*.{js,html}\"",
-    "format:check": "prettier --check \"{src,cypress}/**/*.{ts,tsx,html}\" \"*.{js,html}\"",
+    "format": "prettier --write \"{src,cypress,test}/**/*.{ts,tsx,html}\" \"*.{js,html}\"",
+    "format:check": "prettier --check \"{src,cypress,test}/**/*.{ts,tsx,html}\" \"*.{js,html}\"",
```
languy, line 197: yay! 👍

test/mongo/container.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+jest.setTimeout(300000);
+
+const LOADING_STATE_DELAY = 2500;
+const RENDER_DELAY = 1000;
+
+describe("Collection Add and Delete Mongo spec", () => {
+  it("creates and deletes a collection", async () => {
+    try {
+      const dbId = `TestDatabase${crypto.randomBytes(8).toString("hex")}`;
```
languy, line 13: nit: would be nice to isolate this into a function like `generateUniqueName(baseName: string, length: number): string` that we reuse in other places.

test/mongo/container.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      // validate created
+      // open database menu
+      await frame.waitFor(`span[title="${dbId}"]`);
+      await frame.waitForSelector('div[class="splashScreen"] > div[class="title"]', { visible: true });
+
+      await frame.click(`div[data-test="${dbId}"]`);
+      await frame.waitFor(`span[title="${collectionId}"]`, { visible: true });
+      await frame.waitFor(RENDER_DELAY);
+      await frame.waitFor(`span[title="${collectionId}"]`, { visible: true });
```
languy, line 55: nit: consider breaking this down into two functions (one that creates and one that deletes) if you think they could be reused later in other tests.

languy (approved):

southpolesteve (approved):

PR #215, opened 2020-09-17
[Accessibility] Add horizontal scroll bar if tooltip box goes outside of panel boundary

victor-meng:
Currently if we apply text spacing, the tooltip of the info icon goes outside of the boundary of the add collection panel and becomes partially invisible:
image

In order to show the complete tooltip, we can either add a horizontal scroll bar or remove the `width` and `min-width` setting of the tooltip. However, the latter fix will also make the width of the tooltip box too narrow when text spacing is off so I think the best fix is to simply add a horizontal scroll bar by changing the horizontal overflow from `hidden` to `auto`.

### DIFF
--- a/less/documentDB.less
+++ b/less/documentDB.less
@@ -1739,7 +1739,7 @@ input::-webkit-calendar-picker-indicator {
     padding-right: 34px;
     color: @BaseDark;
     overflow-y: auto;
-    overflow-x: hidden;
+    overflow-x: auto;
     margin: (2 * @MediumSpace) 0px;
 }
 


languy:
Is it possible to make it right-align instead? ie the carret would be located on the top the right side of the text box rather than on the top left.

victor-meng:
> Is it possible to make it right-align instead? ie the carret would be located on the top the right side of the text box rather than on the top left.

It's actually the first thing that I tried but unfortunately some of the tooltip will go outside of the left border of the panel so the issue remains.

I think adding a scrollbar is fine as a short-term solution. Once we move the add collection panel to react and replace the tooltip with the tooltip component from UI Fabric, this issue should be fixed.

southpolesteve (approved):

vchske (approved):

ato9000 (approved):

PR #221, opened 2020-09-21
Fix incorrect usage of TelemetryProcessor

gitanuj:
In TelemetryProcessor's traceSuccess there are 2 optional parameters with the first one being unknown type. This lead to some incorrect usage of TelemetryProcessor functions.

- Remove unknown parameter type from TelemetryProcessor functions to catch incorrect usage at compile time
- Update TelemetryProcessor.getData to work for iframes

This fixes the issue where OpenTerminal log doesn't contain "success" actionModifier

### DIFF
--- a/src/Terminal/index.ts
+++ b/src/Terminal/index.ts
@@ -6,7 +6,8 @@ import { ServerConnection } from "@jupyterlab/services";
 import { JupyterLabAppFactory } from "./JupyterLabAppFactory";
 import { Action } from "../Shared/Telemetry/TelemetryConstants";
 import * as TelemetryProcessor from "../Shared/Telemetry/TelemetryProcessor";
-import { updateUserContext, userContext } from "../UserContext";
+import { updateUserContext } from "../UserContext";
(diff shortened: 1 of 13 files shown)


src/Terminal/index.ts
```diff
@@ -44,6 +45,12 @@ const createServerSettings = (urlVars: { [key: string]: string }): ServerConnect
 
 const main = async (): Promise<void> => {
   const urlVars = getUrlVars();
+
+  // Initialize userContext. Currently only subscrptionId is required by TelemetryProcessor
+  updateUserContext({
+    subscriptionId: urlVars["subscriptionId"]
```
vignesh-mr, line 51: Let's use a constant for the "subscriptionId"

vignesh-mr (approved):

src/Terminal/index.ts
gitanuj: Replaced all Terminal query params with constant

southpolesteve (approved):

PR #222, opened 2020-09-22
Fix resource tree refresh issue

victor-meng:
Currently the resource tree would re-render a total of 5 times when a database node is expanded. 

The first two re-renders are triggered when:
1. A node is selected (expanding a database node also selects it).
2. The collections of a database have finished loading.

The second re-render is expected. The resource tree must re-render after the collections have been loaded. The first one is also necessary because when a node is selected, we not only need to highlight the selected node, but we also need to remove the hightlighting from the previous selected node.

The third re-render is also triggered by loading the collections. This is because we have a duplicate subscription to `database.collections`. The `cleanupDatabasesKoSubs` function is supposed to remove any duplicate subscriptions but due to a bug in the code, it never does.

The fourth and fifth re-render are both triggered by `Database.deleteCollectionsFromList`. However, in the case of expanding a database node, there's no collection to delete. It still triggers a re-render because it doesn't check if the collections to delete array is empty. Then the fifth re-render is triggered because of the duplicate subscription issue.

This PR will remove the last three re-renders by properly cleaning up duplicate subscriptions and adding a check to see if there's no collection to delete.

### DIFF
--- a/src/Explorer/Tree/ResourceTreeAdapter.tsx
+++ b/src/Explorer/Tree/ResourceTreeAdapter.tsx
@@ -64,7 +64,7 @@ export class ResourceTreeAdapter implements ReactAdapter {
 
     this.container.nonSystemDatabases.subscribe((databases: ViewModels.Database[]) => {
       // Clean up old databases
-      this.cleanupDatabasesKoSubs(databases.map((database: ViewModels.Database) => database.id()));
+      this.cleanupDatabasesKoSubs();
 
       databases.forEach((database: ViewModels.Database) => this.watchDatabase(database));
       this.triggerRender();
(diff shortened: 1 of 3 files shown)


test/container.spec.ts
```diff
@@ -1,10 +1,10 @@
 import "expect-puppeteer";
-import crypto from 'crypto'
+import crypto from "crypto";
```
southpolesteve, line 2: @victor-meng Any idea why these weren't already formatted correctly?

southpolesteve (approved):

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -800,11 +800,7 @@ export class ResourceTreeAdapter implements ReactAdapter {
   }
 
   private cleanupDatabasesKoSubs(existingDatabaseIds: string[]): void {
-    const databaseIdsToRemove = this.databaseCollectionIdMap
-      .keys()
-      .filter((id: string) => existingDatabaseIds.indexOf(id) === -1);
-
-    databaseIdsToRemove.forEach((databaseId: string) => {
+    existingDatabaseIds.forEach((databaseId: string) => {
```
languy, line 803: I believe the previous code's purpose was to `dispose()` any existing subscription to a database which is not used anymore.
So you might want to consider using `this.databaseCollectionIdMap.keys()` (without the `filter()`'ing, which, as you pointed out, results in duplicates) rather than `existingDatabaseIds` to make sure it cleans up everything.

test/container.spec.ts
languy: I believe @zfoster was looking into it for one of his PR's.

languy (approved):

src/Explorer/Tree/ResourceTreeAdapter.tsx
victor-meng: Since we are lazy loading the collections now, the `this.databaseCollectionIdMap` will be empty before we expand the database node so we can't just use `this.databaseCollectionIdMap.keys()`.

However I do see your point that if a database is deleted, then it won't be in the `existingDatabaseIds` array so its subscription won't get cleaned up. I think the correct way is to use `this.koSubsDatabaseIdMap.keys()` in this case since it should contain all databases including the deleted ones.

test/container.spec.ts
victor-meng: @southpolesteve This file wasn't formatted correctly when Zach checked it in in this commit: https://github.com/Azure/cosmos-explorer/commit/3fe63e88cb581c802a48919afd8111dbefeb1469. I probably turned formatting off too when I was making changes to some CSS files when I also modified this file.

zfoster:
FYI I fixed the formatting of these e2e tests in the mongo PR and have format working now

Software Engineer
CosmosDB
________________________________

PR #225, opened 2020-09-22
Nests focus inside cassandra promise

zfoster:
Locally, this was working since the cassandra data loading was broken, but then was failing when deployed.

The actual original issue was that it was trying to focus outside the `.then()` in the cassandra condition, this focuses in both places now


vignesh-mr (approved):

zfoster:
Opening a new PR without @southpolesteve's changes

PR #228, opened 2020-09-24
Fix create database for serverless accounts

victor-meng:
By default, the `AddCollectionPane` and `AddDatabasePane` set shared throughput to true. This causes the create database call to fail for serverless accounts because shared throughput is not supported.

The fix is to set shared throughput to false by default if the database account is serverless.

Similarly, the read database offer call would fail if the account is serverless so we shouldn't make the call if it's serverless.

### DIFF
--- a/src/Common/dataAccess/readOffers.ts
+++ b/src/Common/dataAccess/readOffers.ts
@@ -6,11 +6,7 @@ import { client } from "../CosmosClient";
 import { sendCachedDataMessage } from "../MessageHandler";
 import { userContext } from "../../UserContext";
 
-export const readOffers = async (isServerless?: boolean): Promise<Offer[]> => {
-  if (isServerless) {
-    return []; // Reading offers is not supported for serverless accounts
-  }
-
+export const readOffers = async (): Promise<Offer[]> => {
   try {
     if (configContext.platform === Platform.Portal) {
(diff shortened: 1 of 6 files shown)


southpolesteve (approved):
Would like @gitanuj to sign off too

src/Common/dataAccess/readOffers.ts
```diff
@@ -6,11 +6,7 @@ import { client } from "../CosmosClient";
 import { sendCachedDataMessage } from "../MessageHandler";
 import { userContext } from "../../UserContext";
 
-export const readOffers = async (isServerless?: boolean): Promise<Offer[]> => {
-  if (isServerless) {
-    return []; // Reading offers is not supported for serverless accounts
```
gitanuj, line 11: Is reading offers now supported for serverless accounts?

gitanuj (approved):
Please validate if create database/collection and refresh works for serverless accounts.

src/Common/dataAccess/readOffers.ts
victor-meng: No I'm just moving the check to the caller so it won't make the call if the account is serverless. It's cleaner that way because now we have another code path that goes through RP which also doesn't support serverless.

southpolesteve: Its is not, so we need to be sure this change does not break serverless in some way

PR #230, opened 2020-09-25
Don't generate source maps for vendor files

southpolesteve:
This change excludes all vendor chunks from generating source maps. In practice we rarely need these for debugging. On my machine it decreases the total build time from 8min to 2min and the dist size from 200mb to 50mb

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,6 +11,7 @@ const childProcess = require("child_process");
 const BundleAnalyzerPlugin = require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
 const TerserPlugin = require("terser-webpack-plugin");
 const isCI = require("is-ci");
+const webpack = require("webpack");
 
 const gitSha = childProcess.execSync("git rev-parse HEAD").toString("utf8");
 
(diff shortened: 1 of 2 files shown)


vignesh-mr (approved):

webpack.config.js
```diff
@@ -194,7 +204,7 @@ module.exports = function(env = {}, argv = {}) {
       filename: "[name].[chunkhash:6].js",
       path: path.resolve(__dirname, "dist")
     },
-    devtool: mode === "development" ? "cheap-eval-source-map" : "source-map",
+    devtool: false,
```
vignesh-mr, line 207: Looks like this is the default. Do we need this?

https://webpack.js.org/configuration/devtool/#devtool

ato9000 (approved):

webpack.config.js
southpolesteve: Nope! Removed

languy (approved):

swviswan (approved):
:shipit:

southpolesteve:
Unfortunately this change is causing some kind of issue preventing source maps from being properly generated. There are bug reports online about ts-loader, babel-loader, and webpack sourcemaps all not playing nice. Not going to pursue it right now.

PR #234, opened 2020-09-25
Lazy load collection offer

victor-meng:
- Load collection offer and collection quota info only when the user is trying to open the settings tab
- Similar to `readDatabaseOffer` from my previous commit https://github.com/Azure/cosmos-explorer/pull/208, the `readCollectionOffer` is currently in a hybrid state where we get the offer id by calling RP and then pass the offer id to SDK to get the full offer. This is because the response of the RP's get collection throughput call is missing some information such as number of partitions. Once RP fixes this, we will no longer need to pass the offer id to SDK and do an extra call.
- Refactored `readCollectionQuotaInfo`
- Added error logging for `readDatabaseOffer`


src/Explorer/Tree/Collection.ts
```diff
@@ -541,7 +543,7 @@ export default class Collection implements ViewModels.Collection {
     }
   };
 
-  public onSettingsClick = () => {
+  public onSettingsClick = async () => {
```
languy, line 546: nit: return type?

languy (approved):

southpolesteve (approved):

src/Explorer/Tree/Collection.ts
victor-meng: Done

PR #237, opened 2020-09-29
Fix Parent Origin Regex

southpolesteve:
My original change to the new set of Regex's did not account for the bare production domains. So `ms.portal.azure.com` worked but not `portal.azure.com`

### DIFF
--- a/src/ConfigContext.ts
+++ b/src/ConfigContext.ts
@@ -31,10 +31,11 @@ interface ConfigContext {
 let configContext: Readonly<ConfigContext> = {
   platform: Platform.Portal,
   allowedParentFrameOrigins: [
-    `^https:\\/\\/cosmos.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]*portal.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]*ext.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]*microsoftazure.de$`
+    `^https:\\/\\/cosmos\\.azure\\.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*portal\\.azure\\.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*ext\\.azure\\.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*\\.ext\\.microsoftazure\\.de$`,
+    `^https://cosmos-db-dataexplorer-germanycentral.azurewebsites.de$`
   ],
   // Webpack injects this at build time
   gitSha: process.env.GIT_SHA,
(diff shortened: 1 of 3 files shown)


victor-meng (approved):

src/Utils/MessageValidation.ts
```diff
@@ -17,5 +17,6 @@ function isValidOrigin(allowedOrigins: string[], event: MessageEvent): boolean {
       return true;
     }
   }
+  console.error(`Invalid parent frame origin decected: ${eventOrigin}`);
```
zfoster, line 20: decected-> detected

zfoster (approved):

gitanuj (approved):

src/ConfigContext.ts
```diff
@@ -32,9 +32,9 @@ let configContext: Readonly<ConfigContext> = {
   platform: Platform.Portal,
   allowedParentFrameOrigins: [
     `^https:\\/\\/cosmos.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]+.portal.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]+.ext.azure.(com|cn|us)$`,
-    `^https:\\/\\/[\\.\\w]+microsoftazure.de$`
+    `^https:\\/\\/[\\.\\w]*portal.azure.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*ext.azure.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*microsoftazure.de$`
```
ato9000, line 37: please test with malicious.germanycentral.cloudapp.microsoftazure.de

ato9000 (changes requested):

src/ConfigContext.ts
vignesh-mr: I'd vote we use `^https:\\/\\/[\\.\\w]*ext.microsoftazure.de$` so only portal extensions are whitelisted. The one above whitelists any public azure resource in blackforest.

vignesh-mr (approved):

src/Utils/MessageValidation.ts
southpolesteve: Fixed

src/ConfigContext.ts
ato9000: also add a test for the below (I know you regex works, but to avoid regressions by common mistakes)

maliciousazure.com
maliciousportalsazure.com

southpolesteve: My original regex would have allowed the second one :/

I've added the above cases and explicitly whitelisted `https://cosmos-db-dataexplorer-germanycentral.azurewebsites.de/` since it is what we use in German cloud

ato9000 (approved):

PR #239, opened 2020-09-30
Switch notebook editors to monaco

languy:
This integrates the monaco editor made available in nteract by [this PR](https://github.com/nteract/nteract/pull/5267).
Now the code, markdown and raw cells use the monaco editor for regular notebooks. For read-only notebooks (ie gallery), only raw and code cells use it as markdown will be substantially reworked in the future.
Opening a non-notebook file used to open the monaco editor as well and now opens its latest version: updated this code path with the latest monaco component api changes.

This integration required an upgrade of React and nteract to latest which then required upgrading these different packages to latest:
* enzyme
* d3: Data Explorer Graph was importing nteract's "data explorer" (its viz tool) d3 packages, instead of the older d3 version specified in `package.json`. The only way to resolve these version collision was to upgrade d3 to latest and fix `D3ForceGraph` to deal with the version breaking changes.
* JupyterLab: the React upgrade to latest required a new terminal widget (`@jupyterlab/terminal`) which itself required new services (`@jupyterlab/services`). We must use latest rc version which includes fix for [websocket auth bug](https://github.com/jupyterlab/jupyterlab/issues/9070).

Other changes:
* Update obsolete calls to deprecated `rxjs` methods.
* On the nteract side, a notable design change is that configuration is now done with the `@nteract/mythic-configuration` (auto-save interval, show line numbers), as opposed to previously being stored in the redux state.
This `mythic-configuration` package also takes over the instantiation and configuration of the redux store: we simply pass epics, initial state, middleware, etc to it.


southpolesteve (approved):

gitanuj (approved):
:shipit:

PR #240, opened 2020-10-01
Use 2020-03-01 API version for ARM calls

southpolesteve:
We need to support some environments that do not yet have 2020-04-01 ARM API version. However the requests are backwards compatible where possible and the breaking changes are not used in this repo. So this is a bit hacky, but should work fine until we can use 2020-04-01 everywhere.


southpolesteve:
On second thought, I think I will make this a parameter that can be overridden by query param or config.

PR #241, opened 2020-10-02
Allow setting of ARM API version via query param

southpolesteve:
Allows overriding the ARM API version via query param: https://cosmos.azure.com/explorer.html?armAPIVersion=2020-03-01

### DIFF
--- a/src/ConfigContext.ts
+++ b/src/ConfigContext.ts
@@ -25,6 +25,7 @@ interface ConfigContext {
   GITHUB_CLIENT_ID: string;
   GITHUB_CLIENT_SECRET?: string; // No need to inject secret for prod. Juno already knows it.
   hostedExplorerURL: string;
+  armAPIVersion?: string;
 }
 
 // Default configuration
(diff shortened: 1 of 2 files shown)


ato9000 (approved):

src/Utils/arm/request.ts
```diff
@@ -43,7 +44,8 @@ interface Options {
 // TODO: This is very similar to what is happening in ResourceProviderClient.ts. Should probably merge them.
 export async function armRequest<T>({ host, path, apiVersion, method, body: requestBody }: Options): Promise<T> {
   const url = new URL(path, host);
-  url.searchParams.append("api-version", apiVersion);
+  console.log(configContext);
```
languy, line 47: Not sure whether this is for dev purposes or cx-facing. If the latter, consider tracing somewhere (like notification console or something).

src/ConfigContext.ts
```diff
@@ -93,6 +94,10 @@ export async function initializeConfiguration(): Promise<ConfigContext> {
     }
     // Allow override of platform value with URL query parameter
     const params = new URLSearchParams(window.location.search);
+    if (params.has("armAPIVersion")) {
```
languy, line 97: Consider making it all lowercase since query params are case-insensitive (right?)

languy (approved):

src/ConfigContext.ts
southpolesteve: They are case sensitive. Headers are not

src/Utils/arm/request.ts
southpolesteve, line 47:
```suggestion
```

PR #259, opened 2020-10-07
Move update offers call to RP

victor-meng:
There are currently two problems with RP's update offer call:
- RP's update database/collection throughput call has the same issue as the read throughput call where the response is missing properties like number of partitions and minimum throughput. I have to add an extra read database/collection offer call after the updating the offer.
- The migrate to auto pilot/manual call returns a "MethodNotAllowed" error even though the migration itself is successful. I have to add a try catch block to ignore the "MethodNotAllowed" errors.

### DIFF
--- a/src/Common/dataAccess/readDatabaseOffer.ts
+++ b/src/Common/dataAccess/readDatabaseOffer.ts
@@ -17,29 +17,19 @@ import { userContext } from "../../UserContext";
 export const readDatabaseOffer = async (
   params: DataModels.ReadDatabaseOfferParams
 ): Promise<DataModels.OfferWithHeaders> => {
+  if (userContext.defaultExperience === DefaultAccountExperienceType.Table) {
+    throw new Error("Reading database offer is not allowed for tables accounts");
+  }
+
   const clearMessage = logConsoleProgress(`Querying offer for database ${params.databaseId}`);
   let offerId = params.offerId;
   if (!offerId) {
-    if (
-      window.authType === AuthType.AAD &&
-      !userContext.useSDKOperations &&
-      userContext.defaultExperience !== DefaultAccountExperienceType.Table
-    ) {
-      try {
-        offerId = await getDatabaseOfferIdWithARM(params.databaseId);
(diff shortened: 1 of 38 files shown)


src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -426,7 +426,21 @@ export class SettingsComponent extends React.Component<SettingsComponentProps, S
             throw error;
           }
         } else {
-          const updatedOffer: DataModels.Offer = await updateOffer(this.collection.offer(), newOffer, headerOptions);
+          const updateOfferParams: DataModels.UpdateOfferParams = {
+            databaseId: this.collection.databaseId,
+            collectionId: this.collection.id(),
```
srinathnarayanan, line 431: have you tested this using the feature flag "enableSettingsV2" ?
just to be sure, make sure you can update from autopilot -> manual and manual -> autopilot when you have the feature flag set

victor-meng: Tested this change with the new settings tab and it worked fine.

southpolesteve (commented):
One suggestion that I think is a typo

src/Common/dataAccess/readDatabaseOffer.ts
```diff
@@ -75,24 +65,32 @@ const getDatabaseOfferIdWithARM = async (databaseId: string): Promise<string> =>
...
+        rpResponse = await getGremlinDatabaseThroughput(subscriptionId, resourceGroup, accountName, databaseId);
+        break;
+      default:
+        throw new Error(`Unsupported default experience type: ${defaultExperience}`);
+    }
+
+    return rpResponse?.name;
+  } catch (error) {
+    if (error.code !== "NotFound") {
+      throw new error();
```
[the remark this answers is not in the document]
victor-meng: Yep thanks for catching it!

southpolesteve (approved):

srinathnarayanan (approved):

PR #263, opened 2020-10-07
Restore AppInsights fetch telemetry

southpolesteve:
@gitanuj I tested terminal with this and it works. Just turning off the correlation headers entirely.

### DIFF
--- a/src/Shared/appInsights.ts
+++ b/src/Shared/appInsights.ts
@@ -2,7 +2,9 @@ import { ApplicationInsights } from "@microsoft/applicationinsights-web";
 
 const appInsights = new ApplicationInsights({
   config: {
-    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49"
+    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49",
+    disableFetchTracking: false,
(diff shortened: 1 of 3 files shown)


gitanuj (approved):

src/Shared/appInsights.ts
```diff
@@ -2,7 +2,9 @@ import { ApplicationInsights } from "@microsoft/applicationinsights-web";
 
 const appInsights = new ApplicationInsights({
   config: {
-    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49"
+    instrumentationKey: "fa645d97-6237-4656-9559-0ee0cb55ee49",
+    disableFetchTracking: false,
+    disableCorrelationHeaders: true
```
swviswan, line 7: >disableCorrelationHeaders [](start = 4, length = 25)

was the issue last time we had because we did not disable correlation header but juts turned on fetch tracking?

southpolesteve: Correct. There is also an option to turn of the headers for certain domains via regex, but I think this is easier to do right now. The headers are not used AFAIK by any of our services.

swviswan (approved):
:shipit:

PR #264, opened 2020-10-07
Fixed settingsV2  bugs and added experimentation

srinathnarayanan:
This PR does the following
- Replaces settingsv2 feature flag with flights passed on to data explorer initialization 
- Truncates throughput and TTL seconds text input fields when it exceeds the max value
- Fixes a bug due to which auto scale throughput increased by 100 instead of 1000 when the up arrow is pressed
- Adds dirty color change to Autoscale-throughput radio buttons


src/Common/Constants.ts
```diff
@@ -131,6 +130,10 @@ export class Features {
   public static readonly enableSDKoperations = "enablesdkoperations";
 }
 
+export class Flights {
+  public static readonly settingsV2 = "settingsv2";
```
swviswan, line 134: >settingsv2 [](start = 39, length = 10)

Settingsv2? Can be the flight name with upper case?

src/Explorer/Explorer.ts
```diff
@@ -1953,6 +1956,16 @@ export default class Explorer {
     return Q();
   }
 
+  public setFeatureFlagsFromFlights(flights: readonly string[]): void {
+    if (!flights) {
+      return;
+    }
+
+    if (flights.filter((flightName: string) => flightName === Constants.Flights.settingsV2).length === 1) {
```
swviswan, line 1964: >filter [](start = 16, length = 6)

Can we simply use indexof? filter creates new array which seems unnecessary here.

src/Explorer/Explorer.ts
```diff
@@ -1953,6 +1956,16 @@ export default class Explorer {
     return Q();
   }
 
+  public setFeatureFlagsFromFlights(flights: readonly string[]): void {
+    if (!flights) {
+      return;
+    }
+
+    if (flights.filter((flightName: string) => flightName === Constants.Flights.settingsV2).length === 1) {
+      this.isSettingsV2Enabled(true);
```
swviswan, line 1965: >isSettingsV2Enabled [](start = 11, length = 19)

as discussed yesterday, do you want to turn it on for MPAC - just for this feature so we can find issues faster?

src/Common/Constants.ts
srinathnarayanan: resolved offline, flights returned from experimentation package are lowercase

src/Common/Constants.ts
```diff
@@ -131,6 +130,10 @@ export class Features {
   public static readonly enableSDKoperations = "enablesdkoperations";
 }
 
+export class Flights {
```
swviswan, line 133: >Flights [](start = 13, length = 7)

Can you please add a comment here that all flights from the framework will be lowercase?

swviswan (approved):
:shipit:

vignesh-mr (approved):

PR #265, opened 2020-10-08
Add Report Abuse dialog for public gallery notebooks

gitanuj:
image


src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
```diff
@@ -3,10 +3,14 @@ import { Icon, Label, Stack, HoverCard, HoverCardType, Link } from "office-ui-fa
...
-  private getInfoPanel = (iconName: string, labelText: string, url: string): JSX.Element => {
+export interface InfoComponentProps {
+  onReportAbuseClick?: () => void;
+}
+
+export class InfoComponent extends React.Component<InfoComponentProps> {
+  private getInfoPanel = (iconName: string, labelText: string, url?: string, onClick?: () => void): JSX.Element => {
     return (
-      <Link href={url} target="_blank">
+      <Link href={url} target={url && "_blank"} onClick={onClick}>
```
vignesh-mr, line 13: Does this force the browser to report errors if target is undefined? I'd check and confirm before this adds more noise in appinsights.

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
```diff
@@ -25,6 +29,11 @@ export class InfoComponent extends React.Component {
         <Stack.Item>
           {this.getInfoPanel("KnowledgeArticle", "Microsoft Terms of Use", CodeOfConductEndpoints.termsOfUse)}
         </Stack.Item>
+        {this.props.onReportAbuseClick !== undefined && (
```
vignesh-mr, line 32: Would use !this.props.onReportAbuseClick so we also cover nulls

vignesh-mr (approved):

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
```diff
@@ -25,6 +29,11 @@ export class InfoComponent extends React.Component {
...
+          <Stack.Item>
+            {this.getInfoPanel("ReportHacked", "Report Abuse", undefined, () => this.props.onReportAbuseClick())}
```
swviswan, line 34: >ReportHacked [](start = 32, length = 12)

?

src/Explorer/Controls/NotebookViewer/NotebookViewerComponent.tsx
```diff
@@ -140,6 +138,7 @@ export class NotebookViewerComponent extends React.Component<
               onFavoriteClick={this.favoriteItem}
               onUnfavoriteClick={this.unfavoriteItem}
               onDownloadClick={this.downloadItem}
+              onReportAbuseClick={this.state.galleryItem.isSample ? undefined : this.reportAbuse}
```
swviswan, line 141: >onReportAbuseClick [](start = 14, length = 18)

We should not even expose Report abuse button for the sample notebooks.

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -227,7 +227,7 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
           <Stack.Item styles={{ root: { minWidth: 200 } }}>
             <Dropdown options={this.sortingOptions} selectedKey={this.state.sortBy} onChange={this.onDropdownChange} />
           </Stack.Item>
-          {this.props.container?.isGalleryPublishEnabled() && (
+          {(!this.props.container || this.props.container.isGalleryPublishEnabled()) && (
```
swviswan, line 230: >!this.props.container [](start = 12, length = 21)

?
did you mean to do the valid check ? if so, why not just check for container && gallerypublishenabled?

swviswan (approved):
:shipit:

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
gitanuj: This is just to support both opening a url in a new tab and have a custom click handler. In case where url is defined we want to set href and target, otherwise we just se onClick. Verified that it doesn't report any errors for undefined target (it is only expected when href is set).

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
gitanuj: "ReportHacked" is actually the icon name in Office UI Fabric Icons pack that we're using here.
https://uifabricicons.azurewebsites.net/

src/Explorer/Controls/NotebookViewer/NotebookViewerComponent.tsx
gitanuj: This is doing exactly that. If it's a sample notebook we don't set onReportAbuseClick which hides the option altogether.

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
gitanuj: `isGalleryPublishEnabled` is checking the feature flag which only works for Data Explorer. However this component is used by standalone gallery too where we don't support feature flags. So to show `InfoComponent` I added another check (in standalone gallery `this.props.container` is always undefined).

src/Explorer/Controls/NotebookGallery/InfoComponent/InfoComponent.tsx
gitanuj: Addressing in my next PR

PR #269, opened 2020-10-09
Added telelmetry for settings v2 and v1

srinathnarayanan:
- Modified tab activate and close logging to contain tabId
- Overrode this tabId in SettingsV2 to make the tabId contain "SettingsV2"
- Changed type of telemetry action for updatecollection and update offer success and failure to "UpdateSettingsV2" in SettingsComponent
- Added new Action types "DiscardSettings" and "DiscardSettingsV2" for telemetry actions fired when discard button is clicked on setttings v2 and settings v2 respectively

### DIFF
--- a/src/Explorer/Tabs/SettingsTab.ts
+++ b/src/Explorer/Tabs/SettingsTab.ts
@@ -13,7 +13,7 @@ import Q from "q";
 import SaveIcon from "../../../images/save-cosmos.svg";
 import TabsBase from "./TabsBase";
 import * as TelemetryProcessor from "../../Shared/Telemetry/TelemetryProcessor";
-import { Action } from "../../Shared/Telemetry/TelemetryConstants";
+import { Action, ActionModifiers } from "../../Shared/Telemetry/TelemetryConstants";
 import { PlatformType } from "../../PlatformType";
 import { RequestOptions } from "@azure/cosmos/dist-esm";
 import Explorer from "../Explorer";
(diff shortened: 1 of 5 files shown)


src/Shared/Telemetry/TelemetryConstants.ts
```diff
@@ -86,7 +86,10 @@ export enum Action {
   CreateMongoCollectionWithWildcardIndex,
   ClickCommandBarButton,
   RefreshResourceTreeMyNotebooks,
-  ClickResourceTreeNodeContextMenuItem
+  ClickResourceTreeNodeContextMenuItem,
+  DiscardSettings,
+  UpdateSettingsV2,
+  DiscardSettingsV2
```
swviswan, line 92: >DiscardSettingsV2 [](start = 2, length = 17)

uggest: prefix with area like: SettingsUpdated, SettingsV2Updated, NotebooksRunCell...etc. So we can find all telemetry constants related to an area together when writing them in sorted order in this list.

swviswan (approved):
:shipit:

src/Shared/Telemetry/TelemetryConstants.ts
srinathnarayanan: the old action for updating settings was "UpdateSettings", so I followed that pattern. but, sure, will change it for SettingsV2

src/Explorer/Tabs/SettingsTab.ts
```diff
@@ -1233,6 +1233,9 @@ export default class SettingsTab extends TabsBase implements ViewModels.WaitsFor
   };
 
   public onRevertClick = (): Q.Promise<any> => {
+    TelemetryProcessor.trace(Action.DiscardSettings, ActionModifiers.Mark, {
+      message: "Settings Discarded"
```
languy, line 1237: Not sure whether this is needed, since the Action already describes. We can save a few bytes.

languy (approved):

PR #284, opened 2020-10-16
Added Support for editing Mongo Indexing Policy from Settings tab

srinathnarayanan:
- Added a MongoIndexingPolicy Component that allows users to add new indexes, drop and view exisiting Mongo Indexes.
- Added a AddMongoIndex Component that allows for dynamically adding any number of Indexes in one go
- Modified Mongo RP calls for Get and Update 
- Added a method to get index transformation progress for mongo collection

Mongo Index editor


src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,406 @@
...
+  };
+
+  public isMongoIndexingPolicySaveable = (): boolean => {
+    const addErrorsExist = this.props.indexesToAdd.find(
+      (addMongoIndexProps: AddMongoIndexProps) => addMongoIndexProps.notification
+    );
+    return (this.props.indexesToAdd.length > 0 && !addErrorsExist) || this.props.indexesToDrop.length > 0;
+  };
+
+  public isMongoIndexingPolicyDiscardable = (): boolean => {
```
languy, line 149: `isMongoIndexingPolicyChangesDiscardable`  rather?

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,406 @@
...
+    )?.notification.message;
+  };
+
+  private onRenderRow = (props: IDetailsRowProps): JSX.Element => {
+    return <DetailsRow {...props} styles={transparentDetailsRowStyles} />;
+  };
+
+  private getActionButton = (arrayPosition: number, isCurrentIndex: boolean): JSX.Element => {
+    return isCurrentIndex ? (
+      <IconButton
```
languy, line 166: `ariaLabel` possible here?

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,406 @@
...
+  };
+
+  private renderIndexesToBeDropped = (): JSX.Element => {
+    const indexesToBeDropped: MongoIndexDisplayProps[] = this.props.indexesToDrop.map(
+      (dropIndex: number, arrayPosition: number) =>
+        this.getMongoIndexDisplayProps(this.props.mongoIndexes[dropIndex], arrayPosition, false)
+    );
+
+    return (
+      <Stack styles={mediumWidthStackStyles}>
```
languy, line 291: Check if we can't reuse `CollapsiblePanel` in our code or if Fluent-UI has one. If we need a new one (probably), we should make this reusable and accessible. Use the pattern mentioned above (render `{this.props.children}`) so you can let the parent specify whatever component it wants inside this panel.

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,406 @@
...
+      indexTransformationWarning = (
+        <Stack horizontal {...mongoWarningStackProps}>
+          <Text>Refreshing indexing policy update information</Text>
+          <Spinner size={SpinnerSize.medium} />
+        </Stack>
+      );
+
+      // index transformation progress can be 0
+    } else if (this.isIndexingTransforming()) {
+      const updateInfoString = "The indexing policy is being updated in the background. ";
```
languy, line 345: Consider updating the `NotificationConsole` (it has a `Progress` type, so it can display transient messages).

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,406 @@
...
+    );
+  };
+
+  public render(): JSX.Element {
+    return this.props.mongoIndexes ? (
+      <Stack {...subComponentStackProps}>
+        {this.renderWarningMessage()}
+        {mongoIndexingPolicyDisclaimer}
+        {this.renderInitialIndexes()}
+        {longGrayLine}
```
languy, line 399: Looks like you use Fluent-UI's `Separator`.

languy (approved):

src/Common/PortalNotifications.ts
```diff
@@ -1,41 +1,41 @@
...
-    headers
-  });
-
-  if (!response.ok) {
-    throw new Error(await response.text());
-  }
-
-  return (await response.json()) as DataModels.Notification[];
-};
+import * as DataModels from "../Contracts/DataModels";
```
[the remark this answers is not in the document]
srinathnarayanan: I am unsure aswell. Maybe line endings. Will take a look.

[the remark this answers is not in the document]
srinathnarayanan: doesn't look like line endings either.

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
...
+      .container(collectionId)
+      .read({ populateQuotaInfo: true });
+
+    indexTransformationPercentage = parseInt(
+      response.headers[Constants.HttpHeaders.collectionIndexTransformationProgress] as string
+    );
+  } catch (error) {
+    logConsoleError(`Error while reading container ${collectionId}:\n ${JSON.stringify(error)}`);
+    logError(JSON.stringify(error), "ReadCollection", error.code);
+    sendNotificationForError(error);
```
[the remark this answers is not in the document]
srinathnarayanan: only for Forbidden error

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/AddMongoIndexComponent.tsx
```diff
@@ -0,0 +1,91 @@
...
+  MongoWildcardPlaceHolder
+} from "../../SettingsUtils";
+
+export interface AddMongoIndexComponentProps {
+  description: string;
+  type: MongoIndexTypes;
+  notification: MongoNotificationMessage;
+  onIndexAddOrChange: (description: string, type: MongoIndexTypes) => void;
+  onDiscard: () => void;
+  setRef: (textField: ITextField) => void;
```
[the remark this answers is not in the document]
srinathnarayanan: so in the magic rows that appear (new row appears when you begin typing in a previous row), the issues is on re render, focus is lost on the previous row where the user is typing something. Hence, I pass refs that are used when the child element is instantiated and use that to set focus. 
Is there a better way to do this?

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -544,6 +636,36 @@ export class SettingsComponent extends React.Component<SettingsComponentProps, S
     }
   };
 
+  private onIndexDrop = (index: number): void => this.setState({ indexesToDrop: [...this.state.indexesToDrop, index] });
```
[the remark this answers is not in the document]
srinathnarayanan: I need to maintain both the arrayposition and the mongoIndex, so a Set can't be used. maybe a map, but again I think that's overkill

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
srinathnarayanan: I d rather keep it this way, otherwise the update function for this (now named onMongoIndexingPolicyDiscardableChange) will become onMongoIndexingPolicyChangeDiscardableChange, which is not great

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
srinathnarayanan: We need the temp notification with spinner to show up here on top

swviswan:
> 

can this be moved to something like MongoUtils?

---
Refers to: src/Explorer/Controls/Settings/SettingsUtils.tsx:254 in 6c9c246. [](commit_id = 6c9c2469b06852d9003f35c980f09d62c0e24c05, deletion_comment = False)

swviswan:
> 

or just be part of mongo component itself?

---
In reply to: [713276462](https://github.com/Azure/cosmos-explorer/pull/284#issuecomment-713276462) [](ancestors = 713276462)

---
Refers to: src/Explorer/Controls/Settings/SettingsUtils.tsx:254 in 6c9c246. [](commit_id = 6c9c2469b06852d9003f35c980f09d62c0e24c05, deletion_comment = False)

srinathnarayanan:
> > 
> 
> can this be moved to something like MongoUtils?
> 
> Refers to: src/Explorer/Controls/Settings/SettingsUtils.tsx:254 in [6c9c246](https://github.com/Azure/cosmos-explorer/commit/6c9c2469b06852d9003f35c980f09d62c0e24c05). [](commit_id = [6c9c246](https://github.com/Azure/cosmos-explorer/commit/6c9c2469b06852d9003f35c980f09d62c0e24c05), deletion_comment = False)

ya it can be, but I wanted to move all methods that dont depend on a a class member or state or props to Utils for easier testing. They might also come in handy in the future if we have to reuse this elsewhere.

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/AddMongoIndexComponent.tsx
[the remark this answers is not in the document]
srinathnarayanan: added refs to individual components.

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
...
+    const response = await client()
+      .database(databaseId)
+      .container(collectionId)
+      .read({ populateQuotaInfo: true });
+
+    indexTransformationPercentage = parseInt(
+      response.headers[Constants.HttpHeaders.collectionIndexTransformationProgress] as string
+    );
+  } catch (error) {
+    logConsoleError(`Error while reading container ${collectionId}:\n ${JSON.stringify(error)}`);
```
swviswan, line 46: >logConsoleError [](start = 4, length = 15)

what is the philosophy behind console errors?

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
...
+    logError(JSON.stringify(error), "ReadCollection", error.code);
```
swviswan, line 47: >logError [](start = 4, length = 8)

is it Logger.logError?

swviswan:
>     logError(JSON.stringify(error), "ReadCollection", error.code);

is it Logger.logError??

---
Refers to: src/Common/dataAccess/readCollection.ts:22 in 0585e0e. [](commit_id = 0585e0e9bd063243aa991f1fd41858f6cfc3cd66, deletion_comment = False)

src/Common/PortalNotifications.ts
```diff
@@ -1,41 +1,41 @@
-import * as DataModels from "../Contracts/DataModels";
```
swviswan, line 1: >DataModels [](start = 12, length = 10)

just undo this file if no chnage?

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,385 @@
...
+        ariaLabel="Delete index Button"
+        iconProps={{ iconName: "Delete" }}
+        disabled={this.isIndexingTransforming()}
+        onClick={() => {
+          this.props.onIndexDrop(arrayPosition);
+        }}
+      />
+    ) : (
+      <IconButton
+        ariaLabel="Add back Index Button"
```
swviswan, line 160: >Add back Index Button [](start = 19, length = 21)

Add Index Button back

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
   clearMessage();
   return collection;
 }
+
+export async function getMongoCollectionIndexTransformationProgress(
```
southpolesteve, line 30: Why are we reading this via the SDK? Is this not exposed via RP?

src/Common/dataAccess/readCollection.ts
southpolesteve, line 46: JSON.stringify(error) is never what you want. It will log `{}` since errors have no enumerable properties. @victor-meng is fixing this up in other places and I will try to add a linting rule

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
...
+): Promise<MongoDBCollectionResource> {
+  let collection: MongoDBCollectionResource;
+  const subscriptionId = userContext.subscriptionId;
+  const resourceGroup = userContext.resourceGroup;
+  const accountName = userContext.databaseAccount.name;
+
+  const clearMessage = logConsoleProgress(`Reading container ${collectionId}`);
+  try {
+    const response = await getMongoDBCollection(subscriptionId, resourceGroup, accountName, databaseId, collectionId);
+    collection = response.properties.resource as MongoDBCollectionResource;
```
southpolesteve, line 67: Why the cast here? `getMongoDBCollection` should have fully generated types

src/Common/dataAccess/readCollection.ts
```diff
@@ -22,3 +26,51 @@ export async function readCollection(databaseId: string, collectionId: string):
...
+    logConsoleError(`Error while reading container ${collectionId}:\n ${JSON.stringify(error)}`);
+    logError(JSON.stringify(error), "ReadCollection", error.code);
+    sendNotificationForError(error);
+    throw error;
+  }
+  clearMessage();
+  return indexTransformationPercentage;
+}
+
+export async function readMongoDBCollectionThroughRP(
```
southpolesteve, line 55: Why are we exposing this? The idea of having `readCollection` was to abstract all of this away so the caller doesn't have to worry about the API specifics

src/Explorer/Controls/Settings/SettingsComponent.test.tsx
```diff
@@ -35,7 +46,10 @@ describe("SettingsComponent", () => {
       node: undefined,
       hashLocation: "settings",
       isActive: ko.observable(false),
-      onUpdateTabsButtons: undefined
+      onUpdateTabsButtons: undefined,
+      getPendingNotification: Q.Promise<DataModels.Notification>(() => {
```
southpolesteve, line 50: Remove new useage of Q promises wherever possible

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,385 @@
...
+          onDiscard={() => {
+            this.props.onRevertIndexAdd(indexesToAddLength);
+          }}
+        />
+      </Stack>
+    );
+  };
+
+  private renderInitialIndexes = (): JSX.Element => {
+    const initialIndexes: MongoIndexDisplayProps[] = this.props.mongoIndexes
```
southpolesteve, line 242: There are a lot of places you are specifying types where you don't need to. The TS compiler can infer from usage. Example here:

```  ts
const initialIndexes: MongoIndexDisplayProps[] = this.props.mongoIndexes
      .map((mongoIndex: MongoIndex, arrayPosition: number) =>
        this.getMongoIndexDisplayProps(mongoIndex, arrayPosition, true)
      )
      .filter(
        (value: MongoIndexDisplayProps, arrayPosition: number) =>
          !!value && !this.props.indexesToDrop.includes(arrayPosition)
      );
```

Without types

```  ts
const initialIndexes = this.props.mongoIndexes
      .map((mongoIndex, arrayPosition) =>
        this.getMongoIndexDisplayProps(mongoIndex, arrayPosition, true)
      )
      .filter(
        (value, arrayPosition) =>
          !!value && !this.props.indexesToDrop.includes(arrayPosition)
      );
```

TS will tell you if it is trying to infer `any` and fail to compile

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
```diff
@@ -0,0 +1,385 @@
...
+      this.setState({ isRefreshingIndexTransformationProgress: false });
+    }
+  };
+
+  public isIndexingTransforming = (): boolean =>
+    this.props.indexTransformationProgress !== undefined && this.props.indexTransformationProgress !== 100;
+
+  private onClickRefreshIndexingTransformationLink = async () => await this.refreshIndexTransformationProgress();
+
+  private renderIndexTransformationWarning = (): JSX.Element => {
```
southpolesteve, line 312: I see this pattern a few places. No need to create a reference to `let indexTransformationWarning: JSX.Element;`. Just return everywhere:

``` ts
private renderIndexTransformationWarning = (): JSX.Element => {
    if (this.state.isRefreshingIndexTransformationProgress) {
      return (
        <Stack horizontal {...mongoWarningStackProps}>
          <Text>Refreshing indexing policy update information</Text>
          <Spinner size={SpinnerSize.medium} />
        </Stack>
      );

      // index transformation progress can be 0
    } else if (this.isIndexingTransforming()) {
      const updateInfoString = "The indexing policy is being updated in the background. ";
This conversation was marked as resolved by srinathnarayanan

      if (this.props.indexTransformationProgress === 0) {
        return (
          <Text>
            {updateInfoString}
            <Link onClick={this.onClickRefreshIndexingTransformationLink}>
              {`Refresh to check if it has completed.`}
            </Link>
          </Text>
        );
      }
        return (
          <Text>
            {updateInfoString} The update is {this.props.indexTransformationProgress}% complete.
            <Link onClick={this.onClickRefreshIndexingTransformationLink}>{` Refresh to check the progress.`}</Link>
          </Text>
        );
      }
  };
```

src/Common/dataAccess/readCollection.ts
southpolesteve: I sorta understand what is going on now reading the rest of the code. How are connection string users supported? Can they not update their indexes using this new component.

src/Common/PortalNotifications.ts
srinathnarayanan: I am not able to. Not sure why. I copy pasted the code from master but for some reason it still shows it as modified. I checked the line endings too, they are also consistent with what is present in master

src/Common/dataAccess/readCollection.ts
srinathnarayanan: RP doesn't support getting the "index progress" header so this has to be done through the client. Is there anyway we can fetch the cosmos collection using a connectionstring? if not, connectionstring users should not even be able to update their collection in way right now (like ttl, analytical ttl). How does that take place for connection string users? I can use the same pattern.

src/Common/dataAccess/readCollection.ts
srinathnarayanan: will sync with him to make those changes.

src/Common/dataAccess/readCollection.ts
srinathnarayanan: I can instead add a DefaultExperince based override within readCollection and not expose this method. will that work?

src/Explorer/Controls/Settings/SettingsComponent.test.tsx
srinathnarayanan: getPendingNotification generates a Q.Promise which is what is being used here. in settingsv1, the promise is resolved and only then was the tab opened. But in settingsv2, I am trying to push the call into the tab activation for a better experience. Hence this Q.Promise has to be propagated to the new code too

src/Explorer/Controls/Settings/SettingsSubComponents/MongoIndexingPolicy/MongoIndexingPolicyComponent.tsx
srinathnarayanan: will remove

src/Common/dataAccess/readCollection.ts
srinathnarayanan: this I am not sure of. I reused existing patterns.

src/Common/dataAccess/readCollection.ts
srinathnarayanan: synced offline. If user is not using AAD, will show a prompt that they will have to sign in to see the indexing tab.

src/Common/dataAccess/readCollection.ts
srinathnarayanan: Will move to a separate file since we will need both the readCollection through a client and through RP.

src/Common/dataAccess/readCollection.ts
srinathnarayanan: victor added some error handling. I pulled those changes

src/Common/PortalNotifications.ts
swviswan: according to what i see using codeflow, there is some spacing difference. may be the editor you are using has some settings for tabs?

srinathnarayanan: looks like some line ending issues. I checked it out again. The existing file has a Unix style line ending (which shouldn't be commitable, dont known how it got through). My change make it the standard line ending (Widnows style) seen in other files. Let me sync with steve, a rule might have been added after this got commited.

southpolesteve: We don't enforce line endings other than what git does. The default did change at one point. Make sure your local git instance is up to date and setup properly for line endings https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings

I'm ok with checking this in. Prettier v2 has line ending checks and I can PR that upgrade after this PR is merged.

src/Explorer/Controls/Settings/__snapshots__/SettingsRenderUtils.test.tsx.snap
```diff
@@ -310,5 +310,15 @@ exports[`SettingsUtils functions render 1`] = `
   >
     Enable change feed log retention policy to retain last 10 minutes of history for items in the container by default. To support this, the request unit (RU) charge for this container will be multiplied by a factor of two for writes. Reads are unaffected.
   </Text>
+  <Text>
+    This indexing policy editor currently does not support compound indexes. If you need to add a compound index, please add it using a different method. Please refer to
```
swviswan, line 314: >lease add it using a different method [](start = 115, length = 37)

please update text

swviswan (approved):
:shipit:

PR #287, opened 2020-10-16
[ID: 833708][Screen Reader - CosmosDB – New KeySpace] Visual label(K…

Chris-MS-896:
Expected Result:
Visual label(Keyspace id) and aria-label(Database id) should be same for Keyspace id edit field under New KeySpace blade.

current Result: changed

### DIFF
--- a/src/Explorer/Panes/AddDatabasePane.html
+++ b/src/Explorer/Panes/AddDatabasePane.html
@@ -75,7 +75,7 @@
           <input id="database-id" type="text" aria-required="true" autocomplete="off" pattern="[^/?#\\]*[^/?# \\]"
               title="May not end with space nor contain characters '\' '/' '#' '?'"
               size="40" class="collid" data-bind="textInput: databaseId, hasFocus: firstFieldHasFocus, attr: { placeholder: databaseIdPlaceHolder }"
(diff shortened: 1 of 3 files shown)


src/Explorer/Panes/AddDatabasePane.html
```diff
@@ -75,7 +75,7 @@
           <input id="database-id" type="text" aria-required="true" autocomplete="off" pattern="[^/?#\\]*[^/?# \\]"
               title="May not end with space nor contain characters '\' '/' '#' '?'"
               size="40" class="collid" data-bind="textInput: databaseId, hasFocus: firstFieldHasFocus, attr: { placeholder: databaseIdPlaceHolder }"
-              aria-label="Database id" autofocus>
+              aria-label="Keyspace id" autofocus>
```
southpolesteve, line 78: "Keyspace" only applies to Cassandra accounts so you need to change this to ensure it stills says "Database" for other API types

Chris-MS-896: Ok, I see, make sure only the cassandra accounts will change this aria-label

southpolesteve (approved):

ato9000 (approved):

PR #288, opened 2020-10-19
Pass subscriptionId when publishing/accessing published notebooks

gitanuj:
We need to record `subscriptionId` when publishing a notebook, also we want to restrict notebooks to only from a particular `subscriptionId` when accessing `My published work` tab. This change passes the `subscriptionId` as part of the URL when publishing or accessing published notebooks.

### DIFF
--- a/src/Juno/JunoClient.test.ts
+++ b/src/Juno/JunoClient.test.ts
@@ -1,9 +1,12 @@
 import ko from "knockout";
 import { HttpHeaders, HttpStatusCodes } from "../Common/Constants";
-import { IPinnedRepo, JunoClient } from "./JunoClient";
+import { IPinnedRepo, JunoClient, IPublishNotebookRequest } from "./JunoClient";
 import { configContext } from "../ConfigContext";
 import { getAuthorizationHeader } from "../Utils/AuthorizationUtils";
 import { DatabaseAccount } from "../Contracts/DataModels";
+import { updateUserContext, userContext } from "../UserContext";
+
+const sampleSubscriptionId = "subscriptionId";
 
 const sampleDatabaseAccount: DatabaseAccount = {
   id: "id",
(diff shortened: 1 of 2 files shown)


src/Juno/JunoClient.test.ts
```diff
@@ -131,8 +134,16 @@ describe("GitHub", () => {
...
+  const originalUserContext = userContext;
+
+  beforeEach(() => {
+    updateUserContext({
+      subscriptionId: sampleSubscriptionId
+    });
+  });
 
   afterEach(() => {
+    updateUserContext(originalUserContext);
```
southpolesteve, line 146: Does this need to get reset every time? Or can you just do it once in a beforeAll()

src/Juno/JunoClient.test.ts
```diff
@@ -131,8 +134,16 @@ describe("GitHub", () => {
 
 describe("Gallery", () => {
   const junoClient = new JunoClient(ko.observable<DatabaseAccount>(sampleDatabaseAccount));
+  const originalUserContext = userContext;
```
southpolesteve, line 137: This won't copy the object just create a reference, so see below. I think it is not needed.

southpolesteve (approved):

src/Juno/JunoClient.ts
```diff
@@ -291,7 +300,7 @@ export class JunoClient {
   }
 
   public async favoriteNotebook(id: string): Promise<IJunoResponse<IGalleryItem>> {
-    const response = await window.fetch(`${this.getNotebooksAccountUrl()}/gallery/${id}/favorite`, {
+    const response = await window.fetch(`${this.getNotebooksUrl()}/${this.getAccount()}/gallery/${id}/favorite`, {
```
languy, line 303: What's the reason we don't want `${this.getNotebooksAccountUrl()}` since this `${this.getNotebooksUrl()}/${this.getAccount()}` pattern gets reused multiple times?

languy (approved):

src/Juno/JunoClient.test.ts
gitanuj: Makes sense. Updated.

src/Juno/JunoClient.test.ts
gitanuj: My bad. Didn't realize. Instead just copying and restoring only `subscriptionId` now.

src/Juno/JunoClient.ts
gitanuj: Reverted to reinstate that pattern

PR #291, opened 2020-10-20
RU/GB waiver preview feature

notlaforge:
image


src/Explorer/Tabs/DatabaseSettingsTab.html
```diff
@@ -85,6 +85,13 @@
           <span>Learn more about minimum throughput </span>
           <a href="https://docs.microsoft.com/azure/cosmos-db/set-throughput" target="_blank">here.</a>
         </p>
+        <p data-bind="visible: canRequestBelowMinRU">
+          Need to scale below <span data-bind="text: minRUsText"></span> RU/s? Reach out by filling
+          <a
```
ThomasWeiss, line 90: if possible, it would be better to [embed the survey](https://docs.microsoft.com/en-us/dynamics365/customer-voice/embed-web-page) instead of providing a link

PR #299, opened 2020-10-24
Allow users to switch between manual and autoscale for fixed collections

victor-meng:
- removed the `isFixed` check in both new and old settings tab so that the options to switch between manual and autoscale shows up for fixed collections
- updated the message below the text box to inform the users that the max RU for fixed collections is 10000
- updated validation rule so that the max RU cannot exceed 10000 for fixed collections for both autoscale and manual

### DIFF
--- a/src/Explorer/Controls/Settings/SettingsUtils.test.tsx
+++ b/src/Explorer/Controls/Settings/SettingsUtils.test.tsx
@@ -91,7 +91,7 @@ describe("SettingsUtils", () => {
   it("getSanitizedInputValue", () => {
     const max = 100;
     expect(getSanitizedInputValue("", max)).toEqual(0);
-    expect(getSanitizedInputValue("999", max)).toEqual(99);
+    expect(getSanitizedInputValue("999", max)).toEqual(100);
     expect(getSanitizedInputValue("10", max)).toEqual(10);
   });
 });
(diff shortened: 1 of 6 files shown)


southpolesteve (approved):

src/Explorer/Controls/Settings/SettingsUtils.tsx
```diff
@@ -131,13 +131,12 @@ export const parseConflictResolutionProcedure = (procedureFromBackEnd: string):
 };
 
 export const getSanitizedInputValue = (newValueString: string, max: number): number => {
```
srinathnarayanan, line 133: So the scenario this is for is like this:
Max is 100
user input 9 in the textfiled -> maintained as 9
user inputs another 9 in the textfield -> maintained as 99
user inputs another 9 in the textfield -> exceeds max of 100, so keep it as 99.

I think we should retain the behavior of chopping off the most recent input instead of maxing it min(max, input). It will be jumpy otherwise

src/Explorer/Controls/Settings/SettingsUtils.test.tsx
```diff
@@ -91,7 +91,7 @@ describe("SettingsUtils", () => {
   it("getSanitizedInputValue", () => {
     const max = 100;
     expect(getSanitizedInputValue("", max)).toEqual(0);
-    expect(getSanitizedInputValue("999", max)).toEqual(99);
+    expect(getSanitizedInputValue("999", max)).toEqual(100);
```
srinathnarayanan, line 94: change test back to reflect old behavior

srinathnarayanan (approved):

src/Explorer/Controls/Settings/SettingsUtils.tsx
victor-meng: Discussed with Srinath offline. We are going to stick with this behavior for now.

PR #303, opened 2020-10-27
Added extra logging for mongo index update RP call

srinathnarayanan:
- Also changed error -> error.reason for tracefailure


swviswan (approved):
:shipit:

southpolesteve (approved):

PR #313, opened 2020-11-03
Refactor error handling part 2

victor-meng:
- updated `handleError` function:
	1. Switched the `area` and `consoleErrorPrefix` arguments and made `consoleErrorPrefix` optional
	2. Replaced `CosmosError` with `ARMError`
- updated argument type of `getErrorMessage` function to `string | Error`
- updated `Logger.logError` function to only take in error message string instead of error objects
- updated callers of `Logger.logError` to call `getErrorMessage` on the error object/string first
- replaced places that call both `NotificationConsoleUtils.logConsoleMessage` and `Logger.logError` with `handleError`


gitanuj (approved):
GitHub related changes LGTM! 💯

southpolesteve (approved):

PR #315, opened 2020-11-05
Refactor error handling in data explorer Part 3

victor-meng:
- Make sure we pass the error message string instead of an error object when we call `TelemetryProcessor.traceFailure` since TelemetryProcessor will call `JSON.stringify` on the error object which would result in an empty object
- Removed ErrorParserUtility since it only works on specific error types. We can just log the full error message and manually derive information we need from the message.
- Added option to include stack trace in `getErrorMessage`. This is useful for figuring out where the client side script errors are coming from.
- Some minor refactors


src/Common/ErrorHandlingUtils.ts
```diff
@@ -21,9 +21,15 @@ export const handleError = (error: string | ARMError | Error, area: string, cons
...
-  const errorMessage = typeof error === "string" ? error : error.message;
-  return replaceKnownError(errorMessage);
+export const getErrorMessage = (error: string | Error, includeStackTrace?: boolean): string => {
+  if (typeof error === "string") {
+    return replaceKnownError(error);
+  }
+
+  const errorMessage = replaceKnownError(error.message);
+  return includeStackTrace && error.stack
+    ? `Error message: ${errorMessage}\n Stack trace: ${error.stack}`
```
srinathnarayanan, line 31: can we make an object like 
error: {message: "....", stacktrace: "..."} 
in the telemetry message? Will be easier to parse out when we are looking at kusto logs.

srinathnarayanan (approved):

src/Explorer/Controls/Settings/SettingsComponent.tsx
```diff
@@ -573,7 +573,7 @@ export class SettingsComponent extends React.Component<SettingsComponentProps, S
           defaultExperience: this.container.defaultExperience(),
           dataExplorerArea: Constants.Areas.Tab,
           tabTitle: this.props.settingsTab.tabTitle(),
-          error: reason.message
+          error: getErrorMessage(error, true)
```
southpolesteve, line 576: I agree with @srinathnarayanan's other comment. What about adding a new key to `traceFailure` that adds the stack?

`stack: getErrorStack(error)` can return undefined if it doesn't exist.

src/Common/ErrorHandlingUtils.ts
victor-meng: We want to avoid passing an object to the telemetry processor since it will get stringified. I think what Steve suggested below is a good alternative. I will add a new key for stack trace in traceFailure.

src/Explorer/Controls/Settings/SettingsComponent.tsx
victor-meng: Done

src/Explorer/Panes/DeleteDatabaseConfirmationPane.ts
```diff
@@ -108,20 +108,21 @@ export default class DeleteDatabaseConfirmationPane extends ContextualPaneBase {
           this.databaseDeleteFeedback("");
         }
       },
-      (reason: unknown) => {
+      (error: any) => {
```
southpolesteve, line 111: Should make this "unknown" if possible

southpolesteve (approved):

PR #318, opened 2020-11-11
Added loadAutopilotOfferWithRetry

srinathnarayanan:
A spurious bug that seems to appear when updating offer from manual -> autopilot is that the updated offer does not immediately reflect in the Offer Object received after running updateOffer(0.
So an 'offerAutopilotSettings' field is not present in the returned offer. This surfaces as a  **Cannot   read property 'maxThroughput' of undefined** error when the update for offer is done in the Scale subtab.

This PR adds a retry mechanism to loadOffer when this field is not present, and surfaces a more verbose message when the retries have been exhausted.

A spurious bug that seems to appear when updating offer from manual -> autopilot is that the updated offer does not immediately reflect in the Offer Object received after running updateOffer(0.
So an 'offerAutopilotSettings' field is not present in the returned offer. This surfaces as a  **Cannot   read property 'maxThroughput' of undefined** error when the update for offer is done in the Scale subtab.

This PR adds a retry mechanism to loadOffer when this field is not present, and surfaces a more verbose message when the retries have been exhausted.

### DIFF
--- a/src/Common/Constants.ts
+++ b/src/Common/Constants.ts
@@ -437,3 +437,8 @@ export class TerminalQueryParams {
   public static readonly SubscriptionId = "subscriptionId";
   public static readonly TerminalEndpoint = "terminalEndpoint";
 }
+
+export class LoadOfferRetry {
+  public static readonly MaxRetries = 5;
+  public static readonly RetryAfterMilliSeconds = 5000;
+}
--- a/src/Explorer/Tree/Collection.ts
+++ b/src/Explorer/Tree/Collection.ts
@@ -1343,7 +1343,7 @@ export default class Collection implements ViewModels.Collection {
       retryCount++;
     }
     throw new Error(
-      "Max Retries for loading offer are completed. Offer does not have an offerAutopilotSettings field."
+      "Max retries for loading offer are completed. Offer does not have an Autopilot settings field."
     );
   }
 
--- a/src/Explorer/Tree/Collection.ts
+++ b/src/Explorer/Tree/Collection.ts
@@ -1331,6 +1331,22 @@ export default class Collection implements ViewModels.Collection {
     return this.container.findDatabaseWithId(this.databaseId);
   }
 
+  public async loadAutopilotOfferWithRetry(): Promise<DataModels.Offer> {
+    let retryCount = 0;
+    while (retryCount < Constants.LoadOfferRetry.MaxRetries) {
+      if (this.offer().content.offerAutopilotSettings) {
+        return this.offer();
+      } else {
+        await new Promise(resolve => setTimeout(resolve, Constants.LoadOfferRetry.RetryAfterMilliSeconds));
+        await this.loadOffer();
+      }
+      retryCount++;
+    }
+    throw new Error(
+      "Max Retries for loading offer are completed. Offer does not have an offerAutopilotSettings field."
+    );
+  }
+
   public async loadOffer(): Promise<void> {
     if (!this.container.isServerlessEnabled() && !this.offer()) {
       this.container.isRefreshingExplorer(true);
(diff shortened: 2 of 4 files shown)


src/Common/Constants.ts
```diff
@@ -437,3 +437,8 @@ export class TerminalQueryParams {
   public static readonly SubscriptionId = "subscriptionId";
   public static readonly TerminalEndpoint = "terminalEndpoint";
 }
+
+export class LoadOfferRetry {
```
southpolesteve, line 441: Move this into where it is called. Trying to avoid adding to Constants for things that not widely used

src/Explorer/Tree/Collection.ts
```diff
@@ -1331,6 +1331,20 @@ export default class Collection implements ViewModels.Collection {
     return this.container.findDatabaseWithId(this.databaseId);
   }
 
+  public async loadAutopilotOfferWithRetry(): Promise<DataModels.Offer> {
+    let retryCount = 0;
```
southpolesteve, line 1335: We use p-retry in other places which should do all this for you https://www.npmjs.com/package/p-retry

srinathnarayanan: I checked it out. It gets a little cluttered when 1) Specific retry conditions need to be checked 2) We still have to add a sleep/delay/settimeout for delays between retries. i don't see it being much better than the existing implementation

src/Explorer/Tree/Collection.ts
```diff
@@ -1331,6 +1331,20 @@ export default class Collection implements ViewModels.Collection {
...
+    while (retryCount < Constants.LoadOfferRetry.MaxRetries) {
+      if (this.offer().content.offerAutopilotSettings) {
```
swviswan, line 1337: We do this retry since the RP API for updateoffer doesnt show the right data in its response?

if so, does the RP API owner know about this?

src/Explorer/Tree/Collection.ts
srinathnarayanan: synced offline, going with the existing pattern

src/Explorer/Tree/Collection.ts
srinathnarayanan: Nope, we need to track it in the uber excel sheet.

southpolesteve (approved):

src/Explorer/Tree/Collection.ts
swviswan: Can you please respond to the email thread we have for mongo editor that you see this issue.  

---
In reply to: [521598445](https://github.com/Azure/cosmos-explorer/pull/318#discussion_r521598445) [](ancestors = 521598445)

src/Explorer/Tree/Collection.ts
```diff
@@ -41,8 +41,11 @@ import { userContext } from "../../UserContext";
 import TabsBase from "../Tabs/TabsBase";
 import { fetchPortalNotifications } from "../../Common/PortalNotifications";
 import { getErrorMessage, getErrorStack } from "../../Common/ErrorHandlingUtils";
+import promiseRetry from "p-retry";
```
languy, line 44: Do you still need this?

src/Explorer/Tree/Collection.ts
```diff
@@ -1331,6 +1334,20 @@ export default class Collection implements ViewModels.Collection {
     return this.container.findDatabaseWithId(this.databaseId);
   }
 
+  public async loadAutopilotOfferWithRetry(): Promise<DataModels.Offer> {
+    let retryCount = 0;
+    while (retryCount < Collection.LoadOfferMaxRetries) {
+      if (this.offer().content.offerAutopilotSettings) {
+        return this.offer();
+      } else {
+        await new Promise(resolve => setTimeout(resolve, Collection.LoadOfferRetryAfterMilliSeconds));
```
languy, line 1343: It would be nice to have exponential retries. This doesn't block the main UI thread, right?
What was the issue with p-retry, it seemed like a good idea to not re-invent the wheel? Otherwise, please add unit tests. I would also suggest to make the retry logic re-usable: could be useful somewhere else.

languy (approved):

PR #321, opened 2020-11-12
Use Puppeteer for Emulator Test

southpolesteve:
Switches our last cypress test over to puppeteer and we can get rid of cypress


zfoster (approved):

zfoster:
could maybe just delete that skipped test and the old code?

languy (approved):

PR #322, opened 2020-11-13
Bug fix: match monaco-editor version with @nteract/monaco-editor

languy:
Opening notebook (which contains code cell), then "Items"-> document editor is broken. Our JsonEditor component will hang at `monaco.editor.create()`.

Matching the `monaco-editor` version with `@nteract/monaco-editor` fixes it.

I looked into using one, but we cannot rely nteract, because it does not get loaded if notebook isn't enabled. Forcing nteract to use ours (if at all possible) isn't a good idea, since their code is tuned to their version.

For now, we'll have to keep the versions in sync.


southpolesteve (approved):

src/Explorer/Controls/Editor/EditorComponent.ts
```diff
@@ -52,12 +52,14 @@ class EditorViewModel extends JsonEditorViewModel {
   protected registerCompletionItemProvider() {
     let sqlCompletionItemProvider = new SqlCompletionItemProvider();
     if (EditorViewModel.providerRegistered.indexOf("sql") < 0) {
-      monaco.languages.registerCompletionItemProvider("sql", sqlCompletionItemProvider);
+      // TODO Remove when @azure/cosmos-language-service is updated
```
languy, line 55: I have a [PR open](https://github.com/Azure/cosmos-sql-language-service/pull/5) to fix this.

zfoster (approved):

PR #323, opened 2020-11-13
Added testExplorer and notebooks UI automated tests

srinathnarayanan:
This PR does the following
- Exposes a testExplorer landing page that hosts the explorer.html iframe
- Adds AAD authentication to get auth token to access notebooks and to use arm APIs to get the database account info
- This Info is used to initializae the data explorer iframe, similar to how the portal initializes it
- Adds a test that uploads, opens and deletes a notebook


src/Explorer/Explorer.ts
```diff
@@ -1742,6 +1744,10 @@ export default class Explorer {
       return;
     }
 
+    if (event.data.data.type === MessageTypes.HideConnectScreen) {
```
southpolesteve, line 1747: Why is this needed? If you start explorer with "platform=Portal" in the query string shouldn't it work?

src/TestExplorer.ts
```diff
@@ -0,0 +1,153 @@
+import * as ko from "knockout";
```
southpolesteve, line 1: Can you rework this file without knockout? Would strongly prefer to avoid adding new knockout code.

src/TestExplorer.ts
```diff
@@ -0,0 +1,153 @@
...
+      this.initTestExplorer();
+    };
+    window.addEventListener("message", this.handleMessage.bind(this), false);
+  }
+
+  private parseUrlParams = (): void => {
+    window.location.search
+      .substr(1)
+      .split("&")
+      .forEach(item => {
```
southpolesteve, line 44: This API is built in: new URLSearchParams(window.location.search)

src/TestExplorer.ts
```diff
@@ -0,0 +1,153 @@
...
+        const tmp = item.split("=");
+        const value = decodeURIComponent(tmp[1]);
+        switch (tmp[0]) {
```
southpolesteve, line 47: If you use the above built in APIs, this should become much simpler. You can call `get` for each key

src/TestExplorer.ts
```diff
@@ -0,0 +1,153 @@
...
+    };
+    window.postMessage(content, window.location.href);
+
+    const hideConnectContent = {
+      type: MessageTypes.HideConnectScreen
+    };
+    window.postMessage(hideConnectContent, window.location.href);
+  }
+
+  private sendMessageToExplorerFrame(data: unknown): void {
```
southpolesteve, line 137: Same as other file. Unclear why this is needed if loading in Portal mode.

src/testExplorer.html
```diff
@@ -0,0 +1,15 @@
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="height=device-height, width=device-width, initial-scale=1.0" />
+    <title>Azure Cosmos DB</title>
+    <link rel="shortcut icon" href="images/CosmosDB_rgb_ui_lighttheme.ico" type="image/x-icon" />
+  </head>
+
+  <body>
+    <switch-directory-pane params="{data: switchDirectoryPane}"></switch-directory-pane>
```
southpolesteve, line 10: Is this component needed? Again, trying to avoid new KO usage.

src/testExplorer.html
```diff
@@ -0,0 +1,15 @@
...
+
+    <iframe id="explorerMenu" name="explorer" class="iframe" title="explorer" src="explorer.html?v=1.0.1&platform=Test">
```
southpolesteve, line 12: I see now you are using `platform=Test`. Is there a reason for needing a new one? I think we should be using platform=Portal so the underlying Explorer code knows what to do. We have a lot of checks for == Portal for behavior

test/notebooks/notebookTestUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+  const notebooksTestRunnerApplicationId = process.env.NOTEBOOKS_TEST_RUNNER_TENANT_ID;
+  const notebooksTestRunnerClientId = process.env.NOTEBOOKS_TEST_RUNNER_CLIENT_ID;
+  const notebooksTestRunnerClientSecret = process.env.NOTEBOOKS_TEST_RUNNER_CLIENT_SECRET;
+  const notebooksAccountName = process.env.NOTEBOOKS_ACCOUNT_NAME;
+  const notebooksAccountKey = process.env.NOTEBOOKS_ACCOUNT_KEY;
+  const notebooksAccountSubscriptonId = process.env.NOTEBOOKS_ACCOUNT_SUBSCRIPTION_ID;
+  const notebooksAccountResourceGroup = process.env.NOTEBOOKS_ACCOUNT_RESOURCE_GROUP;
+
+  const prodUrl = `https://localhost:1234/testExplorer.html?
+${TestExplorerParams.notebooksTestRunnerApplicationId}=${encodeURI(notebooksTestRunnerApplicationId)}&
```
southpolesteve, line 22: You can do all of this with `URLSearchParams`:

```
const url = new URL("testExplorer.html", "https://localhost:1234");
url.searchParams.append(TestExplorerParams.notebooksTestRunnerApplicationId, notebooksTestRunnerApplicationId);
```

src/TestExplorer.ts
```diff
@@ -0,0 +1,153 @@
...
+      this.notebooksTestRunnerApplicationId,
+      this.notebooksTestRunnerClientId,
+      this.notebooksTestRunnerClientSecret
+    );
+    const token = await credentials.getToken("https://management.core.windows.net/.default");
+    return token.token;
+  }
+
+  private async getDatabaseAccount(token: string): Promise<DatabaseAccountsGetResponse> {
+    const client = new CosmosDBManagementClient(new CustomSigner(token), this.notebooksAccountSubscriptonId);
```
southpolesteve, line 90: Ugh. I can't believe our arm libraries don't work with the new @azure/identity library. Not your issue, but I will ping someone in central SDK about this. Frustrating.

test/notebooks/notebookTestUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+export const NOTEBOOK_OPERATION_DELAY = 5000;
+export const RENDER_DELAY = 1000;
+
+let testExplorerFrame: Frame;
+export const getTestExplorerFrame = async (): Promise<Frame> => {
+  if (testExplorerFrame) {
+    return testExplorerFrame;
+  }
+
+  const notebooksTestRunnerApplicationId = process.env.NOTEBOOKS_TEST_RUNNER_TENANT_ID;
```
southpolesteve, line 13: Where are all these `process.env` variables expected to be loaded from? Do they need to be added to .env.example so they can be run locally?

test/notebooks/notebookTestUtils.ts
```diff
@@ -0,0 +1,90 @@
+import { ElementHandle, Frame } from "puppeteer";
```
southpolesteve, line 1: Remove this file and inline these all in the test file. I think there is no benefit to abstracting this, especially for end to end tests. It adds another layer of indirection. Sandi Metz articulates my opinions on this very well https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

southpolesteve: By inlne, I mean make the test file one big function. No need to have these all broken up into small functions.

test/notebooks/notebookTestUtils.ts
srinathnarayanan: Yes, they need to be injected in the environment. WIll update .env.example file

src/Explorer/Explorer.ts
srinathnarayanan: changed to "platform=Portal" and removed this.

src/TestExplorer.ts
srinathnarayanan: changed to "platform=Portal" and removed this.

test/notebooks/notebookTestUtils.ts
srinathnarayanan: My thought was, for any notebook UI test, it should be framed as 
upload notebook -> do something (open / run / change kernel and run etc) -> delete notebook 
Hence this abstraction. Upload notebook and delete notebook should be called in the setup and cleanup for a class of tests. I will add one more test to this PR to make that clear.

test/notebooks/testExplorer/TestExplorer.ts
```diff
@@ -0,0 +1,138 @@
...
+};
+
+const getDatabaseAccount = async (
+  token: string,
+  notebooksAccountSubscriptonId: string,
+  notebooksAccountResourceGroup: string,
+  notebooksAccountName: string
+): Promise<DatabaseAccountsGetResponse> => {
+  const client = new CosmosDBManagementClient(new CustomSigner(token), notebooksAccountSubscriptonId);
+  return await client.databaseAccounts.get(notebooksAccountResourceGroup, notebooksAccountName);
```
southpolesteve, line 49:
```suggestion
  return client.databaseAccounts.get(notebooksAccountResourceGroup, notebooksAccountName);
```
No await needed. I should probably add a lint rule for this.

southpolesteve (approved):
Looking good! Thanks for tackling the test explorer setup. I plan to use that elsewhere.

test/notebooks/testNotebooks/GettingStarted.ipynb
```diff
@@ -0,0 +1,110 @@
+{
+  "cells": [
```
swviswan, line 2: >"cells": [ [](start = 2, length = 10)

Nice. Can you rename this file to something so that if we see logs for this file we can quickly identify that it is our test and not the production NB.

swviswan (approved):
:shipit:

test/notebooks/testNotebooks/GettingStarted.ipynb
srinathnarayanan: We wont see any logs here for running of the file. The errors/outputs will be saved to the file itself.
Running the file and parsing out outputs/checking for errors will be more elaborate, so pushed it to the next set of tests we may have to write.

srinathnarayanan: We will have to run the file, download it, parse the json and then validate results.

PR #328, opened 2020-11-21
Remove RUPM

victor-meng:
Remove all RUPM code


src/Explorer/Controls/Settings/__snapshots__/SettingsRenderUtils.test.tsx.snap
```diff
@@ -69,15 +69,15 @@ exports[`SettingsUtils functions render 1`] = `
      
     <b>
       ¥
-      1.29
```
southpolesteve, line 72: Are these supposed to change? This makes me think something is wrong.

victor-meng: Yes the number should be lower now since I removed the `rupmCharge` when calculating the hourly price (PricingUtils.computeRUUsagePriceHourly). In the SettingsRenderUtils.test, we were calling `getEstimatedSpendElement` with `rupmEnabled` set to true so the price included the `rupmCharge`. The daily and monthly prices are both calculated based on the hourly price so it makes sense that these numbers are lower as well.

southpolesteve (approved):

languy (approved):

PR #342, opened 2020-12-16
Cost Estimate Changes

vchske:
Changed cost estimate to table format with a comparison between current vs update cost estimates.


vchske:
@jessica-tsui 
image
image

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
```diff
@@ -32,10 +31,41 @@ import {
...
+  IColumn,
+  SelectionMode,
+  DetailsListLayoutMode,
+  IDetailsRowProps,
+  DetailsRow,
+  IDetailsColumnStyles
 } from "office-ui-fabric-react";
 import { isDirtyTypes, isDirty } from "./SettingsUtils";
 
+export interface IEstimatedSpendingDisplayProps {
```
southpolesteve, line 45: "I" prefix is generally not used in TypeScript. This stack overflow has a good explanation https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript. If you wanna see a bunch of in depth arguing about why check this thread https://github.com/Microsoft/TypeScript-Handbook/issues/121

I had a lint rule for this, but appears it is not working correctly. This is my fault. So don't let this comment block merging. I can clean these up when I fix linting.

southpolesteve (approved):

gitanuj (approved):

ato9000 (approved):

src/Utils/PricingUtils.ts
```diff
@@ -203,7 +192,7 @@ export function getEstimatedSpendHtml(
   regions: number,
   multimaster: boolean
 ): string {
-  const hourlyPrice: number = computeRUUsagePriceHourly(serverId, throughput, regions, multimaster);
+  const hourlyPrice: number = computeRUUsagePriceHourly(serverId, throughput, regions, multimaster, false);
```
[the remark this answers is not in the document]
southpolesteve: If you are feeling adventurous, just make these objects and destructure the parameters https://mariusschulz.com/blog/typing-destructured-object-parameters-in-typescript

Much clearer what is going on. Positional booleans 🚫

src/Explorer/Controls/Settings/SettingsRenderUtils.tsx
vchske: I changed the names of the interfaces. I ain't going to make more work for you!

ato9000: I saws this, but he other interfaces in this file already are using this convention. Office-UI also uses this convention. Me, personally, I am not a fan of the TypeScript conventions, and I can argue against all their pro arguments. Especially, they claim this will produce better names, but what I have seen in practice is people give the same name to interfaces and classes and put them in different modules to differentiate between them-- this is worse than having a I prefix because now you cannot tell the difference unless you hover in your IDE.

This is a religious war that I don't want to fight, but for consistency, we already allow I in this same file so I would suggest not to have to rename interfaces in this PR

vchske: Too late. I also don't agree with the reasons to not use I but it isn't a hill I'm even remotely willing to die on. I do prefer to be consistent with the convention used by the other members of the team. So I think it's appropriate to change it. We can't change React's type names but we can remain consistent within ourselves.

src/Utils/PricingUtils.ts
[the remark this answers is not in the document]
vchske: Talked with Steve, and while I think there is a limit to the usefulness of destructured parameters, his argument that requesting a comment was the smell test convinced me. Changing this.

src/Utils/PricingUtils.ts
```diff
@@ -159,28 +163,13 @@ export function getAutoPilotV3SpendHtml(maxAutoPilotThroughputSet: number, isDat
...
-}
-
 export function getEstimatedAutoscaleSpendHtml(
   throughput: number,
   serverId: string,
   regions: number,
   multimaster: boolean
 ): string {
-  const hourlyPrice: number = computeAutoscaleUsagePriceHourly(serverId, throughput, regions, multimaster);
+  const hourlyPrice: number = computeRUUsagePriceHourly(serverId, throughput, regions, multimaster, true);
```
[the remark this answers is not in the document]
vchske: Refactored to use an args interface.

jessica-tsui (dismissed):
Thanks for explaining why you went with the severe warning message bar. However I don't think this is truly a severe warning - this should really only be applied in situations where users will likely cause severe error such as deleting container/some other errors. We should keep using the typical warning message bar even though it's used in places like indexing policy tab. Other than that, everything looks good to me

PR #351, opened 2020-12-22
Replace $.ajax with fetch

victor-meng:
Replace $.ajax with fetch


southpolesteve (approved):

notlaforge (approved):
Less `Q`, less `$`... what's not to like?

PR #352, opened 2020-12-22
Switch to ESBuild for transpiling

southpolesteve:


southpolesteve:
Not currently viable because of decorator usage in Self Serve. Will need to revisit.

PR #360, opened 2021-01-03
Migrated Hosted Explorer to React

southpolesteve:
This PR migrates the Hosted Explorer wrapper from KO/React mix entirely to React. 

High level list of changes:

- Remove ADAL and Replace with MSAL v1. Ideally we would move to MSAL v2, but I ran into issues with it. It needs to use the [new auth code flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/tutorial-v2-javascript-auth-code) but the tokens it returned were not useable for GraphV1 or ARM APIs. Will need to revisit moving to v2 later. I think its possible if we make some configuration changes in the first party identity portal
- Remove KO entirely from Hosted Explorer
- Migrated existing React components in Hosted Explorer to use hooks
- Refactor and split up a few components in Hosted Explorer
- Move loading screen and connection string form up to Hosted Explorer layer. These currently are responsibility of the child Explorer iframe which leads to a significant amount of complexity. With this change there is no longer any cross iframe communication required for Hosted Explorer
- Hosted Explorer lives in the same domain as explorer.html in hosted mode, so instead of passing iframe messages, we can configure explorer.html by setting window variable directly.
- Adds custom hooks for a few hosted resources

TODO:
- [x] Fix lint
- [x] Cleanup Types
- [x] Fix mongo/cassandra accounts for AAD login. They are working partially, but the passed down tokens are not working for some reason.
- [x] Fix switching AAD tenant. I have not yet figured out how to do this with MSAL
- [x] Make sure resource token logins work. Ideally write a test for this case.
- [ ] ~~Load Database Account in Hosted frame and pass down to explorer iframe~~.
- [x] Add component unit tests
- [x] Make sure the portal still works!
  - [x] SQL
  - [x] Graph
  - [x] Cassandra
  - [x] Mongo
  - [x] Tables
- [x] Cassandra tables load no data


src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <div className="connectExplorer">
+          <p className="connectExplorerContent">
+            <img src="images/HdeConnectCosmosDB.svg" alt="Azure Cosmos DB" />
+          </p>
+          <p className="welcomeText">Welcome to Azure Cosmos DB</p>
+          {isFormVisible ? (
+            <form
+              id="connectWithConnectionString"
+              onSubmit={async event => {
+                event.preventDefault();
```
languy, line 38: nit: move to separate function?

languy (approved):

victor-meng (approved):

PR #363, opened 2021-01-04
Add two files of notebook component in Matser

Chris-MS-896:
Add two files
  "/home/steve/cosmos-explorer/src/Explorer/Notebook/NotebookComponent/__mocks__/rx-jupyter.ts",
  "/home/steve/cosmos-explorer/src/Explorer/Notebook/NotebookContentClient.ts",


src/Explorer/Notebook/NotebookComponent/__mocks__/rx-jupyter.ts
```diff
@@ -1,6 +1,6 @@
 import { Observable, of } from "rxjs";
 import { AjaxResponse } from "rxjs/ajax";
-import { ServerConfig } from "rx-jupyter";
+import { ServerConfig } from "@nteract/types";
```
ato9000, line 3: Check with Laurent

ato9000 (changes requested):

src/Explorer/Notebook/NotebookComponent/__mocks__/rx-jupyter.ts
Chris-MS-896: Hi, Laurent, if we "import { ServerConfig } from "rx-jupyter"; we will meet TS2459 error

southpolesteve: @languy

languy: The full error is: `src/Explorer/Notebook/NotebookComponent/__mocks__/rx-jupyter.ts(3,10): error TS2459: Module '"../../../../../node_modules/rx-jupyter/lib"' declares 'ServerConfig' locally, but it is not exported.`

It's mentioned in line 14: ```typescript
export const sessions = {
  create: (serverConfig: ServerConfig, body: object): Observable<AjaxResponse> => of(fakeAjaxResponse),
  __setResponse: (response: AjaxResponse) => {
    fakeAjaxResponse = response;
  },
  createSpy: undefined as any
};
```
The actual `serverConfig` is never used, so typescript probably doesn't export it.
Replace with: `create(serverConfig: unknown, body: object)` so you don't have to worry about it.

Comment (Chris-MS-896) [reply src/Explorer/Notebook/NotebookComponent/__mocks__/rx-jupyter.ts:3]: Ok, thanks

File: src/Explorer/Notebook/NotebookContentClient.ts [hunk 2]
```diff
@@ -117,8 +117,11 @@ export class NotebookContentClient {
 
   private async checkIfFilepathExists(filepath: string): Promise<boolean> {
     const parentDirPath = NotebookUtil.getParentPath(filepath);
-    const items = await this.fetchNotebookFiles(parentDirPath);
-    return items.some(value => FileSystemUtil.isPathEqual(value.path, filepath));
+    if (parentDirPath) {
+      const items = await this.fetchNotebookFiles(parentDirPath);
+      return items.some(value => FileSystemUtil.isPathEqual(value.path, filepath));
+    }
+    return undefined as any;
```
ato9000, line 124: we should return either false or throw an error

ato9000 (changes requested):

src/Explorer/Notebook/NotebookContentClient.ts
Chris-MS-896: Sure, I will changed that

src/Explorer/Notebook/NotebookContentClient.ts
```diff
@@ -189,7 +192,7 @@ export class NotebookContentClient {
         const dir = xhr.response;
         const item = NotebookUtil.createNotebookContentItem(dir.name, dir.path, dir.type);
         item.parent = parent;
-        parent.children.push(item);
+        parent.children!.push(item);
```
southpolesteve, line 195: In general we should avoid `!` as it may lead to a runtime error. Instead use `?` here. I think that will give you the same result but safer

southpolesteve (approved):
Left one change. But approved after that.

ato9000 (approved):
please address Steve's comments

src/Explorer/Notebook/NotebookContentClient.ts
Chris-MS-896: I will change it after do research

PR #373, opened 2021-01-06
Add all arm request related files to Matser

Chris-MS-896:
Add 18 files related to arm request files to Matser


src/Utils/arm/request.ts
```diff
@@ -66,7 +66,7 @@ export async function armRequest<T>({
   const response = await window.fetch(url.href, {
     method,
     headers: {
-      Authorization: userContext.authorizationToken
+      Authorization: !userContext.authorizationToken ? "" : userContext.authorizationToken
```
southpolesteve, line 69: This is not correct. We must always have an authorization token set. Can we throw an error earlier in the function?

Something like: `if (!userContext.authorizationToken) { throw new Error("No token provided") }`

src/Utils/arm/request.ts
```diff
@@ -100,7 +100,7 @@ export async function armRequest<T>({
 async function getOperationStatus(operationStatusUrl: string) {
   const response = await window.fetch(operationStatusUrl, {
     headers: {
-      Authorization: userContext.authorizationToken
+      Authorization: !userContext.authorizationToken ? "" : userContext.authorizationToken
```
southpolesteve, line 103: Same as other one

southpolesteve (changes requested):

src/Utils/arm/request.ts
Chris-MS-896: OK,I will fix that

src/Utils/arm/request.ts
Chris-MS-896: I make the unit test and change in minor changes

ato9000 (approved):

southpolesteve (approved):

PR #378, opened 2021-01-08
Warn on SubQuery

southpolesteve:
Cosmos does not support non-correlated subqueries https://docs.microsoft.com/en-us/azure/cosmos-db/sql-query-subquery But unfortunately there is no way for a user to know this as the query still executes and returns results. Query team is working on supporting this, but as a stop gap we can use a simple regex to try and direct users to the sub query docs.

image

### DIFF
--- a/src/Explorer/Tabs/QueryTab.html
+++ b/src/Explorer/Tabs/QueryTab.html
@@ -11,6 +11,15 @@
       Start by writing a Mongo query, for example: <strong>{'id':'foo'}</strong> or <strong>{ }</strong> to get all the
       documents.
     </div>
+    <div class="warningErrorContainer" aria-live="assertive" data-bind="visible: maybeSubQuery">
+      <div class="warningErrorContent">
+          <span><img class="paneErrorIcon" src="/error_red.svg" alt="Error"></span>
+          <span class="warningErrorDetailsLinkContainer">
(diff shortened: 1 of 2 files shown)


Chris-MS-896 (approved):

victor-meng (approved):

src/Explorer/Tabs/QueryTab.html
```diff
@@ -11,6 +11,17 @@
       Start by writing a Mongo query, for example: <strong>{'id':'foo'}</strong> or <strong>{ }</strong> to get all the
       documents.
     </div>
+    <div class="warningErrorContainer" aria-live="assertive" data-bind="visible: maybeSubQuery">
+      <div class="warningErrorContent">
+        <span><img class="paneErrorIcon" src="/Info.svg" alt="Error"/></span>
```
victor-meng, line 16: I think info_color.svg is better for this case:

image

victor-meng (approved):

src/Explorer/Tabs/QueryTab.html
southpolesteve: Agreed

southpolesteve:
``` suggestion
        <span><img class="paneErrorIcon" src="/info_color.svg" alt="Error"/></span>
```

PR #386, opened 2021-01-13
Move database settings tab to react

victor-meng:
- Created database settings tab V2 which uses the same `SettingsComponent` as the collection settings tab V2.
- Made changes in `SettingsComponent` and `ScaleComponent` to support the database code path.
- Added the `min` property for the autoscale/manual throughput textfield so that pressing the down arrow key won't make the throughput go below the minimum.
- Added a feature flag `enabledbsettingsV1` in case there's any issue with V2 and need to revert back to V1


src/Explorer/Tree/Collection.ts
```diff
@@ -1052,6 +1045,41 @@ export default class Collection implements ViewModels.Collection {
     return deferred.promise;
   };
 
+  public async getPendingThroughputSplitNotification(): Promise<DataModels.Notification> {
```
southpolesteve, line 1048: Is this still used? I thought this data came down via a header in the offers call? I could be wrong here.

victor-meng: From the code it looks like it's still being used, but I'm not sure if it's actually needed in any scenario.  This function fetches notification from the portal by calling `fetchPortalNotifications` so it's different from the `offerReplacePending` header.

I think we should keep it for now just to be safe and I can investigate whether we actually need this or not.

southpolesteve (approved):

PR #390, opened 2021-01-15
Remove Q from ViewModels

notlaforge:
I got cold feet at the thought of merging #324 in one go, so I'm going to split it into smaller chunks and keep rebasing the large one until there's no more Q.


southpolesteve (approved):

src/Explorer/Tree/Collection.ts
```diff
@@ -1119,46 +1104,35 @@ export default class Collection implements ViewModels.Collection {
...
-        const pendingNotification = _.find(notifications, (notification: DataModels.Notification) => {
-          const throughputUpdateRegExp: RegExp = new RegExp("Throughput update (.*) in progress");
-          return (
-            notification.kind === "message" &&
-            notification.collectionName === this.id() &&
-            notification.description &&
-            throughputUpdateRegExp.test(notification.description)
-          );
-        });
+    const notifications = await fetchPortalNotifications();
```
notlaforge, line 1112: I forgot to handle this exception grrr

src/Explorer/Tree/Collection.ts
```diff
@@ -1119,46 +1104,35 @@ export default class Collection implements ViewModels.Collection {
...
+    if (!notifications) {
```
victor-meng: The original condition is `!notifications || notifications.length === 0`. Do we no longer need to check for empty array?

src/Explorer/Tree/ResourceTokenCollection.ts
```diff
@@ -41,9 +41,9 @@ export default class ResourceTokenCollection implements ViewModels.CollectionBas
     this.isCollectionExpanded = ko.observable<boolean>(true);
   }
 
-  public expandCollection(): Q.Promise<void> {
+  public async expandCollection(): Promise<void> {
```
victor-meng, line 44: Same as above, looks like this function should be synchronous.

victor-meng (approved):

src/Explorer/Tree/Collection.ts
notlaforge: I don't think so, because `[].find()` will return `undefined`.

src/Explorer/Tree/Collection.ts
```diff
@@ -254,9 +253,9 @@ export default class Collection implements ViewModels.Collection {
     });
   }
 
-  public expandCollection(): Q.Promise<any> {
+  public async expandCollection(): Promise<any> {
```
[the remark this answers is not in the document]
notlaforge: The problem is this:

 ```js
export interface CollectionBase extends TreeNode {
  expandCollection(): Promise<any>;
}
```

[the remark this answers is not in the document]
notlaforge: So, I could change the function to no longer be `async`, at the cost of either inverting the initial `if` logic or adding two `return Promise.resolve()` statements. Worth it?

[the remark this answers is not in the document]
notlaforge: (or changing the interface, of course, but I dunno why the interface is the way it is)

PR #396, opened 2021-01-21
Add basic static code metrics

southpolesteve:
We need better progress tracking to some of our big long term goals like "Migrate off Knockout". These is can tough to track directly, but we can make a reasonable proxy metric and statically analyze our code. 

This PR adds a few simple metrics: 
- Number of HTML files 
  - Should go **down** as we remove knockout
- Number of unlinted files
  - Should go **down** over time
- Number of strict mode files
  - Should go **up** over time

Other ideas for future metrics:
- Files importing "ko"
- Files importing "jquery"
- Total build size
- Total number of files
- Test coverage metrics
- React component count
- PR submission to merge time
- Deployments per day
- Most touched files/code


utils/codeMetrics.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-console */
+const fs = require("fs");
+const fg = require("fast-glob");
+const appInsights = require("applicationinsights");
+appInsights.setup("a19ced82-51c3-4fce-9d26-3b13a77a70be").start();
```
swviswan, line 5: >a19ced82-51c3-4fce-9d26-3b13a77a70be [](start = 19, length = 36)

Not a good idea to put it in open source code. please remove it and rotate this value.

southpolesteve: Struggling to find a way to regen this key. Best I can tell App Insights does not consider this a secret. https://github.com/microsoft/ApplicationInsights-JS/issues/281#issuecomment-696305726

But I don't see any reason not to put it in an env variable. I will investigate.

swviswan: The risk is if someone else grabs this key, they can start sending the events to this resource. In essence, it should be treated as a secret. In fact, I remember reading an official Microsoft post where app insights even recommends not to use this value directly in client code. Rather the recommendation is we should have it in our server and send to client and use it. 

---
In reply to: [562164083](https://github.com/Azure/cosmos-explorer/pull/396#discussion_r562164083) [](ancestors = 562164083)

swviswan: Lets sync offline. 

---
In reply to: [562428205](https://github.com/Azure/cosmos-explorer/pull/396#discussion_r562428205) [](ancestors = 562428205,562164083)

languy (approved):

PR #403, opened 2021-01-26
Cover null of notebookServerEndpoint

Chris-MS-896:


southpolesteve:
Can you elaborate more on this change. What bug does this resolve and how to reproduce it?

swviswan:
Is this for the client null error? if so, can you elaborate here on why you think this is the rootcause?

---
In reply to: [768017515](https://github.com/Azure/cosmos-explorer/pull/403#issuecomment-768017515) [](ancestors = 768017515)

PR #406, opened 2021-01-27
Added localization for the Self Serve Model

srinathnarayanan:
This PR
- Adds the i18n localization packages and initialization file
- Adds a translate component as parent of the selfs serve component
- Adds a English translation json file to be used with SelfServe classes


src/Localization/en/translations.json
```diff
@@ -0,0 +1,33 @@
+{
+    "translations": {
```
swviswan, line 2: >translations [](start = 5, length = 12)

StringConstants

swviswan:
>     min: 400,

dont use hard coded value. recommend to always get from RP. 

---
Refers to: src/SelfServe/Example/SelfServeExample.tsx:238 in ac707a3. [](commit_id = ac707a387dba5cc532e947fdf0987ac4feb2a5bc, deletion_comment = False)

swviswan (approved):
:shipit:

src/i18n.js
```diff
@@ -0,0 +1,33 @@
...
+.use(XHR)
+.use(LanguageDetector)
+.use(initReactI18next)
+.init({
+    resources: {
+        en: EnglishTranslations
+    },
+    fallbackLng: "en",
+    detection: {order: ['navigator', 'cookie', 'localStorage', 'sessionStorage', 'querystring', 'htmlTag']},
+    debug: true,
```
languy, line 17: Should this be `false` in production mode? (we have code somewhere to check whether we're in webpack prod or dev mode).

src/i18n.js
```diff
@@ -0,0 +1,33 @@
...
+    keySeparator: ".",
+    interpolation: {
+        formatSeparator: ","
+    },
+    react: {
+        wait: true,
+        bindI18n: 'languageChanged loaded',
+        bindStore: 'added removed',
+        nsMode: 'default',
+        useSuspense: false
```
languy, line 29: Have you tried `true`? I'm assuming this is the graceful way to not showing anything until the translations are loaded.

src/i18n.js
```diff
@@ -0,0 +1,33 @@
...
+import XHR from "i18next-http-backend"
+import EnglishTranslations from './Localization/en/translations.json'
+
+i18n
+.use(XHR)
+.use(LanguageDetector)
+.use(initReactI18next)
+.init({
+    resources: {
+        en: EnglishTranslations
```
languy, line 13: It would be nice if we can load multiple resource files for one given language. Ideally we have one for core data explorer and one for each use case so the developers don't mess with each other (and data explorer).  Also, it would be nice to only load what we need for a given use case and not load ALL the translations.

One idea: in the function where you dynamically import the right component, import the right translation file(s) as well there.
Another idea: import the translation file inside that self-serve component, so that it gets automatically (and conditionally) imported only when you dynamically import this component.

src/SelfServe/SelfServeComponent.tsx
```diff
@@ -312,41 +322,52 @@ export class SelfServeComponent extends React.Component<SelfServeComponentProps,
...
+                        {getTranslation(this.state.notification.message)}
+                      </MessageBar>
+                    )}
+                    <SmartUiComponent
+                      disabled={this.state.refreshResult?.isUpdateInProgress}
+                      descriptor={this.state.root as SmartUiDescriptor}
+                      currentValues={this.state.currentValues}
+                      onInputChange={this.onInputChange}
+                      onError={this.onError}
+                      getTranslation={getTranslation}
```
languy, line 362: This might work for self-serve, but seems a bit too constraining for DE: does this mean for nested components we have to punch holes through all the intermediate ones? Ideally, we would want to just import the translation function from the "i18n" package anywhere in our code and use it like a global thing accessible from anywhere. Or have the top `<Translation>` node by a react Provider.

languy (approved):

src/Explorer/Controls/SmartUi/SmartUiComponent.tsx
```diff
@@ -122,10 +124,10 @@ export class SmartUiComponent extends React.Component<SmartUiComponentProps, Sma
   private renderInfo(info: Info): JSX.Element {
     return (
       <MessageBar styles={{ root: { width: 400 } }}>
-        {info.message}
```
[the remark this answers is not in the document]
srinathnarayanan: A message does appear when the key is not present, from the i18n framework. Will rename the strings to have 'TKey' suffix.

southpolesteve (approved):
I want to hold off on expanding any translation code to the rest of data explorer. We need a more thorough discussion of the libraries and approaches that exist.

src/SelfServe/SelfServeComponent.tsx
```diff
@@ -312,41 +322,52 @@ export class SelfServeComponent extends React.Component<SelfServeComponentProps,
...
-                descriptor={this.state.root as SmartUiDescriptor}
-                currentValues={this.state.currentValues}
-                onInputChange={this.onInputChange}
-                onError={this.onError}
-              />
-            </>
-          )}
-        </Stack>
-      </div>
+      <Translation>
```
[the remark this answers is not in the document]
srinathnarayanan: Will do in next PR.

src/i18n.js
srinathnarayanan: will do in next PR

src/SelfServe/SelfServeComponent.tsx
srinathnarayanan: will tackle this in separate PR.

src/i18n.js
srinathnarayanan: will tackle in next PR

PR #411, opened 2021-01-29
Add support for Schema Analyzer

gitanuj:
- [x] Add `enableSchemaAnalyzer` feature flag to show `Schema (Preview)` option only for mongo collections
- [x] Honor `hidden` flag in kernel spec metadata to hide Mongo kernel from dropdown
- [x] Add SchemaAnalyzerTab

image
image

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/411?feature.enableSchemaAnalyzer=true)


src/Contracts/ActionContracts.ts
```diff
@@ -4,6 +4,7 @@
 export enum TabKind {
   SQLDocuments,
   MongoDocuments,
+  MongoSchema,
```
swviswan, line 7: >MongoSchema [](start = 2, length = 11)

want to call it as Schema sice we may be able to extend it outside of mongoAPI as well

src/Contracts/ViewModels.ts
```diff
@@ -361,6 +362,7 @@ export enum CollectionTabKind {
   Schema = 19,
```
swviswan, line 362: >Schema [](start = 2, length = 6)

interesting.. whats this?

src/Explorer/OpenActions.ts
```diff
@@ -79,6 +79,14 @@ function openCollectionTab(
           break;
         }
 
+        if (
+          action.tabKind === ActionContracts.TabKind.MongoSchema ||
+          (<any>action).tabKind === ActionContracts.TabKind[ActionContracts.TabKind.MongoSchema]
```
notlaforge, line 84: What is going on here?

src/Explorer/Tree/Collection.ts
```diff
@@ -526,6 +527,51 @@ export default class Collection implements ViewModels.Collection {
...
+        tabPath: "",
+        collection: this,
+        node: this,
+        hashLocation: `${Constants.HashRoutePrefixes.collectionsWithIds(this.databaseId, this.id())}/mongoSchema`,
+        isActive: ko.observable(false),
+        onLoadStartKey: startKey,
+        onUpdateTabsButtons: this.container.onUpdateTabsButtons,
+      });
+      this.container.tabsManager.activateNewTab(mongoSchemaTab);
+    }
```
notlaforge, line 572: I realize this is just an established pattern, but you can actually avoid using `let` and the `as MongoSchemaTab[]` here and keep things type-safe, like so:

```suggestion
    for (const tab of this.container.tabsManager.openedTabs()) {
      if (
        tab instanceof MongoSchemaTab &&
        tab.collection?.databaseId === this.databaseId &&
        tab.collection?.id() === this.id()
      ) {
        return this.container.tabsManager.activateTab(tab);
      }
    }

    const startKey = TelemetryProcessor.traceStart(Action.Tab, {
      databaseName: this.databaseId,
      collectionName: this.id(),
      dataExplorerArea: Constants.Areas.Tab,
      tabTitle: "Schema",
    });
    this.documentIds([]);

    this.container.tabsManager.activateNewTab(
      new MongoSchemaTab({
        account: userContext.databaseAccount,
        masterKey: userContext.masterKey || "",
        container: this.container,
        tabKind: ViewModels.CollectionTabKind.MongoSchema,
        title: "Schema",
        tabPath: "",
        collection: this,
        node: this,
        hashLocation: `${Constants.HashRoutePrefixes.collectionsWithIds(this.databaseId, this.id())}/mongoSchema`,
        isActive: ko.observable(false),
        onLoadStartKey: startKey,
        onUpdateTabsButtons: this.container.onUpdateTabsButtons,
      })
    );
```

notlaforge (approved):

src/Explorer/OpenActions.ts
gitanuj: Not sure. I was just following the pattern here. I think these are contracts used to programmatically launch a tab (triggered by some window message?). Looks like `tabKind` here can be enum (int) or a string.

src/Contracts/ViewModels.ts
gitanuj: This is for the Parquet Schema tab which only works for accounts with analytical store enabled.

src/Contracts/ActionContracts.ts
gitanuj: Since there's already a `Schema` tab we need to think of a different name. I can rename it once we decide on a better name and start adapting for other API types.

gitanuj: Renaming this to `SchemaAnalyzer` instead for now.

ato9000 (approved):

PR #417, opened 2021-02-02
Move delete collection confirmation pane to react

victor-meng:
- Created `PanelContainerComponent` which contains the panel header and the actual panel content. When we open a new panel, we simply need to set the panel content and make `PanelContainerComponent` visible.
- Created `PanelManager` which contains all panel related logic. It opens the delete collection confirmation pane by using react hook to set the content of the `PanelContainerComponent`.
- Created `PanelErrorComponent` and `PanelFooterComponent` which can be re-used in other panels as well
- Created `DeleteCollectionConfirmationPaneComponent` which consists of the `PanelErrorComponent`, `PanelFooterComponent`, and the actual content.
- Added a new feature flag `enableKoPanel` which allows the users to switch back to using the old knockout panels in case of issues with the new react panels.
- Added back some styles for free tier improvement that were accidentally deleted in https://github.com/Azure/cosmos-explorer/commit/2b2de7c645c3257e01fc01f39c8a1ea689ed03c7
- Some minor refactoring


src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+import Explorer from "../Explorer";
+import LoadingIndicator_3Squares from "../../../images/LoadingIndicator_3Squares.gif";
+
+export interface DeleteCollectionConfirmationPaneProps {
+  explorer: Explorer;
+  closePanel: () => void;
+  openNotificationConsole: () => void;
+}
+
+export interface DeleteCollectionConfirmationPaneStates {
```
southpolesteve, line 26:
```suggestion
export interface DeleteCollectionConfirmationPaneState {
```

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,201 @@
+import * as NotificationConsoleUtils from "../../Utils/NotificationConsoleUtils";
```
southpolesteve, line 1: Meta comment: I think we should stop suffixing all of our component files with "Component". Most of the things we work on are components. I don't think its providing value.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+      formError: "",
+      isExecuting: false,
+    };
+  }
+
+  render(): JSX.Element {
+    return (
+      <div className="panelContentContainer">
+        <PanelErrorComponent {...this.getPanelErrorProps()} />
+        {this.renderConfirmDeleteSection()}
```
southpolesteve, line 51: I'm not a fan of this pattern of calling a function to render more JSX. Especially in the case it is not stateful. I think these should either A) be just inlined. Or B) split out into separate components entirely. 

I think by doing the `this.render*` pattern we are getting the worst of both worlds. Its not very idiomatic to react and we end up coupling our components making them more difficult to reuse later.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { deleteCollection } from "../../Common/dataAccess/deleteCollection";
+import { getErrorMessage, getErrorStack } from "../../Common/ErrorHandlingUtils";
+import { DefaultExperienceUtility } from "../../Shared/DefaultExperienceUtility";
+import { PanelErrorComponent, PanelErrorProps } from "./PanelErrorComponent";
+import DeleteFeedback from "../../Common/DeleteFeedback";
+import Explorer from "../Explorer";
+import LoadingIndicator_3Squares from "../../../images/LoadingIndicator_3Squares.gif";
+
+export interface DeleteCollectionConfirmationPaneProps {
+  explorer: Explorer;
```
southpolesteve, line 21: This is not ideal. Would like to avoid coupling more stuff to the entire Explorer class. But I don't see any way around right now. Maybe once we extract resource tree....

src/Explorer/Panes/PanelManager.tsx
```diff
@@ -0,0 +1,49 @@
...
+  panelContent: JSX.Element;
+  onPanelClose: () => void;
+}
+
+export interface PanelManagerParams {
+  setPanelParams: (panelParams: PanelParams) => void;
+  setIsNotificationConsoleExpanded: (isExpanded: boolean) => void;
+}
+
+export class PanelManager {
```
southpolesteve, line 17: Lets sync on this offline. I want to talk about how this can be done with hooks.

southpolesteve (approved):

src/Explorer/Panes/PanelFooterComponent.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { PrimaryButton } from "office-ui-fabric-react";
+
+export interface PanelFooterProps {
+  buttonLabel: string;
+  onOKButtonClicked: () => void;
+}
+
+export class PanelFooterComponent extends React.Component<PanelFooterProps> {
```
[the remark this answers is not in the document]
victor-meng: Done

src/Explorer/Panes/PanelErrorComponent.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Icon, Text } from "office-ui-fabric-react";
+
+export interface PanelErrorProps {
+  message: string;
+  isWarning: boolean;
+  showErrorDetails: boolean;
+  openNotificationConsole?: () => void;
+}
+
+export class PanelErrorComponent extends React.Component<PanelErrorProps> {
```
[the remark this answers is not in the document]
victor-meng: Done

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
victor-meng: Done

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
victor-meng: I agree. I just thought that this would make it a bit easier to read. I've moved them in-line in my latest commit.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
victor-meng: Yeah I actually tried to remove this dependency but it turned out to be a lot of work so I'm going remove it in a separate PR.

src/Explorer/Panes/PanelManager.tsx
victor-meng: Discussed with Steve offline. I will remove PanelManager and pass panel state hooks to Explorer directly.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
victor-meng: Agreed. I'll do PR to rename all the react components.

src/Explorer/Panes/PanelContainerComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface PanelContainerProps {
+  headerText: string;
+  panelContent: JSX.Element;
+  isConsoleExpanded: boolean;
+  isOpen: boolean;
+  closePanel: () => void;
+}
+
+export class PanelContainerComponent extends React.Component<PanelContainerProps> {
+  private readonly consoleHeaderHeight = 32;
```
languy, line 13:
```suggestion
  private static readonly consoleHeaderHeight = 32;
```

src/hooks/useSidePanel.ts
```diff
@@ -0,0 +1,29 @@
...
+  panelContent: JSX.Element;
+  headerText: string;
+  openSidePanel: (headerText: string, panelContent: JSX.Element) => void;
+  closeSidePanel: () => void;
+}
+
+export const useSidePanel = (): SidePanelHooks => {
+  const [isPanelOpen, setIsPanelOpen] = useState<boolean>(false);
+  const [panelContent, setPanelContent] = useState<JSX.Element>();
+  const [headerText, setHeaderText] = useState<string>();
```
languy, line 14: Very nice. Makes it very clean.

languy (approved):
Great work. What about `GenericPaneComponent`? Is it possible to replace it or retrofit with the latest?

victor-meng:
> Great work. What about `GenericPaneComponent`? Is it possible to replace it or retrofit with the latest?

Yep I will replace it with the new pattern.

src/Explorer/Explorer.tsx
```diff
@@ -111,6 +112,8 @@ export interface ExplorerParams {
   setIsNotificationConsoleExpanded: (isExpanded: boolean) => void;
   setNotificationConsoleData: (consoleData: ConsoleData) => void;
   setInProgressConsoleDataIdToBeDeleted: (id: string) => void;
+  openSidePanel: (headerText: string, panelContent: JSX.Element) => void;
```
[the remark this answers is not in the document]
victor-meng: This should work for the left panel as well. The fluent UI panel component has a property that allows the panel to be opened on either side of the screen. I think we can re-use `PanelContainerComponent` for left panels too assuming we cannot have both a left and a right panel open at the same time.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,188 @@
...
+  closePanel: () => void;
```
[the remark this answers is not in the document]
victor-meng: I'll use Panel for everything.

src/Explorer/Panes/DeleteCollectionConfirmationPaneComponent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    };
+  }
+
+  render(): JSX.Element {
+    return (
+      <div className="panelContentContainer">
+        <PanelErrorComponent {...this.getPanelErrorProps()} />
+        <div className="panelMainContent">
+          <div className="confirmDeleteInput">
+            <div>
```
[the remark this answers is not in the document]
victor-meng: Nop, I think I was trying to add a style class here but realized it wasn't needed. Thanks for catching this!

src/Explorer/Panes/PanelContainerComponent.tsx
victor-meng: Done

PR #419, opened 2021-02-03
Changes to notebook publish pane

srinathnarayanan:
- Made custom image as default
- Reset image to empty when option is changed
- Make cover image mandatory
- Make sure Name, descriptiona and cover image check shows error message on notification console


gitanuj (approved):

src/Explorer/Panes/__snapshots__/PublishNotebookPaneComponent.test.tsx.snap
```diff
@@ -52,28 +52,29 @@ exports[`PublishNotebookPaneComponent renders 1`] = `
...
         label="Cover image"
         onChange={[Function]}
         options={
           Array [
-            Object {
-              "key": "URL",
-              "text": "URL",
-            },
             Object {
               "key": "Custom Image",
```
swviswan, line 61: >key [](start = 15, length = 3)

nt: why key should have space in its name?

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
```diff
@@ -146,11 +146,15 @@ export class PublishNotebookPaneAdapter implements ReactAdapter {
 
     let startKey: number;
 
-    try {
-      if (!this.name || !this.description || !this.author) {
-        throw new Error("Name, description, and author are required");
-      }
+    if (!this.name || !this.description || !this.author || !this.imageSrc) {
+      const formError = `Failed to publish ${this.name} to gallery`;
```
swviswan, line 150: >name [](start = 50, length = 4)

will name have .ipynb extension

gitanuj, line 150:
```suggestion
      const formError = `Failed to publish ${FileSystemUtil.stripExtension(this.name, "ipynb")} to gallery`;
```

gitanuj: Btw why can't we just throw like before?

srinathnarayanan: the clearPublishingMessage is used to remove the "Publishing to gallery ..." message i suppose. An unintended side effect is that it wipesout the Notification console message for the error that is thrown from the validation. So had to separate it out.

srinathnarayanan: yes

src/Explorer/Panes/__snapshots__/PublishNotebookPaneComponent.test.tsx.snap
srinathnarayanan: it can have a space, we are using the same enum for both key and display text.

src/Explorer/Panes/PublishNotebookPaneAdapter.tsx
srinathnarayanan: yes

src/Explorer/Panes/__snapshots__/PublishNotebookPaneComponent.test.tsx.snap
srinathnarayanan: it can have a space, we are using the same enum for both key and display text.

swviswan (approved):

PR #423, opened 2021-02-04
Disable dark overlay for Dialog

gitanuj:
Earlier we used to have a light overlay but now it is a dark overlay. May be it regressed somehow. This change makes Dialog use light overlay again.

Current look
image

After the fix
image


swviswan (approved):
:shipit:

srinathnarayanan (approved):

PR #437, opened 2021-02-09
Remove window.authType

southpolesteve:
Mostly a pure refactor to get out of relying on `window.authType`. We shouldn't be storing this as a global as its making it difficult to move logic over to React. PR also makes much of the data explorer inputs optional so we can simplify configuration.


southpolesteve:
Moving this to draft. Need to think through some of the issues here for refreshing tokens.

victor-meng (approved):

PR #440, opened 2021-02-17
Remove SplashScreenComponentAdapter

notlaforge:
Really, merge `SplashScreenComponentAdapter` and `SplashScreenComponent` into one `SplashScreen` React component.


.eslintignore
```diff
@@ -377,8 +377,7 @@ src/Explorer/Notebook/temp/inputs/editor.tsx
 src/Explorer/Notebook/temp/markdown-cell.tsx
 src/Explorer/Notebook/temp/source.tsx
 src/Explorer/Notebook/temp/syntax-highlighter/index.tsx
-src/Explorer/SplashScreen/SplashScreenComponent.tsx
-src/Explorer/SplashScreen/SplashScreenComponentApdapter.tsx
+src/Explorer/SplashScreen/SplashScreen.tsx
```
southpolesteve, line 380: Do you think its possible to remove this from eslintignore in this step or should it be out of scope for these migration PRs?

notlaforge: Both! In the case of this file it was just a bunch of "don't pass null" warnings, so super easy. But it also muddies the waters, I think?

southpolesteve: Fair. I am fine with doing it later.

southpolesteve (approved):

src/Explorer/SplashScreen/SplashScreen.tsx
```diff
@@ -1,63 +1,161 @@
...
+    this.container.tabsManager.openedTabs.subscribe(() => this.setState({}));
+    this.container.selectedNode.subscribe(() => this.setState({}));
+    this.container.isNotebookEnabled.subscribe(() => this.setState({}));
   }
 
-  public forceRender = (): void => {
-    window.requestAnimationFrame(() => this.parameters(Date.now()));
-  };
+  public shouldComponentUpdate() {
+    return this.container.tabsManager.openedTabs.length === 0;
```
languy, line 52: Note that there's some talk to move splash screen to a tab. But it's easy to remove later.

src/Explorer/SplashScreen/SplashScreen.tsx
```diff
@@ -209,20 +307,27 @@ export class SplashScreenComponentAdapter implements ReactAdapter {
...
         iconSrc: null,
         title: "Configure automatic failover",
         description: "Learn more about Cosmos DB high-availability",
-        onClick: () => window.open(SplashScreenComponentAdapter.failoverUrl),
+        onClick: () => window.open(SplashScreen.failoverUrl),
       },
     ];
   }
+
+  private onSplashScreenItemKeyPress(event: React.KeyboardEvent, callback: () => void) {
```
languy, line 327: Note that we now have an `<AccessibleElement as="li" />` which does this for you.

languy (approved):

src/Explorer/SplashScreen/SplashScreen.tsx
notlaforge: Fantastic, good to know. I've seen this pattern in a few different places, so this seems like a good PR to do in one fell swoop?

PR #442, opened 2021-02-18
Fixing issue with autoscale A/B test

deborahc:
Currently, the first time a user opens the New Container pane, the autoscale setting is not being read from the flight info. This is because the flight into is being set in resetData(), which is not called the 1st time a user opens the pane. 

This change adds logic to set the flight info on the open call.


southpolesteve (approved):

src/Explorer/Panes/AddDatabasePane.ts
```diff
@@ -271,6 +271,7 @@ export default class AddDatabasePane extends ContextualPaneBase {
   public open() {
     super.open();
     this.resetData();
+    this.isAutoPilotSelected(this.container.isAutoscaleDefaultEnabled());
```
swviswan, line 274: >isAutoPilotSelected [](start = 9, length = 19)

this is not necessary right - since reset data already does this?

swviswan (approved):
:shipit:

src/Explorer/Panes/AddDatabasePane.ts
deborahc: Yes, I removed the redundant check.

PR #445, opened 2021-02-23
Remove AdHoc Access and Token Renewal Pane

southpolesteve:
Cleans up a bunch of obsolete or no longer functioning UI. Some of this is functionality we likely want in the future, but its low priority right now and will be easier to add back post-react migration


src/Explorer/Explorer.tsx
```diff
@@ -1728,60 +1581,6 @@ export default class Explorer {
     this._addSynapseLinkDialogProps.valueHasMutated();
   };
 
-  public handleMessage(message: any) {
```
victor-meng, line 1731: Do we no longer need this? I remember this function handles opening the gettings started notebook when user clicks on the "setup container" button in the quickstart blade. Is that being handled somewhere else now or is the worflow changed/removed?

southpolesteve: Good question. I will try it out. It may be broken.

southpolesteve: I was able to fix open actions

victor-meng (approved):

PR #446, opened 2021-02-24
Remove Dialog Adapter

southpolesteve:
Refactor similar to #440 that removes the unnecessary adapter layer


src/Explorer/Controls/Dialog.tsx
```diff
@@ -50,7 +50,7 @@ const DIALOG_TITLE_FONT_SIZE = "17px";
 const DIALOG_TITLE_FONT_WEIGHT = 400;
 const DIALOG_SUBTEXT_FONT_SIZE = "15px";
 
-export class DialogComponent extends React.Component<DialogProps, {}> {
+export class Dialog extends React.Component<DialogProps> {
```
victor-meng, line 53: We should make this a functional component later.

victor-meng (approved):

PR #449, opened 2021-02-24
Move "Open Full Screen" to React Panel

southpolesteve:
Migrates the "Open Full Screen" experience to a self contained React component and panel. This was a place we were still using jQueryUI!

image


victor-meng (approved):

swviswan:
> 

can all tests be under a test folder?

---
Refers to: src/Explorer/OpenFullScreen.test.tsx:18 in 480f6bb. [](commit_id = 480f6bb41942be8aa9e55daec4b521adfe0a676b, deletion_comment = False)

southpolesteve:
Were using the Jest default which loads all files that match `*.test.*`. Part of the reason we stuck with that is because before we did have a top level test folder and it made migrating to Jest really easy so Jest didn't try and run tests written in Karma. I kind of like it because it makes it easy to find the test files. They are right next to the code being tested.

We do have a top level test folder but it only has end to end tests. Separation makes it easy to run just unit tests or just E2E test which is what happens in CI now.

I don't really feel that strongly about this setup. Its just the convention now, but open to other options if they would improve our eng systems and productivity.

PR #457, opened 2021-02-27
Removed published, favourites tabs for hosted gallery

srinathnarayanan:
This PR removes published and favourites tabs when users sees the gallery through the cosmos.azure.com/gallery.html landing page and is not logged in. In such a case, the container passed to GalleryViewerComponent is undefind, and it used as a check.

Portal experience:
portal - all tabs

Stand alone Gallery experience:
public gallery-no tabs

### DIFF
--- a/src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
+++ b/src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
@@ -33,7 +33,6 @@ import { InfoComponent } from "./InfoComponent/InfoComponent";
 import { handleError } from "../../../Common/ErrorHandlingUtils";
 import { trace } from "../../../Shared/Telemetry/TelemetryProcessor";
 import { Action, ActionModifiers } from "../../../Shared/Telemetry/TelemetryConstants";
-import { configContext, Platform } from "../../../ConfigContext";
 
 export interface GalleryViewerComponentProps {
   container?: Explorer;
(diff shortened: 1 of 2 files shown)


src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -161,8 +162,10 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
       )
     );
 
-    tabs.push(this.createFavoritesTab(GalleryTab.Favorites, this.state.favoriteNotebooks));
-    tabs.push(this.createPublishedNotebooksTab(GalleryTab.Published, this.state.publishedNotebooks));
+    if (configContext.platform !== Platform.Hosted) {
+      tabs.push(this.createFavoritesTab(GalleryTab.Favorites, this.state.favoriteNotebooks));
```
swviswan, line 166: I am afraid this is incorrect.. We could have standalone DE but logged in user in which case we should show these tabs. 

If you click a notebook from gallery, you will see "Download to my notebooks" only if it is not stand alone gallery viewer. Please refer that code and use a similar check here.

srinathnarayanan: Modified it to use same logic as the download notebooks button

swviswan (approved):
:shipit:

ramrajprabu (approved):

PR #461, opened 2021-03-02
Notebook viewer bug fix

srinathnarayanan:
When the notebook viewer gets a tab name in the url, it does not propagate it to the Gallery link which is opened when the back arrow is clicked. This PR fixes that error. This is needed now since public gallery is going to be available as well.

notebook viewer fix


swviswan (approved):
:shipit:

southpolesteve (approved):

PR #463, opened 2021-03-03
MostRecentActivity changes

notlaforge:
This changes the public API a bit, so that recording activity (the most common use) is less involved.

### DIFF
--- a/src/Explorer/MostRecentActivity/MostRecentActivity.ts
+++ b/src/Explorer/MostRecentActivity/MostRecentActivity.ts
@@ -8,21 +8,18 @@ export enum Type {
 }
 
 export interface OpenNotebookItem {
+  type: Type.OpenNotebook;
   name: string;
   path: string;
 }
 
 export interface OpenCollectionItem {
+  type: Type.OpenCollection;
   databaseId: string;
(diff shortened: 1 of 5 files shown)


src/Explorer/MostRecentActivity/MostRecentActivity.ts
```diff
@@ -32,7 +31,7 @@ interface StoredData {
  * Stores most recent activity
  */
 class MostRecentActivity {
-  private static readonly schemaVersion: string = "1";
+  private static readonly schemaVersion: string = "2";
```
languy, line 34: You might want to do something about the old items saved under version 1. We don't need to maintain support, but we should probably ignore them (so they don't break the code) and maybe even delete them from local storage.

languy (approved):

src/Explorer/MostRecentActivity/MostRecentActivity.ts
notlaforge: I think someone (you?) already handled that. When it gets loaded, if the schema doesn't match, the recent activity is discarded.

southpolesteve (approved):

PR #465, opened 2021-03-03
Fix subscription leak

notlaforge:
> [2:52 AM] @languy
>     
> @notlaforge I'm looking at SplashScreen as a template to migrate ResourceTree. I noticed that it subscribes to some ko's in the constructor: https://github.com/Azure/cosmos-explorer/blob/7188e8d8c26b9fde55e8e138875db8205f81a536/src/Explorer/SplashScreen/SplashScreen.tsx#L48 I suggest we do the subscribing in componentDidMount() and unsubscribe in componentWillUnmount()in order to properly cleanup and only subscribe when the component is being used. Thoughts? 
>

### DIFF
--- a/src/Explorer/SplashScreen/SplashScreen.tsx
+++ b/src/Explorer/SplashScreen/SplashScreen.tsx
@@ -41,17 +41,27 @@ export class SplashScreen extends React.Component<SplashScreenProps> {
   private static readonly failoverUrl = "https://docs.microsoft.com/azure/cosmos-db/high-availability";
 
   private readonly container: Explorer;
+  private subscriptions: Array<{ dispose: () => void }>;
 
   constructor(props: SplashScreenProps) {
     super(props);
     this.container = props.explorer;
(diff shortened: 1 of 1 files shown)


src/Explorer/SplashScreen/SplashScreen.tsx
```diff
@@ -41,19 +41,29 @@ export class SplashScreen extends React.Component<SplashScreenProps> {
...
-    this.container.selectedNode.subscribe(() => this.setState({}));
-    this.container.isNotebookEnabled.subscribe(() => this.setState({}));
   }
 
   public shouldComponentUpdate() {
     return this.container.tabsManager.openedTabs.length === 0;
   }
 
+  public componentWillUnmount() {
+    while (this.subscriptions.length) this.subscriptions.pop().dispose();
```
languy, line 56: nit:
```suggestion
    while (this.subscriptions.length) {
      this.subscriptions.pop().dispose();
    }
```

languy (approved):

southpolesteve (approved):

PR #467, opened 2021-03-04
Remove injected cell before download

srinathnarayanan:
Link Injection adds a cell with a link to the notebook viewer, for a published notebook.
This cell is stripped out before displaying it in hte notebook viewer.
But when the notebook is downloaded, this cell was not being removed before the download.
This PR fixed that bug.

### DIFF
--- a/src/Utils/GalleryUtils.ts
+++ b/src/Utils/GalleryUtils.ts
@@ -13,6 +13,8 @@ import { getErrorMessage, getErrorStack, handleError } from "../Common/ErrorHand
 import { HttpStatusCodes } from "../Common/Constants";
 import { trace, traceFailure, traceStart, traceSuccess } from "../Shared/Telemetry/TelemetryProcessor";
 import { Action, ActionModifiers } from "../Shared/Telemetry/TelemetryConstants";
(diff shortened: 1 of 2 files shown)


swviswan (approved):
:shipit:

src/Utils/GalleryUtils.ts
```diff
@@ -281,6 +286,17 @@ export function downloadItem(
   );
 }
 
+export const removeNotebookViewerLink = (notebook: Notebook, newCellId: string): void => {
+  if (!newCellId) {
+    return;
+  }
+  const notebookV4 = notebook as NotebookV4;
+  if (notebookV4 && notebookV4.cells[0].source[0].search(newCellId)) {
```
languy, line 294: Consider using optional chaining in case `cells` or `source` length is < 1.
```suggestion
  if (notebookV4?.cells[0]?.source[0]?.search(newCellId)) {
```
This works, but consider using nteract's selectors to access and update the inside of the notebook model so that we don't have worry if the format changes, etc. (I know you can find a cell, and probably delete one at least).

languy (approved):

PR #473, opened 2021-03-04
Allow publishing same nuget package versions

southpolesteve:


southpolesteve:
Already covered in another PR

PR #476, opened 2021-03-05
Drop IE11 Support

southpolesteve:
Do not open box until March 31st 2021


notlaforge:
Man that’s a lot of code removed

webpack.config.js
```diff
@@ -56,56 +56,40 @@ const htmlRule = {
...
-  ],
-  include: /node_modules/,
-  // Exclude large modules we know don't need transpiling
-  exclude: /vega|monaco|plotly/,
-};
-
 module.exports = function (env = {}, argv = {}) {
   const mode = argv.mode || "development";
-  const rules = [fontRule, lessRule, imagesRule, cssRule, htmlRule, typescriptRule];
+  const target = mode === "development" ? "es2018" : "es2017";
```
notlaforge, line 61: Curious, why?

southpolesteve: Good Q. I am mostly going off this thing that Google put out that says ES2017 is the safe bet for "modern" https://web.dev/publish-modern-javascript/. But ES2018 does have some nice things like `for await` and in dev mode its nice to use see the actual code rather than transpiled code. So... does it matter? Prob not. ES2017 everywhere would be fine.

southpolesteve:
For future code spelunkers: I tried to use ESbuild here because it is way fast, but unfortunately its decorators are not compatible with the SelfServe project so I had to abandon it.

package.json
```diff
@@ -44,9 +44,7 @@
     "@types/node-fetch": "2.5.7",
     "@uifabric/react-cards": "0.109.110",
     "@uifabric/styling": "7.13.7",
-    "abort-controller": "3.0.0",
```
notlaforge, line 47: Was this just not used anywhere?

notlaforge (approved):

package.json
notlaforge: nvm I see now it provided a polyfill

PR #485, opened 2021-03-05
Remove openMongo test

southpolesteve:
This is already covered by the index policy test. No need to test it twice.


southpolesteve:
Already done in another PR

PR #487, opened 2021-03-08
Fix lint query utils

hardiknai-techm:
Fix Lint errors in QueryUtils.ts


notlaforge (approved):
Sorry to butt in, I was just having a hard time reading the diff due to the re-arranging of the functions.

PR #535, opened 2021-03-10
Deprecate Explorer Properties

southpolesteve:
One of the major challenges in migrating to React has been how coupled components are to `Explorer.ts` and the class `Explorer`. It results in us passing Explorer around a lot in places we don't need to. It also makes it difficult to write truly isolated components that can be rendered and tested without Explorer. A first step toward fixing this problem is deprecating `Explorer` properties that:

1. Never change in the lifecycle of the application
2. Are available on `userContext` or `configContext`
3. Can be determined before we even instantiate `Explorer` in the outer React layer

This PR 
- Marks some of the properties meeting this criteria as deprecated so it is obvious to avoid them
- Shows a simple example of removing these properties in `DatabaseSettingsTab.ts`
- Adds a new property `apiType` to `userContext` that serves as a single source of truth for the API type of an account


notlaforge (approved):
Thank you!!

notlaforge:
What’s with all the import statement shuffle?

southpolesteve:
@notlaforge New contractors asked we add VSCode Organize Imports as an on save action which is causing it to happen on any file I touch. Not sure how I feel about it yet. It does remove unused imports which I love. But it does cause some unnecessary PR changes. If you have opinions do let me know.

PR #537, opened 2021-03-11
Organize all imports at once

notlaforge:
I "wrote" this by running a global find-and-replace on vscode to remove the `;` at the end of each `import` statement, which caused vscode to save all files, reinserting the `;` and thus running the organize-imports-on-save thing.

Why? It makes pull requests less filled with these reordered lines.


notlaforge:
https://github.com/microsoft/TypeScript/pull/42630

notlaforge:
In hindsight, I'm going to abandon this.

PR #540, opened 2021-03-11
Dedicated Gateway Portal Changes

fnbalaji:
The following are the changes to enable portal experience for the dedicated gateway. The current set of changes involves changes to the text being presented to the users. 

Before Enabling:
image

While Enabling:
image

After Enabling:
image

Before Disabling:
image

While Disabling:
image

After Disabling:
image


src/Localization/en/translations.json
```diff
@@ -37,6 +37,46 @@
...
+            "Disable": "Disable",
+            "LearnAboutDedicatedGateway": "read more about dedicated gateway.",
+            "SKUs": "SKUs",
+            "SKUsPlaceHolder": "Select SKUs",
+            "NumberOfInstances": "Number of instances",
+            "CosmosD4s": "Cosmos.D4s (General Purpose Cosmos Compute with 4 vCPUs, 16 GB Memory)",
+            "CosmosD8s": "Cosmos.D8s (General Purpose Cosmos Compute with 8 vCPUs, 32 GB Memory)",
+            "CosmosD16s": "Cosmos.D16s (General Purpose Cosmos Compute with 16 vCPUs, 64 GB Memory)",
+            "CosmosD32s": "Cosmos.D32s (General Purpose Cosmos Compute with 32 vCPUs, 128 GB Memory)",
+            "CreateMessage": "DedicatedGateway resource is being created.",
```
srinathnarayanan, line 52: Need to finalize these strings

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -9,91 +10,281 @@ import {
...
+  @OnChange(onSKUChange)
   @Values({
     labelTKey: "SKUs",
     choices: getSkus,
-    placeholderTKey: "Select SKUs",
+    placeholderTKey: "SKUsPlaceHolder",
   })
   sku: ChoiceItem;
 
   @Values({
```
srinathnarayanan, line 269: We need to add another dynamic description that has a reference to the keys blade

src/Localization/en/translations.json
fnbalaji: Let's keep these strings, these are correct for now!

src/Localization/en/translations.json
```diff
@@ -37,6 +37,56 @@
             "OnSaveFailureMessage": "Data save operation not currently permitted."
         },
         "SqlX": {
+            "DedicatedGatewayDescription": "Provision a dedicated gateway cluster for your Azure Cosmos DB account. A dedicated gateway is compute that is a front-end to data in your Azure Cosmos DB account. Your dedicated gateway automatically includes the integrated cache, which can improve read performance. ",
+            "DedicatedGateway": "Dedicated Gateway",
+            "Enable": "Enable",
+            "Disable": "Disable",
+            "LearnAboutDedicatedGateway": "Learn more about dedicated gateway.",
+            "DeprovisioningDetailsText": "Learn more about deprovisioning the dedicated gateway.",
+            "DedicatedGatewayPricing": "Learn more about dedicated gateway pricing",
```
srinathnarayanan, line 46: this shows up in both the warning banner and in the cost description. Seems like redundant info. DO we need it?

src/Localization/en/translations.json
```diff
@@ -37,6 +37,56 @@
...
+            "SKUs": "SKUs",
+            "SKUsPlaceHolder": "Select SKUs",
+            "NumberOfInstances": "Number of instances",
+            "CosmosD4s": "Cosmos.D4s (General Purpose Cosmos Compute with 4 vCPUs, 16 GB Memory)",
+            "CosmosD8s": "Cosmos.D8s (General Purpose Cosmos Compute with 8 vCPUs, 32 GB Memory)",
+            "CosmosD16s": "Cosmos.D16s (General Purpose Cosmos Compute with 16 vCPUs, 64 GB Memory)",
+            "CosmosD32s": "Cosmos.D32s (General Purpose Cosmos Compute with 32 vCPUs, 128 GB Memory)",
+            "CreateMessage": "DedicatedGateway resource is being created.",
```
srinathnarayanan, line 54: should it be 2 separate words? Here and everywhere else in translations.json

src/Localization/en/translations.json
```diff
@@ -37,6 +37,56 @@
...
+            "CosmosD8Details": "General Purpose Cosmos Compute with 8 vCPUs, 32 GB Memory",
+            "CosmosD16Details": "General Purpose Cosmos Compute with 16 vCPUs, 64 GB Memory",
+            "CosmosD32Details": "General Purpose Cosmos Compute with 32 vCPUs, 128 GB Memory",
+            "Cost": "Cost",
+            "CostText": "Hourly cost of the dedicated gateway resource depends on the SKU selection and number of instances used.",
+            "ConnectionString": "Connection String",
+            "ConnectionStringText": "To use the dedicated gateway, use the connection string shown in ",
+            "KeysBlade": "the keys blade",
+            "WarningBannerOnUpdate": "Adding or modifying dedicated gateway nodes may affect your bill.",
+            "WarningBannerOnDelete": "After deprovisioning the dedicated gateway, you must update any applications using the old dedicated gateway connection string."
```
srinathnarayanan, line 89: this is kind of confusiong. Not sure "old dedicated gateway connection string" comes through in its meaning. Can we instead say what should be used after deprovisining?

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -9,91 +10,325 @@ import {
...
+  link: {
+    href: generateBladeLink(BladeType.SqlKeys),
+    textTKey: "KeysBlade",
+  },
+};
+
+const getSKUDetails = (sku: string): string => {
+  if (sku === "Cosmos.D4s") {
+    return "CosmosD4Details";
+  } else if (sku === "Cosmos.D8s") {
```
srinathnarayanan, line 42: Declare "Cosmos.D8s" and rest as constants here and reuse?

southpolesteve (approved):

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -9,91 +10,325 @@ import {
...
+  public onSave = async (
+    currentValues: Map<string, SmartUiInput>,
+    baselineValues: Map<string, SmartUiInput>
+  ): Promise<OnSaveResult> => {
+    const dedicatedGatewayCurrentlyEnabled = currentValues.get("enableDedicatedGateway")?.value as boolean;
+    const dedicatedGatewayOriginallyEnabled = baselineValues.get("enableDedicatedGateway")?.value as boolean;
+
+    currentValues.set("warningBanner", undefined);
+
+    //TODO : Add try catch for each RP call and return relevant notifications
```
srinathnarayanan, line 179: Do you want to catch any errors and give a well formed error message via a new error? right now we just show whatever is returned as part of the thrown error.
try {
deleteDedicatedGateway(...)
} catch (e) {
  if (e.message === "...") {
    throw new Error ("well formed error message")
   }
  throw e
}

src/Localization/en/translations.json
fnbalaji: I think it is okay. Eventually, in the mid-term, we will replace cost description with the actual estimate. Also, this was approved by Tim and we should be okay with it. Changing this is fairly straight-forward

src/Localization/en/translations.json
fnbalaji: Yes, this makes sense. Changed all occurences of DedicatedGateway to Dedicated gateway.

src/SelfServe/SqlX/SqlX.tsx
fnbalaji: This is okay, because the RP message is what is propagated in the other services, like "exclusive account lock" etc... 

Unless there is a need, I don't think we need to pre-emptively add some message.

srinathnarayanan (approved):

PR #542, opened 2021-03-12
Migration/delete database confirmation in react

sunilyadav840:
Created DeleteDatabaseConfirmationPanel.tsx (React component for Delete Database confirmation panel)  which consists of the PanelErrorComponent, PanelFooterComponent, and the actual content.
Created DeleteDatabaseConfirmationPanel.test.tsx for test cases of DeleteDatabaseConfirmationPanel.tsx.
Added flag enableKoPanel which allows the users to switch back to using the old knockout panels.
Knockout components still exist.

image


src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setFormError("");
+    setHiddenFalse();
+
+    const startKey: number = TelemetryProcessor.traceStart(Action.DeleteDatabase, {
+      databaseId: selectedDatabase.id(),
+      dataExplorerArea: Constants.Areas.ContextualPane,
+      paneTitle: "Delete Database",
+    });
+
+    return Q(
```
southpolesteve, line 75: Remove the Q() call here. We do not want to use Q promises in new places

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    return {
+      isWarning: true,
+      showErrorDetails: false,
+      message:
+        "Warning! The action you are about to take cannot be undone. Continuing will permanently delete this resource and all of its children resources.",
+    };
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const submit = (): Q.Promise<any> => {
```
southpolesteve, line 55: Remove usage of Q

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    // Get selected db
+    const selectedDatabase = props.explorer.findSelectedDatabase();
```
southpolesteve, line 57: Can you pass in the name of the database as a prop instead when the user clicks to open? We are trying to remove coupling to explorer

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (!isValid()) {
+      setFormError("Input database name does not match the selected database");
+      NotificationConsoleUtils.logConsoleMessage(
+        ConsoleDataType.Error,
+        `Error while deleting collection ${selectedDatabase && selectedDatabase.id()}`
+      );
+      return Q.resolve();
```
southpolesteve, line 64: Remove usage of Q()

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
southpolesteve, line 55: After removing Q, turn this into an async function

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setHiddenFalse();
+
+    const startKey: number = TelemetryProcessor.traceStart(Action.DeleteDatabase, {
+      databaseId: selectedDatabase.id(),
+      dataExplorerArea: Constants.Areas.ContextualPane,
+      paneTitle: "Delete Database",
+    });
+
+    return Q(
+      deleteDatabase(selectedDatabase.id()).then(
```
southpolesteve, line 76: Use async.await here

src/Explorer/Explorer.tsx
```diff
@@ -2548,4 +2554,18 @@ export default class Explorer {
           />
         );
   }
+
+  public openDeleteDatabaseConfirmationPane(): void {
+    // enableKoPanel which allows the users to switch back to using the old knockout panels.
+    this.isFeatureEnabled(Constants.Features.enableKOPanel)
```
southpolesteve, line 2560: Remove this and delete any code from the old panel. I think we need to be aggressive in our migration strategy. We can always revert the commit if there is a major problem

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { shallow } from "enzyme";
+import * as ko from "knockout";
+import React from "react";
+import { deleteDatabase } from "../../Common/dataAccess/deleteDatabase";
+import * as ViewModels from "../../Contracts/ViewModels";
+import Explorer from "../Explorer";
+import { DeleteDatabaseConfirmationPanel } from "./DeleteDatabaseConfirmationPanel";
+
+describe("Delete Database Confirmation Pane", () => {
+  describe("Explorer.isLastDatabase() and Explorer.isLastNonEmptyDatabase()", () => {
```
southpolesteve, line 12: Move this block of tests to a new `Explorer.test.ts` file. This is not testing anything related to the panel

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
+import { useBoolean } from "@uifabric/react-hooks";
+import { Text, TextField } from "office-ui-fabric-react";
+import Q from "q";
```
southpolesteve, line 3: Remove Q entirely

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+  explorer: Explorer;
+  closePanel: () => void;
+  openNotificationConsole: () => void;
+}
+
+export const DeleteDatabaseConfirmationPanel: FunctionComponent<DeleteDatabaseConfirmationPanelProps> = (
+  props: DeleteDatabaseConfirmationPanelProps
+): JSX.Element => {
+  // For showing/hiding panel
+  const [isHidden, { setTrue: setHiddenTrue, setFalse: setHiddenFalse }] = useBoolean(true);
```
southpolesteve, line 30: Rename this to `isLoading` since that seems to be what it is doing down below

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
southpolesteve, line 55: Remove any. It looks like this should return Promise<void>

southpolesteve (changes requested):

src/Explorer/Explorer.tsx
sunilyadav840: Okay Sure. Removed

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Done

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: We can pass props from explorer.ts like below in the code.

`public openDeleteDatabaseConfirmationPane(): void {
    this.openSidePanel(
      "Delete Database",
      <DeleteDatabaseConfirmationPanel
        explorer={this}
        openNotificationConsole={this.expandConsole}
        closePanel={this.closeSidePanel}
        selectedDatabase={this.findSelectedDatabase()}
      />
    );
  }`
  
  But we already passing this as explorer props in the component.
  So Is it make sense to pass separate props for database.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Removed Q

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Removed

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Removed

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Resolved Promise using async await

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Used async await

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
sunilyadav840: Moved these test cases in new file `Explorer.test.tsx`

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Removed

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Removed

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+      );
+    }
+  };
+
+  const shouldRecordFeedback = (): boolean => {
+    const { explorer } = props;
+    return explorer.isLastNonEmptyDatabase() || (explorer.isLastDatabase() && explorer.isSelectedDatabaseShared());
+  };
+
+  const isValid = (): boolean => {
```
victor-meng, line 137: We should be able to do this check inline:

`if (!selectedDatabase?.id() || databaseInput !== selectedDatabase.id()) {
	...
}`

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      (deleteDatabase as jest.Mock).mockResolvedValue(undefined);
+      (TelemetryProcessor.trace as jest.Mock).mockReturnValue(undefined);
+    });
+
+    beforeEach(() => {
+      const props = {
+        explorer: fakeExplorer,
+        closePanel: (): void => undefined,
+        openNotificationConsole: (): void => undefined,
+        selectedDatabase: fakeExplorer.findSelectedDatabase(),
```
victor-meng, line 74: You need to override the `findSelectedDatabase` function to return a mock database. Right now `fakeExplorer.findSelectedDatabase()` would always return null.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      };
+      wrapper = mount(<DeleteDatabaseConfirmationPanel {...props} />);
+      props.explorer.isLastNonEmptyDatabase = () => true;
+      wrapper.setProps(props);
+    });
+
+    it("Should call delete database", () => {
```
victor-meng, line 81: I don't think this test case actually tests anything. You are not checking if `deleteDatabase` is called here as the title suggested. You don't have any `expect` statements after you click on the OK button.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      expect(wrapper.exists("#sidePanelOkButton")).toBe(true);
+      wrapper.find("#sidePanelOkButton").hostNodes().simulate("click");
+      wrapper
+        .find("#confirmDatabaseId")
+        .hostNodes()
+        .simulate("change", { target: { value: selectedDatabaseId } });
+      wrapper.unmount();
+    });
+
+    it("should record feedback", async () => {
```
victor-meng, line 93: Similar to my comments above, you are not actually testing anything here. You need to override the `findSelectedDatabase` function to make it return a mock database. Otherwise the `isValid` check will always fail and you will never be able reach the code beyond that. Please see the test is set up in `DeleteCollectionConfirmationPane.test.ts`.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+import Explorer from "../Explorer";
+import { ConsoleDataType } from "../Menus/NotificationConsole/NotificationConsoleComponent";
+import { PanelErrorComponent } from "./PanelErrorComponent";
+import { PanelFooterComponent } from "./PanelFooterComponent";
+
+interface DeleteDatabaseConfirmationPanelProps {
+  explorer: Explorer;
+  closePanel: () => void;
+  openNotificationConsole: () => void;
+  selectedDatabase: {
```
victor-meng, line 23: The type here should be `ViewModels.Database`.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
+import { useBoolean } from "@uifabric/react-hooks";
+import { Text, TextField } from "office-ui-fabric-react";
+import React, { FunctionComponent, useState } from "react";
+import LoadingIndicator_3Squares from "../../../images/LoadingIndicator_3Squares.gif";
+import * as Constants from "../../Common/Constants";
```
victor-meng, line 5: The only thing we are using in this file is `Constants.Area`. Instead of importing the entire file we should just import area: `import { Area } from "../../Common/Constants";`

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { deleteDatabase } from "../../Common/dataAccess/deleteDatabase";
+import DeleteFeedback from "../../Common/DeleteFeedback";
+import { getErrorMessage, getErrorStack } from "../../Common/ErrorHandlingUtils";
+import * as ViewModels from "../../Contracts/ViewModels";
```
victor-meng, line 9: Similar to my comments above, we only use `ViewModels.Collection` and `ViewModels.Database` here. We should only import the ones we need.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+      message:
+        "Warning! The action you are about to take cannot be undone. Continuing will permanently delete this resource and all of its children resources.",
+    };
+  };
+
+  const submit = async (): Promise<void> => {
+    const { selectedDatabase, explorer } = props;
+    if (!isValid()) {
+      setFormError("Input database name does not match the selected database");
+      NotificationConsoleUtils.logConsoleMessage(
```
victor-meng, line 60: The `logConsoleMessage` function is deprecated. Use `logConsoleError` instead.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,187 @@
...
+import LoadingIndicator_3Squares from "../../../images/LoadingIndicator_3Squares.gif";
+import * as Constants from "../../Common/Constants";
+import { deleteDatabase } from "../../Common/dataAccess/deleteDatabase";
+import DeleteFeedback from "../../Common/DeleteFeedback";
+import { getErrorMessage, getErrorStack } from "../../Common/ErrorHandlingUtils";
+import * as ViewModels from "../../Contracts/ViewModels";
+import { DefaultExperienceUtility } from "../../Shared/DefaultExperienceUtility";
+import { Action, ActionModifiers } from "../../Shared/Telemetry/TelemetryConstants";
+import * as TelemetryProcessor from "../../Shared/Telemetry/TelemetryProcessor";
+import * as NotificationConsoleUtils from "../../Utils/NotificationConsoleUtils";
```
victor-meng, line 13: `import { logConsoleError }` from "../../Utils/NotificationConsoleUtils";

victor-meng (changes requested):

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Updated code.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
sunilyadav840: Update changes

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
sunilyadav840: update changes.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.test.tsx
sunilyadav840: Updated changes, Thanks for reference.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Updated changes

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Updated changes

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Updated

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Updated

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Done

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
victor-meng: The condition that I wrote is for checking whether the input is invalid, which is what we are doing in the `submit` function (see line 55 of DeleteDatabaseConfirmationPanel.tsx).

What I meant was that we don't really need this function. We can just do this check inline (in line 55).

southpolesteve (approved):

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
sunilyadav840: Changes updated. Thank you.

src/Explorer/Panes/DeleteDatabaseConfirmationPanel.tsx
```diff
@@ -0,0 +1,169 @@
...
+      isWarning: true,
+      showErrorDetails: false,
+      message:
+        "Warning! The action you are about to take cannot be undone. Continuing will permanently delete this resource and all of its children resources.",
+    };
+  };
+
+  const submit = async (): Promise<void> => {
+    const { selectedDatabase, explorer } = props;
+    if (selectedDatabase?.id() && databaseInput !== selectedDatabase.id()) {
```
victor-meng, line 55: I think we can just do `if (databaseInput !== selectedDatabase?.id())` here.

victor-meng (approved):

PR #544, opened 2021-03-12
Move ResourceTree to React

languy:
* Move `ResourceTreeAdapter` code to `useNotebooks` hook.
* Remove unnecessary `triggerRender()` calls


src/Main.tsx
```diff
@@ -158,7 +186,15 @@ const App: React.FunctionComponent = () => {
...
-                  <div style={{ overflowY: "auto" }} data-bind="if: !isAuthWithResourceToken(), react:resourceTree" />
+                  <div style={{ overflowY: "auto" }} data-bind="if: !isAuthWithResourceToken()">
+                    <ResourceTree
+                      explorer={explorer}
+                      lastRefreshedTime={lastRefreshTime}
+                      galleryContentRoot={galleryContentRoot}
+                      myNotebooksContentRoot={myNotebooksContentRoot}
+                      gitHubNotebooksContentRoot={gitHubNotebooksContentRoot}
+                    />
+                  </div>
```
notlaforge, line 197:
```suggestion
                  <div style={{ overflowY: "auto" }}>
                    isAuthWithResourceToken() || <ResourceTree
                      explorer={explorer}
                      lastRefreshedTime={lastRefreshTime}
                      galleryContentRoot={galleryContentRoot}
                      myNotebooksContentRoot={myNotebooksContentRoot}
                      gitHubNotebooksContentRoot={gitHubNotebooksContentRoot}
                    />
                  </div>
```

notlaforge: Is this an opportunity to remove one more knockout binding?

src/Main.tsx
```diff
@@ -99,10 +113,24 @@ const App: React.FunctionComponent = () => {
...
     closeDialog,
+    onRefreshNotebookList: refreshList,
+    initializeGitHubRepos,
+    getMyNotebooksContentRoot,
   };
   const config = useConfig();
   const explorer = useKnockoutExplorer(config?.platform, explorerParams);
 
+  // TODO fix this
+  context.container = explorer;
```
southpolesteve, line 124: I have some ideas about how to avoid this

src/Main.tsx
languy: It's a bit more involved as `isAuthWithResourceToken` isn't available in `Main.tsx`, but in `Explorer` (btw, you probably meant to enclose this expression inside `{{ }}`). So we'll first move everything to react, then will come back to remove the ko's (ie replace them as some state in `Main.tsx`).

PR #557, opened 2021-03-18
Remove Explorer.isPreferredAPIMongo

southpolesteve:


src/UserContext.ts
```diff
@@ -30,8 +30,10 @@ export type PortalEnv = "localhost" | "blackforest" | "fairfax" | "mooncake" | "
 const userContext: UserContext = { isTryCosmosDBSubscription: false, portalEnv: "prod" };
 
 function updateUserContext(newContext: UserContext): void {
+  if (newContext.databaseAccount) {
+    newContext.apiType = apiType(userContext.databaseAccount);
```
notlaforge, line 34: Does this mutate what was passed? Does it matter?

notlaforge (approved):

src/UserContext.ts
hardiknai-techm: userContext => newContext need to replace i have correct in this PR

southpolesteve:
@hardiknai-techm can you take another look at this?

PR #564, opened 2021-03-19
Reset cell status after execution is canceled

gitanuj:
Currently if there's an error when executing a code cell the UI keeps on showing circular progress bar. This is because the transient cell status is never reset after an execution has been canceled. Nteract simply stops updating the transient cell status as soon as an error is encountered and issues EXECUTE_CANCELED action for all cells.

This change adds an epic to always reset cell status after EXECUTE_CANCELED action.

### DIFF
--- a/src/Explorer/Notebook/NotebookComponent/epics.ts
+++ b/src/Explorer/Notebook/NotebookComponent/epics.ts
@@ -1,4 +1,4 @@
-import { EMPTY, merge, of, timer, concat, Subject, Subscriber, Observable, Observer } from "rxjs";
+import { EMPTY, merge, of, timer, concat, Subject, Subscriber, Observable, Observer, from } from "rxjs";
 import { webSocket } from "rxjs/webSocket";
 import { StateObservable } from "redux-observable";
 import { ofType } from "redux-observable";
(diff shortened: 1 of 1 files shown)


src/Explorer/Notebook/NotebookComponent/epics.ts
```diff
@@ -944,6 +944,39 @@ const traceNotebookKernelEpic = (
   );
 };
 
+const resetCellStatusOnExecuteCanceledEpic = (
+  action$: Observable<actions.ExecuteCanceled>,
+  state$: StateObservable<AppState>
+): Observable<actions.UpdateCellStatus> => {
+  return action$.pipe(
+    ofType(actions.EXECUTE_CANCELED),
```
swviswan, line 952: >EXECUTE_CANCELED [](start = 19, length = 16)

your description says EXECUTE_FAILED? hopr this is the right one?

gitanuj: Oops! Updated the description. The code is correct.

swviswan (approved):
:shipit:

languy (approved):
This works, so let's get it in.
It seems that nteract doesn't have this problem: if I execute 'aaa', the indicator goes to [*] then back to [1].
Because we are overriding some of the nteract's epics, we have to manually enumerate all epics (nteract's and ours) in an array. We could be either overriding one that was doing this or we are missing the nteract epic that does this. I suggest we double-check that we are including all nteract's epics that we are not overriding.

gitanuj:
> This works, so let's get it in.
> It seems that nteract doesn't have this problem: if I execute 'aaa', the indicator goes to [*] then back to [1].
> Because we are overriding some of the nteract's epics, we have to manually enumerate all epics (nteract's and ours) in an array. We could be either overriding one that was doing this or we are missing the nteract epic that does this. I suggest we double-check that we are including all nteract's epics that we are not overriding.

I did verify cell status updates by setting breakpoint in reducers\src\core\entities\contents\notebook.ts:updateCellStatus and the status seems to be going from undefined -> "queued" -> "busy" only for Nteract also. I'm also talking to Safia to see if this is a bug in Nteract too. Will also verify that we are including all epics.

PR #569, opened 2021-03-19
Make ready message backwards compatible

ato9000:
This change makes the ready message compatible with Ibiza SDK IFrame v1 control, currently used by the portal. The previous change made the ready message compatible with IFrame v2, but breaks current portal users.

### DIFF
--- a/src/Common/MessageHandler.ts
+++ b/src/Common/MessageHandler.ts
@@ -69,12 +69,12 @@ export function sendReadyMessage(): void {
       {
         signature: "pcIframe",
         kind: "ready",
+        data: "ready",
       },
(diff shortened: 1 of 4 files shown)


languy (approved):

src/Common/MessageHandler.ts
```diff
@@ -69,13 +69,13 @@ export function sendReadyMessage(): void {
       {
         signature: "pcIframe",
         kind: "ready",
+        data: "ready",
```
swviswan, line 72: I believe, Srinath should be doing the same for the self serve iframe as well?

swviswan (approved):
:shipit:

PR #577, opened 2021-03-23
Use window instead of window.parent in extractFeatures function

victor-meng:
Since we are now calling `extractFeatures` in `UserContext` which is in the outer iframe layer instead of `explorer` which is in the inner iframe, we should use `window.location.search` instead of `window.parent.location.search`.


notlaforge (approved):

notlaforge:
How did my original PR pass the end-to-end tests, I wonder.

victor-meng:
> How did my original PR pass the end-to-end tests, I wonder.

My guess is that the frames in the e2e tests are different from production.

gitanuj (approved):

PR #581, opened 2021-03-25
Migrate Notebook Upload File to React

hardiknai-techm:
https://user-images.githubusercontent.com/80053762/112623268-27bb9780-8e52-11eb-9e22-1926fd78fc79.mp4


package.json
```diff
@@ -164,6 +164,7 @@
     "html-loader": "0.5.5",
     "html-loader-jest": "0.2.1",
     "html-webpack-plugin": "3.2.0",
+    "husky": "5.2.0",
```
notlaforge, line 167: Can we talk about this in a different PR, with the whole team? I'm personally not a fan.

hardiknai-techm: ok i have remove code. we will discuss in the call

PR #582, opened 2021-03-25
Fix broken MongoDocumentsTab

notlaforge:
I assumed during #534 that `MongoDocumentsTab.html` was used for the corresponding component, but I was wrong. I now think that file is not used at all.


victor-meng (approved):

gitanuj (approved):

swviswan (approved):
:shipit:

PR #585, opened 2021-03-26
Hotfix - Fix table API query projections

ato9000:
When building queries with projections, the resulting query does not include the "id" property as part of the projection. The "id" property is used by the results grid to display as the RowKey so the result is queries wih projections do not show the RowKey.

This change fixes this by including "id" as part of the projections.


victor-meng (approved):

ato9000:
I will merge to master first, then cherry pick to hotfix branch

PR #617, opened 2021-04-05
SelfServe - Telemetry and Localization improvements

srinathnarayanan:
this PR
- Makes SelfServeTelemetryProcessor dependent on TelemetryProcessor
- renamed the methods in selfServeTelemetryProcessor
- extends trace, traceStart etc to take a message type as well ("TelemetryInfo" string enum is used for SelfServe as opposed to TelemetryInfo = 0 numeric enum for DE telemetry)
- Adds webpack chunk name for localization files dynamically loaded
- Added/Modified SqlX specific translation keys

### DIFF
--- a/src/SelfServe/SelfServe.tsx
+++ b/src/SelfServe/SelfServe.tsx
@@ -23,9 +23,11 @@ const loadTranslationFile = async (className: string): Promise<void> => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let translations: any;
   try {
-    translations = await import(`../Localization/${language}/${fileName}`);
+    translations = await import(
+      /* webpackChunkName: "Localization-[request]" */ `../Localization/${language}/${fileName}`
+    );
   } catch (e) {
-    translations = await import(`../Localization/en/${fileName}`);
+    translations = await import(/* webpackChunkName: "Localization-en-[request]" */ `../Localization/en/${fileName}`);
   }
   i18n.addResourceBundle(language, className, translations.default, true);
 };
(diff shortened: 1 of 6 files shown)


src/SelfServe/SelfServe.tsx
```diff
@@ -23,9 +23,11 @@ const loadTranslationFile = async (className: string): Promise<void> => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let translations: any;
   try {
-    translations = await import(`../Localization/${language}/${fileName}`);
+    translations = await import(
+      /* webpackChunkName: "Localization-[request]" */ `../Localization/${language}/${fileName}`
+    );
   } catch (e) {
-    translations = await import(`../Localization/en/${fileName}`);
+    translations = await import(/* webpackChunkName: "Localization-en-[request]" */ `../Localization/en/${fileName}`);
```
southpolesteve, line 30: Not a blocker but why is it necessary to name these? Can't webpack handle it without a specific name?

src/Shared/Telemetry/TelemetryProcessor.ts
```diff
@@ -1,15 +1,23 @@
 ﻿import { sendMessage } from "../../Common/MessageHandler";
 import { configContext } from "../../ConfigContext";
 import { MessageTypes } from "../../Contracts/ExplorerContracts";
+import { SelfServeMessageTypes } from "../../Contracts/SelfServeContracts";
 import { userContext } from "../../UserContext";
 import { appInsights } from "../appInsights";
 import { Action, ActionModifiers } from "./TelemetryConstants";
 
-type TelemetryData = { [key: string]: unknown };
+type TelemetryType = MessageTypes.TelemetryInfo | SelfServeMessageTypes.TelemetryInfo;
```
southpolesteve, line 9: Put a note here about why we have two different types. I think future people will appreciate knowing

src/SelfServe/SelfServe.tsx
srinathnarayanan: it can. but this is easier to debug, esp in this case when we want to make sure the correct files got loaded.

src/SelfServe/SelfServeTelemetryProcessor.ts
```diff
@@ -1,69 +1,24 @@
-import { sendMessage } from "../Common/MessageHandler";
```
southpolesteve, line 1: Alternative suggestion. You could just inspect the URL to see if it contains `selfServe` inside the existing telemetry processor and then send the correct type that way. Then you would not need this file and you can use the same exact API we have now.

southpolesteve (approved):

src/SelfServe/SelfServeTelemetryProcessor.ts
srinathnarayanan: discussed offline, going with this approach for now since we need to enforce the SelfServeTelemetryData to have a mandatory "SelfServeClassName" field.

ato9000 (approved):

notlaforge (approved):

PR #621, opened 2021-04-06
Move tabs state out into React

notlaforge:
This PR is just about moving the tabs array. I'm hoping to let it bake for a bit before merging the rest of the tabs in react work.

Preview here: https://ms.portal.azure.com/?dataExplorerSource=https%3A%2F%2Fcosmos-explorer-preview.azurewebsites.net%2Fcommit%2Fda809beb82bb54dc82da18eda41caaf7b9b6597f%2Fexplorer.html#@microsoft.onmicrosoft.com/resource/subscriptions/b9c77f10-b438-4c32-9819-eef8a654e478/resourceGroups/stfaul/providers/Microsoft.DocumentDb/databaseAccounts/stfaul-sql/dataExplorer

### DIFF
--- a/src/Explorer/SplashScreen/SplashScreen.tsx
+++ b/src/Explorer/SplashScreen/SplashScreen.tsx
@@ -50,10 +50,6 @@ export class SplashScreen extends React.Component<SplashScreenProps> {
     this.subscriptions = [];
   }
 
-  public shouldComponentUpdate() {
-    return this.container.tabsManager.openedTabs.length === 0;
-  }
-
   public componentWillUnmount() {
     while (this.subscriptions.length) {
       this.subscriptions.pop().dispose();
@@ -62,7 +58,6 @@ export class SplashScreen extends React.Component<SplashScreenProps> {
 
   public componentDidMount() {
     this.subscriptions.push(
-      this.container.tabsManager.openedTabs.subscribe(() => this.setState({})),
       this.container.selectedNode.subscribe(() => this.setState({})),
       this.container.isNotebookEnabled.subscribe(() => this.setState({}))
     );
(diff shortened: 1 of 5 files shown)


src/Explorer/SplashScreen/SplashScreen.tsx
```diff
@@ -80,7 +75,13 @@ export class SplashScreen extends React.Component<SplashScreenProps> {
     const tipsItems = this.createTipsItems();
     const onClearRecent = this.clearMostRecent;
 
-    return (
+    const formContainer = (jsx: JSX.Element) => (
```
victor-meng, line 78: Do we always want to show this? In the original code, it looks like we only show this when there's no tab open.

src/hooks/useTabs.ts
```diff
@@ -0,0 +1,15 @@
...
+import { TabsManager } from "../Explorer/Tabs/TabsManager";
+import { useObservableState } from "./useObservableState";
+
+export type UseTabs = {
+  tabs: readonly TabsBase[];
+  tabsManager: TabsManager;
+};
+
+export function useTabs(): UseTabs {
+  const tabsManager = new TabsManager();
```
southpolesteve, line 11: Does this create a new tabs manager instance on every render and if so is that an issue?

`useState(() => newTabsManager())` would ensure you get only one instance

notlaforge: Yes... but so far only `<App>` uses this. I'm guessing that if `<App>`  is being re-rendered it's because we're restarting the whole thing?

gitanuj (approved):

southpolesteve (approved):

PR #624, opened 2021-04-07
Sandbox outputs for every code cell using <iframe>

gitanuj:
To provide some context - currently we don't sandbox notebook outputs. This can be dangerous because malicious scripts can do unexpected things when just viewing a notebook. We're trying to solve that by sandboxing all outputs in `<iframe>`.

- [x] Add feature flag `sandboxNotebookOutputs` to enable this new behavior
- [x] Add `IFrameOutputs` which sandboxes all outputs for a code cell using `<iframe>`
- [x] The sandbox restricts everything except `allow-downloads allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts allow-popups-to-escape-sandbox`. More info [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe).

Verified that the following type of outputs work:
- [x] Bokeh
- [x] `%%upload` progress bar
- [x] Nteract data explorer
- [x] matplotlib graphs
- [x] error text
- [x] HTML
- [x] Javascript

Preview URL for standalone gallery: https://cosmos-explorer-preview.azurewebsites.net/commit/0f4c261edbc746686ba26f1c1f593557c538c1c6/gallery.html?feature.sandboxNotebookOutputs=true

Unfortunately we can't preview in Data Explorer due to feature flags no being honored in Portal. And AAD doesn't work for Hosted mode for preview endpoints.

### DIFF
--- a/src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
+++ b/src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
@@ -1,0 +1,70 @@
+import { AppState, ContentRef, selectors } from "@nteract/core";
+import { Output } from "@nteract/outputs";
+import Immutable from "immutable";
+import React from "react";
+import { connect } from "react-redux";
+import { SandboxFrame } from "./SandboxFrame";
+
+// Adapted from https://github.com/nteract/nteract/blob/main/packages/stateful-components/src/outputs/index.tsx
+// to add support for sandboxing using <iframe>
+
+interface ComponentProps {
+  id: string;
+  contentRef: ContentRef;
+  children: React.ReactNode;
+}
+
+interface StateProps {
+  hidden: boolean;
+  expanded: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  outputs: Immutable.List<any>;
+}
+
+export class IFrameOutputs extends React.PureComponent<ComponentProps & StateProps> {
+  render(): JSX.Element {
+    const { outputs, children, hidden, expanded } = this.props;
+    return (
+      <SandboxFrame
+        style={{ border: "none", width: "100%" }}
+        sandbox="allow-downloads allow-forms allow-pointer-lock allow-popups allow-same-origin allow-scripts allow-popups-to-escape-sandbox"
+      >
+        <div className={`nteract-cell-outputs ${hidden ? "hidden" : ""} ${expanded ? "expanded" : ""}`}>
+          {outputs.map((output, index) => (
+            <Output output={output} key={index}>
+              {children}
+            </Output>
+          ))}
+        </div>
+      </SandboxFrame>
+    );
+  }
+}
+
+export const makeMapStateToProps = (
+  initialState: AppState,
+  ownProps: ComponentProps
+): ((state: AppState) => StateProps) => {
+  const mapStateToProps = (state: AppState): StateProps => {
+    let outputs = Immutable.List();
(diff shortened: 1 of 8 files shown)


src/Utils/StyleUtils.ts
```diff
@@ -0,0 +1,23 @@
+// Adapted from https://gist.github.com/davidgilbertson/ed3c8bb8569bc64b094b87aa88bed5fa
+export function copyStyles(sourceDoc: Document, targetDoc: Document): void {
```
captainsafia, line 2: Interesting -- what's the motivation for injecting the parent page's styles to the iframe?

gitanuj: I noticed that there were some output transforms (like Nteract Data Explorer, StreamText etc) which depend on styles which were present in the document. May be the right thing to do here is to make sure the output transforms inject their own styles when rendered?

src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
```diff
@@ -0,0 +1,103 @@
...
+    copyStyles(document, this.frame.contentWindow.document);
+
+    const { hidden, expanded } = this.props;
+    this.outputsContainer = document.createElement("div");
+    this.outputsContainer.className = `nteract-cell-outputs ${hidden ? "hidden" : ""} ${expanded ? "expanded" : ""}`;
+
+    this.resizeObserver = new ResizeObserver(
+      _.debounce(() => (this.frame.height = this.outputsContainer.scrollHeight + "px"), 200)
+    );
+    this.resizeObserver.observe(this.outputsContainer);
```
southpolesteve, line 56: What removes this observer? Will this continue to observe after the component unmounts?

src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
```diff
@@ -0,0 +1,103 @@
...
+  id: string;
+  contentRef: ContentRef;
+  children: React.ReactNode;
+}
+
+interface StateProps {
+  hidden: boolean;
+  expanded: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  outputs: Immutable.List<any>;
```
southpolesteve, line 24: does `unknown` work here instead of `any`?

src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
gitanuj: Added code to disconnect this observer when component unmounts.

src/Explorer/Notebook/NotebookRenderer/outputs/IFrameOutputs.tsx
gitanuj: Unfortunately `unknown` doesn't work here. This component uses `<Output />` which relies on `outputs` with this same exact signature.

srinathnarayanan (approved):

vignesh-mr (approved):

src/Utils/StyleUtils.ts
gitanuj: I did some more investigation and looks like Nteract Data Explorer is indeed injecting its own CSS styles. And still we don't see those styles in the `<iframe>` by default. This leads me to believe that React portals are injecting the styles in parent document.

southpolesteve (approved):

PR #627, opened 2021-04-07
Added className to SelfServeBaseClass

srinathnarayanan:
This PR
- There was mismatch in prod when the classname was set before the object instantiation earlier. now we use constructor.name throughout
- Removed setting className for selfServeDescriptor in SelfServeUtil
- Modified tests

### DIFF
--- a/src/SelfServe/SelfServe.tsx
+++ b/src/SelfServe/SelfServe.tsx
@@ -38,20 +38,26 @@ const loadTranslations = async (className: string): Promise<void> => {
 };
 
 const getDescriptor = async (selfServeType: SelfServeType): Promise<SelfServeDescriptor> => {
+  let descriptor: SelfServeDescriptor;
   switch (selfServeType) {
     case SelfServeType.example: {
       const SelfServeExample = await import(/* webpackChunkName: "SelfServeExample" */ "./Example/SelfServeExample");
-      await loadTranslations("SelfServeExample");
-      return new SelfServeExample.default().toSelfServeDescriptor();
+      const selfServeExample = new SelfServeExample.default();
+      await loadTranslations(selfServeExample.getClassName());
+      descriptor = selfServeExample.toSelfServeDescriptor();
+      break;
     }
     case SelfServeType.sqlx: {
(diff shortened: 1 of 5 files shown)


src/SelfServe/SelfServe.tsx
```diff
@@ -38,20 +38,26 @@ const loadTranslations = async (className: string): Promise<void> => {
 };
 
 const getDescriptor = async (selfServeType: SelfServeType): Promise<SelfServeDescriptor> => {
+  let descriptor: SelfServeDescriptor;
```
southpolesteve, line 41: You don't need this variable. You can always return directly from inside the switch statement:

`return selfServeExample.toSelfServeDescriptor();`

src/SelfServe/SelfServeTypes.ts
```diff
@@ -68,27 +68,37 @@ export abstract class SelfServeBaseClass {
     baselineValues: ReadonlyMap<string, SmartUiInput>
   ) => Promise<OnSaveResult>;
   public abstract onRefresh: () => Promise<RefreshResult>;
+  private className: string;
+
+  constructor() {
+    this.className = this.constructor.name;
+  }
+
+  public getClassName(): string {
```
southpolesteve, line 77: Getter methods around private state isn't really idiomatic in TS. I would just use `this.constructor.name` everywhere

src/SelfServe/SelfServeUtils.test.tsx
```diff
@@ -150,7 +152,7 @@ describe("SelfServeUtils", () => {
     ]);
     const expectedDescriptor = {
       root: {
-        id: "TestClass",
+        id: undefined as string,
```
southpolesteve, line 155: Why this cast? Did you mean to use `""`?

southpolesteve (approved):

src/SelfServe/SelfServeTypes.ts
srinathnarayanan: Will remove that.

gitanuj (approved):

PR #630, opened 2021-04-08
add collection panel improvements

victor-meng:
1. Adjusted the margin between each field from 4px to 8px.
2. Bolded the title of the throughput input component so it has the same style as the other field titles.
3. When the advanced section is expanded, automatically scroll down the panel content so that everything inside the advanced section is visible.
4. Reduced the top and bottom margin of the panel footer button from 20px to 16px.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/630)


southpolesteve (approved):

src/Utils/APITypeUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+      break;
+    case "Cassandra":
+    case "Tables":
+      collectionName = "Table";
+      break;
+    case "Gremlin":
+      collectionName = "Graph";
+      break;
+    default:
+      throw new Error(`Unknown API type: ${userContext.apiType}`);
```
notlaforge, line 25: If you handle the default like this:

```suggestion
    default: {
      const assertUnreachable = (type: never): never => {
        throw new Error(`Unknown API type: ${type}`);
      };
      assertUnreachable(userContext.apiType);
    }
```

then if a new possible value is added to `userContext.apiType`, then this code doesn't type-check anymore, which is great! See https://stackoverflow.com/a/39419171 for more.

notlaforge (approved):

src/Utils/APITypeUtils.ts
southpolesteve: I wish there was a way to enforce we do this on all switch statements via linting

src/Utils/APITypeUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (options?.isLowerCase) {
+    collectionName = collectionName.toLocaleLowerCase();
```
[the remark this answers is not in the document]
victor-meng: Agreed I'll remove this option.

src/Utils/APITypeUtils.ts
victor-meng: That's very cool pattern! I'll start using this in the future.

PR #633, opened 2021-04-09
Fix telemetry from child windows of Data Explorer

gitanuj:
With the new iFrame component in Ibiza SDK the portal window only accepts messages coming from the windows that it created. This broke our Kusto telemetry logging from the Terminal windows because terminal windows are created by Data Explorer.

To fix this we now send messages from Terminal window to Portal window via Data Explorer.

Verified that this fix works by inspecting /extensiontelemetry payload.

### DIFF
--- a/src/Common/MessageHandler.ts
+++ b/src/Common/MessageHandler.ts
@@ -48,32 +48,18 @@ export function sendCachedDataMessage<TResponseDataModel>(
 }
 
 export function sendMessage(data: any): void {
-  if (canSendMessage()) {
-    // We try to find data explorer window first, then fallback to current window
-    const portalChildWindow = getDataExplorerWindow(window) || window;
-    portalChildWindow.parent.postMessage(
-      {
-        signature: "pcIframe",
-        data: data,
-      },
-      portalChildWindow.document.referrer || "*"
-    );
-  }
+  sendMessageImpl({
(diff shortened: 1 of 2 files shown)


ato9000 (approved):

src/hooks/useKnockoutExplorer.ts
```diff
@@ -275,6 +277,10 @@ async function configurePortal(explorerParams: ExplorerParams): Promise<Explorer
   });
 }
 
+function shouldForwardMessage(message: PortalMessage) {
```
vignesh-mr, line 280: Let's confirm the parent frame isn't sending anything down with this signature else we'll end up in an infinite loop. :)

vignesh-mr (approved):

src/Common/MessageHandler.ts
```diff
@@ -89,3 +75,17 @@ export function runGarbageCollector() {
     }
   });
 }
+
+const sendMessageImpl = (message: any): void => {
```
southpolesteve, line 79:
```suggestion
const _sendMessage = (message: any): void => {
```

src/hooks/useKnockoutExplorer.ts
gitanuj: Good callout. I'll add protection around that anyway.

PR #636, opened 2021-04-09
Call readCollections for Mongo using ARM

southpolesteve:
ICM: https://portal.microsofticm.com/imp/v3/incidents/details/235371314/home

`readCollections` for Mongo accounts is still using the SQL SDK.  Calling via the SQL SDK was preventing private endpoint Mongo accounts for loading properly in the portal and those users are unable to see their collections. This original reason we special cased Mongo was due to some ARM gaps, but it appears those have been addressed.

This also fixes a bug in the development mode iframe logic for the portal. It was not updating userContext correctly when loading the message from the cache.

### DIFF
--- a/src/hooks/useKnockoutExplorer.ts
+++ b/src/hooks/useKnockoutExplorer.ts
@@ -198,19 +198,20 @@ async function configurePortal(explorerParams: ExplorerParams): Promise<Explorer
   return new Promise((resolve) => {
     // In development mode, try to load the iframe message from session storage.
     // This allows webpack hot reload to function properly in the portal
-    if (process.env.NODE_ENV === "development" && !window.location.search.includes("disablePortalInitCache")) {
-      const initMessage = sessionStorage.getItem("portalDataExplorerInitMessage");
-      if (initMessage) {
-        const message = JSON.parse(initMessage);
-        console.warn(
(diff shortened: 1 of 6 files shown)


src/hooks/useKnockoutExplorer.ts
```diff
@@ -237,29 +238,7 @@ async function configurePortal(explorerParams: ExplorerParams): Promise<Explorer
             inputs.extensionEndpoint = configContext.PROXY_PATH;
```
victor-meng, line 238: I think this if statement is already included in `updateContextsFromPortalMessage`. We should be able to delete it.

victor-meng (approved):

notlaforge (approved):
I'm sure you know what you're doing, but if I have to be 100% honest it's a bit scary to see so much `foo as Bar` around.

southpolesteve:
@notlaforge That is fair feedback. The context here is we have historically abused the `Collection` and `Database` type that extend from `TreeNode` to stand in as data models. This is a bad idea and probably not something we should keep doing but its a big lift to fix it. We should instead be using proper unions of types pulled from ARM and the SDK. The tricky bit is they are not always the same. There are small discrepancies between what ARM and SDK supports.

I have tested manually the ARM vs non-ARM cases here.

PR #643, opened 2021-04-12
[Hosted] AAD implementation for item operations

zfoster:
Adds AAD token fetching and use for item CRUD operations. This token is fetched by hosted explorer and passed into the explorer iframe. It is not used by any ARM operations and only used if present so shouldn't affect the portal data explorer.

Next step: Pass token from portal for DE use

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/643?feature.someFeatureFlagYouMightNeed=true)


src/Common/dataAccess/createDocument.ts
```diff
@@ -1,8 +1,8 @@
 import { CollectionBase } from "../../Contracts/ViewModels";
+import { logConsoleInfo, logConsoleProgress } from "../../Utils/NotificationConsoleUtils";
```
zfoster, line 2: i think my auto formatter from this package's modules was changing the order of these. Is this supposed to happen?

southpolesteve: yes

src/HostedExplorerChildFrame.ts
```diff
@@ -6,13 +6,17 @@ export interface HostedExplorerChildFrame extends Window {
   hostedConfig: HostedConfig;
 }
 
-export interface AAD {
+interface AADTokenAuth {
+  aadToken: string;
+}
```
southpolesteve, line 11: Move this down to the `AAD` interface below. None of the other auth types will have an AAD token.

src/hooks/useAADAuth.ts
```diff
@@ -81,9 +83,16 @@ export function useAADAuth(): ReturnType {
           authority: `https://login.microsoftonline.com/${tenantId}`,
           scopes: ["https://management.azure.com//.default"],
         }),
-      ]).then(([graphTokenResponse, armTokenResponse]) => {
+        msal.acquireTokenSilent({
+          // There is a bug in MSALv1 that requires us to refresh the token. Their internal cache is not respecting authority
+          forceRefresh: true,
+          // authority: `https://login.microsoftonline.com/${tenantId}`,
+          scopes: ["https://stfaul-sql.documents.azure.com/.default"],
```
zfoster, line 90: for testing

src/Platform/Hosted/extractFeatures.ts
```diff
@@ -55,6 +56,8 @@ export function extractFeatures(given = new URLSearchParams(window.location.sear
     enableSpark: "true" === get("enablespark"),
     enableTtl: "true" === get("enablettl"),
     executeSproc: "true" === get("dataexplorerexecutesproc"),
+    fetchesAADToken: true,
+    // fetchesAADToken: "true" === get("fetchesaadtoken"),
```
zfoster, line 60: for testing

southpolesteve: Uncomment this. We can't deploy this to MPAC

src/hooks/useAADAuth.ts
southpolesteve: This has to be done different. We shouldn't deploy this with a hard coded value. We should move this to another place where it can be loaded after the database account is loaded

southpolesteve (changes requested):

src/Contracts/ExplorerContracts.ts
```diff
@@ -30,6 +30,7 @@ export enum MessageTypes {
   ExplorerClickEvent,
   LoadingStatus,
   GetArcadiaToken,
+  GetAadToken,
```
southpolesteve, line 33: Doesn't the order matter here? This gets transpile to a number. I think it has to go at the end

southpolesteve (changes requested):

src/Contracts/ExplorerContracts.ts
zfoster: the types don't fail but I can remove this either way. I don't need it yet

southpolesteve (approved):

PR #656, opened 2021-04-13
Remove Explorer.isPreferredApiTable

hardiknai-techm:


src/Explorer/Menus/CommandBar/CommandBarComponentAdapter.tsx
```diff
@@ -29,7 +30,7 @@ export class CommandBarComponentAdapter implements ReactAdapter {
 
     // These are the parameters watched by the react binding that will trigger a renderComponent() if one of the ko mutates
     const toWatch = [
-      container.isPreferredApiTable,
+      userContext.apiType === "Tables",
```
notlaforge, line 33: Unfortunately this doesn't work in this context. I'm not sure how the type system didn't flag it, but `toWatch` is an `Observable[]`, and this is just a plain `boolean`.

src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
```diff
@@ -137,7 +137,13 @@ describe("ContainerSampleGenerator", () => {
...
     const experience = "not supported api";
     const explorerStub = createExplorerStub(undefined);
-    explorerStub.isPreferredApiTable = ko.computed<boolean>(() => true);
+    updateUserContext({
+      databaseAccount: {
+        properties: {
+          capabilities: [{ name: "EnableTable" }],
+        },
+      } as DatabaseAccount,
+    });
```
notlaforge, line 146: @southpolesteve, this bugs me a bit: because of the way `userContext` is, and because of the way it is updated, this test can't just do `updateUserContext({ apiType: 'flef' })`, so it ends up having to inadvertently exercise and test  `updateUserContext()` itself.

I don't know what the solution is though. It wouldn't bother me much if it was just this test, but because it's `userContext`, it has the potential to spread out fast.

southpolesteve: Discussed with Jordi offline. This bugs me too. But I think we need to tackle this problem with `userContext` in a separate step. Getting out from under Explorer is more important IMO and this PR helps.

southpolesteve:
@hardiknai-techm The unit tests are still failing. Can you take another look?

src/Explorer/Menus/CommandBar/CommandBarComponentAdapter.tsx
hardiknai-techm: ok i can remove

hardiknai-techm:
@southpolesteve i have check yesterday all test case are pass

southpolesteve (approved):

PR #657, opened 2021-04-13
Fix for Mongo shard key loaded via ARM

southpolesteve:
ICM: https://portal.microsofticm.com/imp/v3/incidents/details/236412873/home
This fixes a bug introduced by https://github.com/Azure/cosmos-explorer/pull/636

Mongo collections were previously loaded via the SQL SDK which returned a `parititonKey` property. ARM returns a `shardKey` property in a different format. This PR still loads the collections via ARM but it adds a `partitionKey` property to the response calculated from `shardKey`. This is a bit of a hack, but lots of downstream code expects `Collection.partitionKey` to exist.

I have manually tests CRUD for the following cases:
- No shard key
- Simple shard key: `foo`
- Complex shard key: `foo.bar.baz`

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/657?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Tree/Database.ts
+++ b/src/Explorer/Tree/Database.ts
@@ -172,6 +172,27 @@ export default class Database implements ViewModels.Database {
   public async loadCollections(): Promise<void> {
     const collectionVMs: Collection[] = [];
     const collections: DataModels.Collection[] = await readCollections(this.id());
+    // TODO Remove
+    // This is a hack to make Mongo collections read via ARM have a SQL-ish partitionKey property
+    if (userContext.apiType === "Mongo") {
+      collections.map((collection) => {
+        if (collection.shardKey) {
+          const shardKey = Object.keys(collection.shardKey)[0];
+          collection.partitionKey = {
+            version: undefined,
+            kind: "Hash",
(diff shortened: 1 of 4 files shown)


src/Explorer/Tree/Database.ts
```diff
@@ -172,6 +172,27 @@ export default class Database implements ViewModels.Database {
   public async loadCollections(): Promise<void> {
     const collectionVMs: Collection[] = [];
     const collections: DataModels.Collection[] = await readCollections(this.id());
+    // TODO Remove
+    // This is a hack to make Mongo collections read via ARM have a SQL-ish partitionKey property
+    if (userContext.apiType === "Mongo") {
+      collections.map((collection) => {
+        if (collection.shardKey) {
+          const shardKey = Object.keys(collection.shardKey)[0];
```
notlaforge, line 180: Is this order guaranteed in JS?

notlaforge (approved):

src/Explorer/Tree/Database.ts
southpolesteve: I don't think so, but there is only every going to be a single key in this object. SQL in some future may have multiple but not Mongo

PR #664, opened 2021-04-14
Tighten notebook code cell output sandbox and enable it by default

gitanuj:
- Use a custom `Html` output transform so that we can sanitize html to rename `iframe` to `iframe-disabled`
- Remove `allow-popups*` from output iframe sandbox
- Enable `sandboxnotebookoutputs` by default

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/664)


southpolesteve (approved):

src/Explorer/Notebook/NotebookClientV2.ts
```diff
@@ -34,6 +34,7 @@ import * as TelemetryProcessor from "../../Shared/Telemetry/TelemetryProcessor";
 import { userContext } from "../../UserContext";
 import configureStore from "./NotebookComponent/store";
 import { CdbAppState, makeCdbRecord } from "./NotebookComponent/types";
+import HTML from "./NotebookRenderer/outputs/html";
```
swviswan, line 37: >HTML [](start = 7, length = 4)

nit: diff name instead of using gneric name- ma be something liek SanitizedHTML?

swviswan (approved):
:shipit:

PR #677, opened 2021-04-15
Hide TTL in settings tab for Mongo API

victor-meng:
Hide ttl options for Mongo in settings tab and display a message to ask the customers to create a `_ts` index instead.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/677)


southpolesteve (approved):

southpolesteve (changes requested):
I tried out the preview link and it seems like this is hiding TTL for SQL accounts too. Cassandra shows.

victor-meng:
> I tried out the preview link and it seems like this is hiding TTL for SQL accounts too. Cassandra shows.

Thanks for catching that! I had a typo when I switched from `userContext.defaultExperience` to `userContext.apiType`.

southpolesteve (approved):

PR #683, opened 2021-04-15
Keep active tab state in one place (manager)

notlaforge:
With this change TabsBase objects will retain a reference to the TabsManager they belong to, so they can ask it if they're the
active tab or not.

This removes the possibility for bugs like activating an unmanaged tab, or having more than one active tab, etc.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/683)


src/Explorer/Tabs/TabsManager.ts
```diff
@@ -4,23 +4,25 @@ import Explorer from "../Explorer";
...
-  public activeTab: ko.Observable<TabsBase>;
+  private idx = ko.observable(0);
+  public openedTabs = ko.observableArray<TabsBase>([]);
+  public activeTab = ko.pureComputed(() => this.openedTabs()[this.idx()]);
 
   constructor() {
-    this.openedTabs = ko.observableArray<TabsBase>([]);
-    this.activeTab = ko.observable<TabsBase>();
+    this.openedTabs.subscribe(
+      (log) => log.forEach(({ value: tab, status }) => tab.manager(status === "deleted" ? undefined : this)),
```
southpolesteve, line 13: Extra log left behind?

src/Explorer/Tabs/TabsBase.ts
```diff
@@ -32,6 +33,7 @@ export default class TabsBase extends WaitsForTemplateViewModel {
   public isExecutionError: ko.Observable<boolean>;
   public isExecuting: ko.Observable<boolean>;
   public pendingNotification?: ko.Observable<DataModels.Notification>;
+  public manager = ko.observable<TabsManager>(undefined);
```
southpolesteve, line 36: Does this need to be wrapped in observable? No problem if yes just would be nice to skip if not needed

southpolesteve (approved):

src/Explorer/Tabs/TabsManager.ts
notlaforge: No, it's the array change*log*. I'll give it a better name.

src/Explorer/Tabs/TabsBase.ts
notlaforge: Thanks, I made some changes and it is no longer an Observable, nice catch.

PR #684, opened 2021-04-16
Adds retry logic to Upload JSON

southpolesteve:
Unfortunately the new SDK bulk methods are TOO FAST! In production, uploading a JSON file of just 50 items will cause throttling. This PR adds some basic retry logic for throttled document creations when uploading a JSON file

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/684?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Tree/Collection.ts
+++ b/src/Explorer/Tree/Collection.ts
@@ -18,6 +18,7 @@ import { UploadDetailsRecord } from "../../Contracts/ViewModels";
 import { Action, ActionModifiers } from "../../Shared/Telemetry/TelemetryConstants";
 import * as TelemetryProcessor from "../../Shared/Telemetry/TelemetryProcessor";
 import { userContext } from "../../UserContext";
+import { logConsoleInfo } from "../../Utils/NotificationConsoleUtils";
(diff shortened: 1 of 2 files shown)


victor-meng (approved):

src/Explorer/Tree/Collection.ts
```diff
@@ -1145,3 +1161,7 @@ export default class Collection implements ViewModels.Collection {
     }
   }
 }
+
+function sleep(seconds: number) {
```
zfoster, line 1165: seconds doesn't seem like the right var name for this

zfoster: possibly could move this to a util file if it exists, may be useful elsewhere in the project

zfoster (approved):
would change the param name for sleep, otherwise good.

src/Explorer/Tree/Collection.ts
zfoster: seconds actually is fine I think, I just found it confusing that we're passing `retryAttempts` as a `seconds` param without any backoff description

PR #685, opened 2021-04-16
Fix Stored Procedures for Non-Partitioned Collections

southpolesteve:
Changing the default here from `undefined` to  `""` ended up breaking non-partitioned stored procedure execution 

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/685?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Panes/ExecuteSprocParamsPanel/index.tsx
+++ b/src/Explorer/Panes/ExecuteSprocParamsPanel/index.tsx
@@ -30,7 +30,7 @@ export const ExecuteSprocParamsPanel: FunctionComponent<ExecuteSprocParamsPanePr
 }: ExecuteSprocParamsPaneProps): JSX.Element => {
   const [isLoading, { setTrue: setLoadingTrue, setFalse: setLoadingFalse }] = useBoolean(false);
   const [paramKeyValues, setParamKeyValues] = useState<UnwrappedExecuteSprocParam[]>([{ key: "string", text: "" }]);
-  const [partitionValue, setPartitionValue] = useState<string>("");
+  const [partitionValue, setPartitionValue] = useState<string>(); // Defaulting to undefined here is important. It is not the same partition key as ""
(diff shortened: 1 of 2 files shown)


src/Explorer/Panes/ExecuteSprocParamsPanel/index.tsx
```diff
@@ -30,7 +30,7 @@ export const ExecuteSprocParamsPanel: FunctionComponent<ExecuteSprocParamsPanePr
 }: ExecuteSprocParamsPaneProps): JSX.Element => {
   const [isLoading, { setTrue: setLoadingTrue, setFalse: setLoadingFalse }] = useBoolean(false);
   const [paramKeyValues, setParamKeyValues] = useState<UnwrappedExecuteSprocParam[]>([{ key: "string", text: "" }]);
-  const [partitionValue, setPartitionValue] = useState<string>("");
+  const [partitionValue, setPartitionValue] = useState<string>(); // Defaulting to undefined here is important. It is not the same partition key as ""
```
notlaforge, line 33: Where does that break? Just tryina see if there's a different way to encode that than a comment.

southpolesteve: A stored procedure needs a partition key to execute. For Non-partitioned collections it default to be the "undefined" case which is actually sent as `[]` in the partition key header. This case is unique to non-partitioned collections. 

Normally undefined is sent as `[{}]` for partitioned collections. 

Adding the default `""` caused non-partitioned collections to get their partition key header sent as `[""]` which will return no results since usually* there is no data in that partition.

Unfortunately our partition keys have a lot of strangeness on the wire where they intersect with falsey values. These are all different PKs:
- `[""]`
- `[]`
- `[null]`
- `[{}]`

*except there can be. We started letting people use non-partitioned collections as partitioned and they expose a system defined partition key

notlaforge (approved):

PR #686, opened 2021-04-16
Add CellOutputViewer for SandboxOutputs

gitanuj:
- Adds `SandboxOutputs` which loads `<iframe>` for every code output cell
- Adds `/cellOutputViewer.html` which relies on passed props to render a code cell output independently
- Uses [post-robot](https://github.com/krakenjs/post-robot) for messaging b/w windows
- Uses [iframe-resizer-react](https://github.com/davidjbradshaw/iframe-resizer-react) to auto resize sandbox frame to fit content
- Use `HTMLInlineCSSWebpackPlugin` and `InlineChunkHtmlPlugin` to inline CSS and JS for `cellOutputViewer.html` so that the browser can take advantage of client side caching
- Reduced the size of `cellOutputViewer` bundle by lazy loading all transforms (prod bundle is 654kB)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/686)


src/NotebookOutputViewer/NotebookOutputViewer.tsx
```diff
@@ -0,0 +1,74 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const data = (event as any).data;
+      const props: OutputsProps = {
+        id: data.id,
+        contentRef: data.contentRef,
+        hidden: data.hidden,
+        expanded: data.expanded,
+        outputs: data.outputs,
+        onMetadataChange: data.onMetadataChange,
+      };
```
languy, line 40: nit: can't we just use `event.data` as-is or we actually need a subset of the fields?
```suggestion
      const props: OutputsProps = (event as any).data;
```

src/NotebookOutputViewer/NotebookOutputViewer.tsx
```diff
@@ -0,0 +1,74 @@
...
+                onMetadataChange={(metadata, mediaType) => props.onMetadataChange(metadata, mediaType, index)}
+              />
+              <KernelOutputError />
+              <StreamText />
+            </Output>
+          ))}
+        </div>
+      );
+
+      ReactDOM.render(outputs, document.getElementById("notebookOutput"));
```
languy, line 68: nit: "cellOutput"?

src/NotebookOutputViewer/TransformMedia.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onMetadataChange: (metadata: JSONObject, mediaType: string) => void;
+}
+
+export const TransformMedia = (props: TransformMediaProps): JSX.Element => {
+  const { Media, mediaType, data, metadata } = getMediaInfo(props);
+
+  // If we had no valid result, return an empty output
+  if (!mediaType || !data) {
+    // eslint-disable-next-line no-null/no-null
+    return null;
```
languy, line 80: nit: can we return `undefined` instead?

src/NotebookOutputViewer/TransformMedia.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
```
languy, line 84: Very nice use of `React.lazy()` with `<Suspense>`!

src/NotebookOutputViewer/notebookOutputViewer.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
languy, line 1: nit: `CellOutputViewer`?

languy (approved):

src/NotebookOutputViewer/TransformMedia.tsx
gitanuj: React either needs a react node or null to render. It doesn't understand `undefined`. However I can replace this with `<></>` instead.

southpolesteve: If `null` is really needed that is ok, but it should also get a comment above the eslint comment stating why it is needed. I think we should do this for anytime we disable linting

externals/iframeResizer.contentWindow.min.js
```diff
@@ -0,0 +1,10 @@
+/*! iFrame Resizer (iframeSizer.contentWindow.min.js) - v4.3.1 - 2021-01-11
+ *  Desc: Include this file in any page being loaded into an iframe
+ *        to force the iframe to resize to the content size.
+ *  Requires: iframeResizer.min.js on host page.
+ *  Copyright: (c) 2021 David J. Bradshaw - dave@bradshaw.net
+ *  License: MIT
```
swviswan, line 6: >  License: MIT [](http://example.com/codeflow?start=2&length=14)

This license is good for us to use - right?

@southpolesteve  - do we have to declare all the open soure packages we use anywhere? Or because, we ourselves are open source, we dont have to bother?

src/CellOutputViewer/transforms/Vega2.ts
```diff
@@ -0,0 +1 @@
+export { Vega2 as default } from "@nteract/transform-vega";
```
swviswan, line 1: >Vega2 [](http://example.com/codeflow?start=9&length=5)

Can you not export all the transforms from one single file?

do you really need a file per transform?

webpack.config.js
```diff
@@ -233,6 +245,7 @@ module.exports = function (env = {}, argv = {}) {
     watchOptions: isCI ? { poll: 24 * 60 * 60 * 1000 } : {},
     devServer: {
       hot: false,
+      disableHostCheck: true,
```
swviswan, line 248: >disableHostCheck [](http://example.com/codeflow?start=6&length=16)

why this?

"THIS IS NOT RECOMMENDED as apps that do not check the host are vulnerable to DNS rebinding attacks."

src/CellOutputViewer/CellOutputViewer.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+const onInit = async () => {
+  postRobot.on(
+    "props",
+    {
+      window: window.parent,
+      domain: window.location.origin,
+    },
+    (event) => {
```
swviswan, line 32: >event [](http://example.com/codeflow?start=5&length=5)

and because we only allow same origin, we dont have to do additional checks on this message before processing it - right?

src/CellOutputViewer/transforms/Vega2.ts
gitanuj: In order to use `React.lazy` I had to declare separate exports with defaults. In this case it is `Vega2`. Another way is to use it like `React.lazy(() => import("@nteract/transform-vega").then(vega => { default: vega.Vega2 })`. Thought it is cleaner this way.

webpack.config.js
gitanuj: This is needed so that webpack hot reloading does not freak out when using <iframe> with sandbox. 

I noticed that somewhere in sockjs things went wrong when trying to read origin for the child <iframe> which is not possible for sandbox frames. Since we use hot reloading only in dev environment I only disabled host check to get around this issue there. This DOES NOT AFFECT PROD.

src/CellOutputViewer/CellOutputViewer.tsx
gitanuj: Those checks are taken care by `postRobot`. If you notice we provide `window` and `domain` when using `postRobot` to listen for messages. It takes care of making sure the message comes from that window and domain.

externals/iframeResizer.contentWindow.min.js
gitanuj: I think it should be okay. We're not modifying this library and are using it as is.

swviswan (changes requested):
:clock1:

swviswan (approved):
:shipit:

PR #700, opened 2021-04-19
Add a feature flag to override Juno endpoint

gitanuj:
Verified that the juno endpoint override works for
- Standalone DE
- DE running in Portal
- Standalone gallery

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/700)

### DIFF
--- a/src/ConfigContext.ts
+++ b/src/ConfigContext.ts
@@ -26,6 +26,7 @@ export interface ConfigContext {
   GITHUB_CLIENT_SECRET?: string; // No need to inject secret for prod. Juno already knows it.
   hostedExplorerURL: string;
   armAPIVersion?: string;
+  allowedJunoOrigins: string[];
 }
 
 // Default configuration
@@ -53,6 +54,13 @@ let configContext: Readonly<ConfigContext> = {
   GITHUB_CLIENT_ID: "6cb2f63cf6f7b5cbdeca", // Registered OAuth app: https://github.com/settings/applications/1189306
   JUNO_ENDPOINT: "https://tools.cosmos.azure.com",
   BACKEND_ENDPOINT: "https://main.documentdb.ext.azure.com",
+  allowedJunoOrigins: [
+    "https://juno-test.documents-dev.windows-int.net",
+    "https://juno-test2.documents-dev.windows-int.net",
+    "https://tools.cosmos.azure.com",
+    "https://tools-staging.cosmos.azure.com",
+    "https://localhost",
+  ],
 };
 
 export function resetConfigContext(): void {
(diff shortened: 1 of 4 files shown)


src/Platform/Hosted/extractFeatures.ts
```diff
@@ -27,7 +28,8 @@ export type Features = {
 export function extractFeatures(given = new URLSearchParams(window.location.search)): Features {
   const downcased = new URLSearchParams();
   const set = (value: string, key: string) => downcased.set(key.toLowerCase(), value);
-  const get = (key: string, defaultValue?: string) => downcased.get("feature." + key) ?? defaultValue;
+  const get = (key: string, defaultValue?: string) =>
+    downcased.get("feature." + key) ?? downcased.get(key) ?? defaultValue;
```
gitanuj, line 32: @southpolesteve This is needed to honor feature flags passed down by the Portal. I see that in #659 you added support for passing down portal feature flags down to DE. However those flags are not prefixed with "feature." so we were not honoring them in DE.

src/ConfigContext.ts
```diff
@@ -53,6 +54,13 @@ let configContext: Readonly<ConfigContext> = {
   GITHUB_CLIENT_ID: "6cb2f63cf6f7b5cbdeca", // Registered OAuth app: https://github.com/settings/applications/1189306
   JUNO_ENDPOINT: "https://tools.cosmos.azure.com",
   BACKEND_ENDPOINT: "https://main.documentdb.ext.azure.com",
+  allowedJunoOrigins: [
+    "https://juno-test.documents-dev.windows-int.net",
+    "https://juno-test2.documents-dev.windows-int.net",
+    "https://tools.cosmos.azure.com",
+    "https://tools-staging.cosmos.azure.com",
+    "https://localhost",
```
swviswan, line 62: >localhost [](http://example.com/codeflow?start=13&length=9)

could there be problems because of this + this code being open source?

src/Platform/Hosted/extractFeatures.ts
victor-meng: Alternatively we could also strip away any `feature.` in each string inside the `set` function so we don't have to handle both cases in `get`. Might be a bit cleaner this way.

@southpolesteve @notlaforge Which one do you prefer?

swviswan (approved):
:shipit:

src/Platform/Hosted/extractFeatures.ts
southpolesteve: No real preference from me. We can always refactor this later.

southpolesteve (approved):

PR #720, opened 2021-04-23
Lazy loaded Monaco

notlaforge:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/720)

Before:

image

After:

image


notlaforge:
There's still a lot of room for improvement!

image

southpolesteve (approved):

southpolesteve:
👏 👏 👏 

This has been a dream ever since we switched to Webpack. I think our customers on slower connections will be see a real improvement.

notlaforge:
Depending on which Monaco features we're willing to give up on, or at the very least which languages, we could shrink the loaded monaco a bit too by turning some on and off in the webpack plugin that bundles monaco:

image

PR #730, opened 2021-04-27
Remove userContext.defaultExperience

sunilyadav840:
We are using `userContext.apiType` instead


src/Common/DocumentUtility.ts
```diff
@@ -2,7 +2,7 @@ import { DefaultAccountExperienceType } from "../DefaultAccountExperienceType";
 import { userContext } from "../UserContext";
 
 export const getEntityName = (): string => {
-  if (userContext.defaultExperience === DefaultAccountExperienceType.MongoDB) {
+  if (userContext.apiType === DefaultAccountExperienceType.MongoDB) {
```
southpolesteve, line 5: This right here should be throwing a compiler error. It does in vscode for me! But not when running tsc or in CI.
We should wait on this PR until we have unraveled this mystery

sunilyadav840: I have changed value of DefaultAccountExperienceType. as we have discussed mapping between DefaultAccountExperienceType and apiType value.


`export enum DefaultAccountExperienceType {
  DocumentDB = "SQL",
  Graph = "Gremlin",
  MongoDB = "Mongo",
  Table = "Tables",
  Cassandra = "Cassandra",
  ApiForMongoDB = "Mongo",
}
`

type ApiType = "SQL" | "Mongo" | "Gremlin" | "Tables" | "Cassandra";

If I don't change the value of DefaultAccountExperienceType, so It will give following error and also give typescript error.

This condition will always return 'false' since the types 'ApiType' and 'DefaultAccountExperienceType.MongoDB' have no overlap.ts(2367)

southpolesteve: I actually want that behavior. Remove `DefaultAccountExperienceType` completely and change all comparisons to use the literal strings:

userContext.apiType === "Mongo"

southpolesteve (approved):

src/Common/DocumentUtility.ts
southpolesteve: Resolve by removing DefaultAccountExperienceType

PR #733, opened 2021-04-28
Migration/left panel to react

sunilyadav840:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/733)


src/Main.tsx
```diff
@@ -93,6 +94,29 @@ const App: React.FunctionComponent = () => {
...
+    min: ExplorerMetrics.SplitterMinWidth,
+    max: ExplorerMetrics.SplitterMaxWidth,
+  };
+
+  const splitter = new Splitter({
+    splitterId: "h_splitter1",
+    leftId: "resourcetree",
+    bounds: splitterBounds,
+    direction: SplitterDirection.Vertical,
+  });
```
southpolesteve, line 107: I just tried this in production and the splitter does not work even before this change. Remove it entirely.

southpolesteve (approved):

src/Main.tsx
sunilyadav840: Removed

PR #734, opened 2021-04-28
Enable the "Enable notebooks" button

srinathnarayanan:
Earlier, the "Enable Notebooks" button in the command bar was enabled only if the databaseAccount location was not a disallowed location. With these changes, the same check is made on both 
- database account location
- first write region of the database account

### DIFF
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -285,8 +285,8 @@ export default class Explorer {
           async () => {
             this.isNotebookEnabled(
               userContext.authType !== AuthType.ResourceToken &&
-                ((await this._containsDefaultNotebookWorkspace(this.databaseAccount())) ||
-                  userContext.features.enableNotebooks)
+              ((await this._containsDefaultNotebookWorkspace(this.databaseAccount())) ||
+                userContext.features.enableNotebooks)
             );
 
             TelemetryProcessor.trace(Action.NotebookEnabled, ActionModifiers.Mark, {
@@ -308,7 +308,7 @@ export default class Explorer {
                 this.isSparkEnabledForAccount() &&
                 this.arcadiaWorkspaces() &&
                 this.arcadiaWorkspaces().length > 0) ||
-                userContext.features.enableSpark
+              userContext.features.enableSpark
             );
             if (this.isSparkEnabled()) {
               trackEvent(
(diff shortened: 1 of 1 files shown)


src/Explorer/Explorer.tsx
```diff
@@ -1754,9 +1755,12 @@ export default class Explorer {
         this.isNotebooksEnabledForAccount(true);
         return;
       }
-      const isAccountInAllowedLocation = !disallowedLocations.some(
-        (disallowedLocation) => disallowedLocation === databaseAccountLocation
-      );
+
+      // databaseAccountLocation or firstWriteLocation are not disallowed
+      const isAccountInAllowedLocation =
+        disallowedLocations.indexOf(databaseAccountLocation) === -1 ||
```
swviswan, line 1761: >databaseAccountLocation [](http://example.com/codeflow?start=36&length=23)

do we need this check?

didnt we see that the control plane was using only the first write loc?

srinathnarayanan: modified it.

src/Explorer/Explorer.tsx
```diff
@@ -1754,9 +1754,9 @@ export default class Explorer {
         this.isNotebooksEnabledForAccount(true);
         return;
       }
-      const isAccountInAllowedLocation = !disallowedLocations.some(
-        (disallowedLocation) => disallowedLocation === databaseAccountLocation
-      );
+
+      // firstWriteLocation should not be disallowed
+      const isAccountInAllowedLocation = disallowedLocations.indexOf(firstWriteLocation) === -1;
       this.isNotebooksEnabledForAccount(isAccountInAllowedLocation);
```
swviswan, line 1760: >isNotebooksEnabledForAccount [](http://example.com/codeflow?start=11&length=28)

can you add telemetry for how many ppl click this button and enable it?

srinathnarayanan: there is already telemetry available for "CreateNotebookWorkspace" action which is triggered while enabling notebooks.

src/Explorer/Explorer.tsx
```diff
@@ -1729,7 +1729,7 @@ export default class Explorer {
     }
 
     const databaseAccount = this.databaseAccount();
-    const databaseAccountLocation = databaseAccount && databaseAccount.location.toLowerCase();
+    const firstWriteLocation = databaseAccount?.properties?.writeLocations[0].locationName.toLowerCase();
```
gitanuj, line 1732: Do we need to check if `writeLocations` is defined and contains at least 1 element?

gitanuj (approved):

src/Explorer/Explorer.tsx
srinathnarayanan: Fixing it in the next iterations

PR #735, opened 2021-04-29
cleanup/remove-refresh-spark-enabled-state-for-account

sunilyadav840:
Move Explorer._refreshSparkEnabledStateForAccount into React


southpolesteve:
This is no longer needed. We removed the code that required checking if spark was enabled

PR #742, opened 2021-05-03
Portal changes for DedicatedGateway

fnbalaji:
Portal changes for DedicatedGateway

1. Change Sqlx endpoints to SqlDedicatedGateway endpoint

2. Remove D32s from the SKU list

3. Add telemetry

4. Remove SKU details field per discussion

5. Support dynamic instance scaling.


src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -30,39 +32,66 @@ export const updateDedicatedGatewayResource = async (sku: string, instances: num
...
 export const deleteDedicatedGatewayResource = async (): Promise<string> => {
   const path = getPath(userContext.subscriptionId, userContext.resourceGroup, userContext.databaseAccount.name);
-  const armRequestResult = await armRequestWithoutPolling({
-    host: configContext.ARM_ENDPOINT,
-    path,
-    method: "DELETE",
-    apiVersion,
-  });
-  return armRequestResult.operationStatusUrl;
+  const telemetryData = { selfServeClassName: SqlX.name };
```
srinathnarayanan, line 59: add data to differentiate between DELETE and GET Rest calls

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -206,9 +180,22 @@ export default class SqlX extends SelfServeBaseClass {
           },
         };
       } else {
-        // Check for scaling up/down/in/out
+        let operationStatusUrl;
+        if (baselineValues.get("instances")?.value === currentValues.get("instances")?.value) {
+          currentValues.set("warningBanner", {
```
srinathnarayanan, line 185: I dont think it will ever get to a state where instances value is the same, yet saving is possible right? The save button will be disabled in this case. Any particular reason you added this?

srinathnarayanan (approved):

src/SelfServe/SqlX/SqlX.tsx
fnbalaji: This will happen if you move from 3 to 2 and move back to 3. Save button will unfreeze when you move from 3 to 2 and stay unfrozen when you move back from 2 to 3.

swviswan: What is the scenario: 3 -2 is pending? and when that is the user changes from 2 -3? 

in those cases, how does the rp API behave? will it fail that an update is in progress / does it let it go through?

src/SelfServe/SqlX/SqlX.rp.ts
swviswan: do you also want to add any account name?
@srinathnarayanan  - I beleieve we ll get user PUID, please confirm if the telemetry is wired that way.

src/SelfServe/SqlX/SqlX.tsx
srinathnarayanan: there was a bug on the selfserve model that did not disable save/discard button when an initial value of a property is set to "undefined". PR is out to fix that : https://github.com/Azure/cosmos-explorer/pull/764

This check will no longer be needed after that fix goes in.

src/SelfServe/SqlX/SqlX.rp.ts
srinathnarayanan: We get these params for free on every telemetry message from DE
    databaseAccountName,
    defaultExperience,
    authType,
    subscriptionId,
    platform,
    env

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -37,43 +37,56 @@ const connectionStringValue: Description = {
...
-  });
-
+  const dedicatedGatewayOriginallyEnabled = baselineValues.get("enableDedicatedGateway")?.value as boolean;
+  const baselineInstances = baselineValues.get("instances")?.value as number;
+  if (!dedicatedGatewayOriginallyEnabled || baselineInstances !== newValue) {
+    currentValues.set("warningBanner", {
+      value: {
+        textTKey: "WarningBannerOnUpdate",
+        link: {
+          href: "https://docs.microsoft.com/en-us/azure/cosmos-db/introduction",
```
swviswan, line 81: >https://docs.microsoft.com/en-us/azure/cosmos-db/introductio [](http://example.com/codeflow?start=17&length=60)

please use aka ms links.

src/Localization/en/SqlX.json
```diff
@@ -37,16 +37,20 @@
...
     "CostText": "Hourly cost of the dedicated gateway resource depends on the SKU selection, number of instances per region, and number of regions.",
     "ConnectionString": "Connection String",
     "ConnectionStringText": "To use the dedicated gateway, use the connection string shown in ",
-    "KeysBlade": "the keys blade",
+    "KeysBlade": "the keys blade.",
+    "MetricsString": "Metrics",
+    "MetricsText": "Monitor the \"DedicatedGatewayCpuUsage\" and \"DedicatedGatewayMemoryUsage\" in ",
+    "MetricsBlade": "the metrics blade.",
+    "ResizingDecisionText": "To understand if the dedicated gateway is the right size, ",
+    "ResizingDecisionLink": "learn more about dedicated gateway sizing.",
```
swviswan, line 52: >learn more about dedicated gateway sizing. [](http://example.com/codeflow?start=29&length=42)

do you wnat the hyperlink for just learn more instead of this whole string?

src/Localization/en/SqlX.json
```diff
@@ -37,16 +37,20 @@
...
+    "NoChangesToExistingResource": "No changes were made to the existing resource. Saving will have no effect.",
```
swviswan, line 53: >No changes were made to the existin [](http://example.com/codeflow?start=36&length=35)

i htink based on Srinath's comment, you no longer need this. please follow up with him if needed.

swviswan (approved):
:shipit:

src/Localization/en/SqlX.json
srinathnarayanan: Yes, these need to be removed,

src/Localization/en/SqlX.json
fnbalaji: Yes, that is the plan.

PR #757, opened 2021-05-06
Fix table edit entity bug and add collection panel bug for connection string users

victor-meng:
Table edit entity bug:
- The boolean entity value could be either string or boolean so we can call `.toString` to convert boolean value into string.

Table int64 bug:
- We need to pass the entity value of a int64 type as a string instead of number.

Add collection panel bug:
- Add null check for `databaseAccount.properties` and `databaseAccount.properties.capabilities`.

### DIFF
--- a/src/Explorer/Tables/TableEntityProcessor.ts
+++ b/src/Explorer/Tables/TableEntityProcessor.ts
@@ -189,7 +189,7 @@ export function convertEntityToNewDocument(entity: Entities.ITableEntityForTable
           parsedValue = parseInt(propertyValue, 10);
           break;
         case Constants.TableType.Int64:
-          parsedValue = parseInt(propertyValue, 10).toString();
+          parsedValue = propertyValue.toString();
           break;
         case Constants.TableType.Double:
           parsedValue = parseFloat(propertyValue);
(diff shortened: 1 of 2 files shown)


src/Explorer/Tables/TableEntityProcessor.ts
```diff
@@ -183,12 +183,14 @@ export function convertEntityToNewDocument(entity: Entities.ITableEntityForTable
...
         case Constants.TableType.Boolean:
-          parsedValue = propertyValue.toLowerCase() === "true";
+          parsedValue = propertyValue.toString().toLowerCase() === "true";
           break;
         case Constants.TableType.Int32:
-        case Constants.TableType.Int64:
           parsedValue = parseInt(propertyValue, 10);
           break;
+        case Constants.TableType.Int64:
+          parsedValue = parseInt(propertyValue, 10).toString();
```
southpolesteve, line 192: Remove `parseInt` here. JS will parse this as a double, but in reality it could be larger than that. Just return the property value directly.

victor-meng: Done

southpolesteve (approved):

PR #769, opened 2021-05-11
Configure eslint plugin jsx a11y

sunilyadav840:
Configure eslint-plugin-jsx-a11y and fixes the issues.


southpolesteve:
@sunilyadav840 I tried getting this to pass, but I think the issue is actually we are still using `GenericRightPaneComponent` in several places which is causing an end to end test to fail. Please open a PR that remove `GenericRightPaneComponent` entirely and then we will come back to this.

sunilyadav840:
@southpolesteve I disabled one rule for `GenericRightPaneComponent` , Then all E2E test cases passed. 
Only one test cases failed. `CI / E2E (./test/sql/resourceToken.spec.ts)` I am not sure that this is failing for new eslint configiration.

PR #773, opened 2021-05-12
Fix throughput input component and add database panel

victor-meng:
This PR fixes the following issues:
- `Throughputinput` component should maintain the throughput value, `isAutoscaleSelected` flag, and `isCostAcknowledged` flag as states and re-render itself when these values change. It shouldn't rely on the parent components to keep track of these states and re-render. The parent component should only be responsible for reading these values.
- A lot of small fixes in `AddDatabasePane` which are mostly styling fixes and removing redundant code.
- Fixes a console error in `useSidePanel` where `onCloseCallback.callback` can be undefined and throws an exception when called.

### DIFF
--- a/src/Explorer/Controls/ThroughputInput/ThroughputInput.test.tsx
+++ b/src/Explorer/Controls/ThroughputInput/ThroughputInput.test.tsx
@@ -19,18 +19,4 @@ describe("ThroughputInput Pane", () => {
   it("should render Default properly", () => {
     expect(wrapper).toMatchSnapshot();
   });
-
-  it("test Autoscale Mode select", () => {
-    wrapper.setProps({ isAutoscaleSelected: true });
-    expect(wrapper.find('[aria-label="ruDescription"]').at(0).text()).toBe(
-      "Estimate your required RU/s with capacity calculator."
-    );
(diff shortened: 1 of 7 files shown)


src/Explorer/Controls/ThroughputInput/ThroughputInput.test.tsx
```diff
@@ -19,18 +19,4 @@ describe("ThroughputInput Pane", () => {
...
-      "Estimate your required RU/s with capacity calculator."
-    );
-    expect(wrapper.find('[aria-label="maxRUDescription"]').at(0).text()).toContain("Max RU/s");
-  });
-
-  it("test Manual Mode select", () => {
-    wrapper.setProps({ isAutoscaleSelected: false });
-    expect(wrapper.find('[aria-label="ruDescription"]').at(0).text()).toContain("Estimate your required RU/s with");
-    expect(wrapper.find('[aria-label="capacityLink"]').at(0).text()).toContain("capacity calculator");
-  });
```
southpolesteve, line 35: What's the reason for deleting these? Is this tested anywhere else?

southpolesteve (approved):

PR #799, opened 2021-05-19
Use Zustand for side panel state

southpolesteve:
https://github.com/pmndrs/zustand

How to complete this PR:

1. Fix all typescript errors
2. Remove all open panel methods from Explorer. Panel open method should go where they are called from originally
   1. Anywhere that uses openPanel and is a functional component, call `useSidePanel()` hook
   2. Anywhere there uses openPanel and is a class component or in knockout still, call `useSidePanel().getState()`

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/799?feature.someFeatureFlagYouMightNeed=true)


languy (approved):

src/Explorer/Explorer.tsx
```diff
@@ -1568,160 +1548,115 @@ export default class Explorer {
       return false;
     });
   }
-
   public openDeleteCollectionConfirmationPane(): void {
```
southpolesteve, line 1551: @hardiknai-techm can you also remove all these methods from Explorer? For example instead of having `openDeleteCollectionConfirmationPane` just call `useSidePanel.getState().openSidePanel()` from the place that is calling this. It must already have access to Explorer since it is calling Explorer.openDeleteCollectionConfirmationPane

PR #807, opened 2021-05-20
Update generated ARM clients to latest version

southpolesteve:
1. Updates generated cosmos ARM API to version 2021-04-15
2. Fixes a bug in generator where some enums were being types as strings. These are now typed as correct string unions
3. Adds generated ARM methods for notebooks and Synapse
4. Cleanup arm folder structure a bit and add details to header comments about where the source rest specs are 
5. TODO: Will be doing a second PR to replace notebooks client with generated methods

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/807?feature.someFeatureFlagYouMightNeed=true)


src/Explorer/Tabs/ScriptTabBase.ts
```diff
@@ -205,16 +205,6 @@ export default abstract class ScriptTabBase extends TabsBase implements ViewMode
     return Q();
   };
 
-  public onSaveOrUpdateClick(): Promise<any> {
```
southpolesteve, line 208: Unused

zfoster (approved):

PR #813, opened 2021-05-21
Cleanup Synapse+Spark Logic

southpolesteve:
One of the current challenges in the React/Fluent migration work is how coupled everything it to `Explorer.tsx`. One particularly trouble area is how we load spark pools from Synapse that our notebooks can run on top of instead of Juno. Its a bit like a load bearing wall that is holding up a bunch of other work.

I did some digging and our telemetry shows this feature is **entirely unused**. If needed at some future time we can look at doing it in a React friendly way.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/813?feature.someFeatureFlagYouMightNeed=true)


languy (approved):
I would recommend to remove the unused code entirely (commented and uncommented): 
* if I had to resurrect the feature, I doubt it'd be just searching for commented code to uncomment, but I'd instead look at the changes of this PR.
* by the time we want to put this back, DE would be further along in the react port that the code most likely won't work anymore. "Arcadia" is the old name for "synapse" I believe, so we may as well rename things in the process.
* the remaining live unused code (like the `arcadiaToken`) is a distraction to our react port.

As long as the commit for this PR is easy to find (by title and description), there is no need to maintain live unused code and commented code in in my opinion.

PR #826, opened 2021-05-25
Hide hosted shells and schema analyzer if VNET, Firewall or Private endpoints is enabled

gitanuj:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/826)

### DIFF
--- a/src/Contracts/DataModels.ts
+++ b/src/Contracts/DataModels.ts
@@ -22,6 +22,7 @@ export interface DatabaseAccountExtendedProperties {
   enableAnalyticalStorage?: boolean;
   isVirtualNetworkFilterEnabled?: boolean;
   ipRules?: IpRule[];
+  privateEndpointConnections?: unknown[];
 }
 
 export interface DatabaseAccountResponseLocation {
--- a/src/Explorer/Notebook/SchemaAnalyzer/SchemaAnalyzerUtils.ts
+++ b/src/Explorer/Notebook/SchemaAnalyzer/SchemaAnalyzerUtils.ts
@@ -1,5 +1,6 @@
 import { Notebook } from "@nteract/commutable";
 import { IContent } from "@nteract/types";
+import { userContext } from "../../../UserContext";
 import * as InMemoryContentProviderUtils from "../NotebookComponent/ContentProviders/InMemoryContentProviderUtils";
 
 const notebookName = "schema-analyzer-component-notebook.ipynb";
@@ -42,3 +43,12 @@ export const SchemaAnalyzerNotebook: IContent<"notebook"> = {
   content: notebook,
   format: "json",
 };
+
+export function isSchemaAnalyzerSupported(isNotebookEnabled: boolean): boolean {
+  return (
+    userContext.apiType === "Mongo" &&
+    isNotebookEnabled &&
+    !userContext.databaseAccount?.properties?.isVirtualNetworkFilterEnabled &&
+    !userContext.databaseAccount?.properties?.ipRules?.length
+  );
+}
--- a/src/Explorer/Notebook/SchemaAnalyzer/SchemaAnalyzerUtils.ts
+++ b/src/Explorer/Notebook/SchemaAnalyzer/SchemaAnalyzerUtils.ts
@@ -1,6 +1,5 @@
 import { Notebook } from "@nteract/commutable";
 import { IContent } from "@nteract/types";
-import { userContext } from "../../../UserContext";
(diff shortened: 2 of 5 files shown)


src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -276,6 +276,8 @@ export class ResourceTreeAdapter implements ReactAdapter {
     if (
       userContext.apiType === "Mongo" &&
       this.container.isNotebookEnabled() &&
+      !userContext.databaseAccount.properties.isVirtualNetworkFilterEnabled &&
```
swviswan, line 279: >isVirtualNetworkFilterEnabled [](http://example.com/codeflow?start=46&length=29)

you might need another property for private endpoints..

gitanuj: Still figuring out best way to figure that out. Will send a follow up PR with that change when it's ready.

src/Explorer/Notebook/SchemaAnalyzer/SchemaAnalyzerUtils.ts
```diff
@@ -42,3 +43,12 @@ export const SchemaAnalyzerNotebook: IContent<"notebook"> = {
   content: notebook,
   format: "json",
 };
+
+export function isSchemaAnalyzerSupported(isNotebookEnabled: boolean): boolean {
+  return (
+    userContext.apiType === "Mongo" &&
+    isNotebookEnabled &&
```
swviswan, line 50: >isNotebookEnabled [](http://example.com/codeflow?start=4&length=17)

niy: boolean check is less expensive and can be moved first

src/Explorer/Tree/ResourceTreeAdapter.tsx
```diff
@@ -273,11 +274,7 @@ export class ResourceTreeAdapter implements ReactAdapter {
       contextMenu: ResourceTreeContextMenuButtonFactory.createCollectionContextMenuButton(this.container, collection),
     });
 
-    if (
-      userContext.apiType === "Mongo" &&
-      this.container.isNotebookEnabled() &&
-      userContext.features.enableSchemaAnalyzer
-    ) {
+    if (SchemaAnalyzerUtils.isSchemaAnalyzerSupported(this.container.isNotebookEnabled())) {
```
swviswan, line 277: >isSchemaAnalyzerSupported [](http://example.com/codeflow?start=28&length=25)

should this really be within schemaUtils?

gn that you pass notebook enabled as a apram alone and get usercontext and other directly.. since this file itself has usercontext and container, keep it as a fn here.. 
if u think u ll use it in multiple places - then makes sense to move it to SchemaAnalyzerUtils?

src/Contracts/DataModels.ts
```diff
@@ -22,6 +22,7 @@ export interface DatabaseAccountExtendedProperties {
   enableAnalyticalStorage?: boolean;
   isVirtualNetworkFilterEnabled?: boolean;
   ipRules?: IpRule[];
+  privateEndpointConnections?: unknown[];
```
swviswan, line 25: >unknown [](http://example.com/codeflow?start=31&length=7)

Can you check ?

swviswan (approved):
:shipit:

src/Contracts/DataModels.ts
gitanuj: I referred to generatedArmClient that we have and that also uses `unknown[]` data type 
https://github.com/Azure/cosmos-explorer/blob/0201e6ff92f218085ffcecec180305256d8a0b12/src/Utils/arm/generatedClients/cosmos/types.ts#L424

@southpolesteve do you know if this property can be strongly typed?

srinathnarayanan (approved):

src/Contracts/DataModels.ts
southpolesteve: The generator has some edge cases it does not handle so it will bail out to `unknown `.  File an issue and I'll look into it.

Its ok to edit the generated type here. Write a comment saying why you edited it. I don't think we should get in the habit of editing this file but exceptions will exist.

gitanuj: Is it okay to check this in and have a follow up to change this to a strongly typed definition instead? That would require some testing (for which I'll need Srinath's help) and would delay the fix.

gitanuj: Also created an issue here for tracking
https://github.com/Azure/cosmos-explorer/issues/841

southpolesteve: Yes. If you don't need the contents of the `unknown` its fine to leave this as is.

PR #827, opened 2021-05-25
Redo user endpoint dynamic token

zfoster:

Adds forceRefresh which seems to fix the issue of switching tenants/subscriptions/db accounts. There still seems to be a race condition with db account sometimes not loading before config is given to the embedded explorer


src/Utils/AuthorizationUtils.ts
```diff
@@ -40,3 +41,21 @@ export function decryptJWTToken(token: string) {
...
+    auth: {
+      authority: "https://login.microsoftonline.com/common",
+      clientId: "203f1145-856a-4232-83d4-a43568fba23d",
+    },
+  };
+
+  if (process.env.NODE_ENV === "development") {
+    config.auth.redirectUri = "https://dataexplorer-dev.azurewebsites.net";
+  }
+  const msalInstance = new msal.PublicClientApplication(config);
```
southpolesteve, line 59: This will new up an instance every time its called. Is that needed? Why not create it at the top level and export as a const?

southpolesteve (approved):

src/Utils/AuthorizationUtils.ts
zfoster: I did this to have one instance in Explorer and one in HostedExplorer - not sure if it's needed, a const might work fine since the config is the same

victor-meng (approved):

PR #829, opened 2021-05-26
Disable caching for cellOutputViewer.html

gitanuj:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/829)

We bundle js/css along with `cellOutputViewer.html`. So if we don't disable cache the browser wouldn't pickup new changes in the file. This is causing issues with rendering cell outputs (in both notebooks and schema analyzer) in prod.

### DIFF
--- a/web.config
+++ b/web.config
@@ -42,6 +42,13 @@
       </staticContent>
     </system.webServer>
   </location>
+  <location path="cellOutputViewer.html">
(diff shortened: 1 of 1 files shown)


swviswan (approved):
:shipit:

southpolesteve (approved):

PR #851, opened 2021-05-29
Mongo tabs UX fixes

srinathnarayanan:
This PR 
- FIxes the terminal tabs and mongo shell tabs numbering. 
Earlier, 
**new tab index = number of tabs of the same type open + 1** 
So if 
   - tab 1 was opened
   - tab 2 was opened
   - tab 1 was closed
   - new tab is opened ----> New tab would be "Mongo Terminal 2", when it should be actually be "Mongo Terminal 3". 

  So I introduce an 'index' property. now 
  **new tab index = index of the latest tab (of the same type) open + 1** 
- Adds a spinner to the terminal tab when terminal is loading
- Fixed error due to which terminal tabs did not load if they were opened quickly after DE loads, and the notebookServerEndpoint is not populated yet. Now, the terminal tabs are populated only after the notebookServerEndpoint is populated to a non null value.
 
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/851)

mongo tabs numbering

### DIFF
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -1394,7 +1394,13 @@ export default class Explorer {
       tab.hashLocation().startsWith(hashLocation)
     ) as TerminalTab[];
 
-    const index = terminalTabs.length + 1;
+    let index = 1;
+    if (terminalTabs.length > 0) {
+      index = terminalTabs
+        .map((tab) => tab.getIndex())
+        .reduce((prevMaxIndex, currentIndex) => Math.max(prevMaxIndex, currentIndex));
+    }
+
     const newTab = new TerminalTab({
       account: userContext.databaseAccount,
       tabKind: ViewModels.CollectionTabKind.Terminal,
@@ -1407,6 +1413,7 @@ export default class Explorer {
       onLoadStartKey: null,
       container: this,
       kind: kind,
+      index: index + 1,
     });
 
     this.tabsManager.activateNewTab(newTab);
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -1396,9 +1396,7 @@ export default class Explorer {
 
     let index = 1;
     if (terminalTabs.length > 0) {
-      index = terminalTabs
-        .map((tab) => tab.getIndex())
-        .reduce((prevMaxIndex, currentIndex) => Math.max(prevMaxIndex, currentIndex));
+      index = terminalTabs[terminalTabs.length - 1].getIndex() + 1;
     }
 
     const newTab = new TerminalTab({
@@ -1413,7 +1411,7 @@ export default class Explorer {
       onLoadStartKey: null,
       container: this,
       kind: kind,
-      index: index + 1,
+      index: index,
     });
 
     this.tabsManager.activateNewTab(newTab);
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -1396,7 +1396,7 @@ export default class Explorer {
 
     let index = 1;
     if (terminalTabs.length > 0) {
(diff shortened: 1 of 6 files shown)


src/Explorer/Explorer.tsx
```diff
@@ -1394,7 +1394,11 @@ export default class Explorer {
       tab.hashLocation().startsWith(hashLocation)
     ) as TerminalTab[];
 
-    const index = terminalTabs.length + 1;
+    let index = 1;
+    if (terminalTabs.length > 0) {
+      index = terminalTabs[terminalTabs.length - 1].getIndex() + 1;
```
vignesh-mr, line 1399: What happens if I open 3 tabs, close tab1 and open a new one? Looks like it'll name it tab3 with this logic when we already have one with the same name?

vignesh-mr (approved):

src/Explorer/Explorer.tsx
vignesh-mr, line 1399: Do we know if these are sorted based on index?  Wondering if there will ever be a scenario where the index of the last tab will ever be smaller than the ones before it

srinathnarayanan: the new tab always gets added to the end, and will have a higher index than any preceding tab. There will be no case where index of the last tab < index of any other tab.

src/Explorer/Tabs/TerminalTab.tsx
```diff
@@ -42,16 +44,22 @@ export default class TerminalTab extends TabsBase {
   public readonly html = '<div style="height: 100%" data-bind="react:notebookTerminalComponentAdapter"></div>  ';
   private container: Explorer;
   private notebookTerminalComponentAdapter: NotebookTerminalComponentAdapter;
+  private index: number;
```
swviswan, line 47: >private index: number; [](http://example.com/codeflow?start=2&length=22)

should/can this be in TabsBase so all tabs will get it?

what about cassandra shell?

src/Explorer/Tree/Collection.ts
```diff
@@ -690,14 +690,23 @@ export default class Collection implements ViewModels.Collection {
...
 
   public onNewMongoShellClick() {
-    const id = this.container.tabsManager.getTabs(ViewModels.CollectionTabKind.MongoShell).length + 1;
+    const mongoShellTabs = this.container.tabsManager.getTabs(
+      ViewModels.CollectionTabKind.MongoShell
+    ) as MongoShellTab[];
+
+    let index = 1;
+    if (mongoShellTabs.length > 0) {
+      index = mongoShellTabs[mongoShellTabs.length - 1].getIndex() + 1;
```
swviswan, line 699: >getIndex [](http://example.com/codeflow?start=56&length=8)

i like this index being part of the tab..

but just curious: would just an incrementing index as the count in tabsManager would not just work?

srinathnarayanan: discussed offline. Synced with Deborah as well. She liked the experience in my PR as opposed to keeping an overall tab count for each type and incrememnting it, irrespective of what was the last opened tab.

src/Explorer/Tabs/TerminalTab.tsx
srinathnarayanan: done

swviswan (approved):
:shipit:

src/Explorer/Tabs/TabsBase.ts
```diff
@@ -61,6 +63,10 @@ export default class TabsBase extends WaitsForTemplateViewModel {
     };
   }
 
+  public getIndex(): number {
```
southpolesteve, line 66: No blocking nit. You don't need this method. Mark the `index` property `public readonly` and you can only write it in the constructor.

srinathnarayanan: nice catch, will change it.

PR #876, opened 2021-06-09
Migrate notebooks workspaces to generated clients

southpolesteve:
Needed step to remove window.dataExplorer

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/876?feature.someFeatureFlagYouMightNeed=true)


src/Explorer/Explorer.tsx
```diff
@@ -581,37 +584,19 @@ export default class Explorer {
...
+      databaseAccount.name,
+      "default"
+    );
 
-      this.notebookServerInfo(connectionInfo);
-      this.notebookServerInfo.valueHasMutated();
-      this.refreshNotebookList();
-    }
+    this.notebookServerInfo({
+      notebookServerEndpoint: userContext.features.notebookServerUrl || connectionInfo.notebookServerEndpoint,
```
gitanuj, line 595: nit: we don't need to make API call to fetch notebook connection info if we know that we're going to use custom notebookServerUrl and notebookServerToken from feature flags.

src/Explorer/Explorer.tsx
```diff
@@ -18,14 +18,19 @@ import * as ViewModels from "../Contracts/ViewModels";
...
-import { NotebookWorkspaceManager } from "../NotebookWorkspaceManager/NotebookWorkspaceManager";
 import { RouteHandler } from "../RouteHandlers/RouteHandler";
 import { ExplorerSettings } from "../Shared/ExplorerSettings";
 import { Action, ActionModifiers } from "../Shared/Telemetry/TelemetryConstants";
 import * as TelemetryProcessor from "../Shared/Telemetry/TelemetryProcessor";
 import { userContext } from "../UserContext";
 import { getCollectionName, getDatabaseName, getUploadName } from "../Utils/APITypeUtils";
 import { update } from "../Utils/arm/generatedClients/cosmos/databaseAccounts";
+import {
+  get as getWorkspace,
```
gitanuj, line 29: nit: If these method names are not auto generated, can we rename them to be more descriptive?

gitanuj (approved):

src/Utils/arm/request.ts
```diff
@@ -144,13 +144,13 @@ async function getOperationStatus(operationStatusUrl: string) {
...
   const status = body.status;
-  if (!status && response.status === 200) {
-    return body;
-  }
   if (status === "Canceled" || status === "Failed") {
     const errorMessage = body.error ? JSON.stringify(body.error) : "Operation could not be completed";
     const error = new Error(errorMessage);
     throw new AbortError(error);
   }
+  if (response.status === 200) {
```
vignesh-mr, line 152: I remember there was a scenario in the control plane where it returned a 2XX status code with status that was neither canceled nor failed (was "dequeued" I believe), and we kept retrying because we did not handle that. Did we ever fix that on our end?

vignesh-mr (approved):

src/Utils/arm/request.ts
southpolesteve: I believe that is handled here. I see dequeued a lot and it is always followed up with as a success/failed/cancelled status

src/Explorer/Explorer.tsx
southpolesteve: This is autogenerated. Agreed its not ideal but I would rather not modify them right now.

src/Explorer/Explorer.tsx
southpolesteve: Correct, but since its on the happy path it is much easier to just leave it in.

PR #879, opened 2021-06-10
 Fix query tab/pane related issues

victor-meng:
Fixed the following issues:
1. `Cannot read property 'continuationToken' of undefined` error: The `MongoProxyClient.queryDocuments` calls `MongoProxyClient.errorHandling` when the response contains an error. However, since `errorHandling` is an async function, we have to add `await` in the caller. Otherwise, `queryDocuments` would return an undefined result before the error is thrown, which then causes more script errors.
2. Save and browse query panes do not correctly check if the `___Cosmos` database has already been created: This is because we are lazy loading the containers. If we don't expand the `___Cosmos` database, then we do not load its containers. Therefore, we can't find the `___Query` container and the check returns false. The fix is to always load the containers if the database id is `___Cosmos`.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/879?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Common/MongoProxyClient.ts
+++ b/src/Common/MongoProxyClient.ts
@@ -111,7 +111,7 @@ export function queryDocuments(
           headers: response.headers,
         };
       }
-      errorHandling(response, "querying documents", params);
+      await errorHandling(response, "querying documents", params);
       return undefined;
     });
 }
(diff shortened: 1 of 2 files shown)


southpolesteve (approved):

zfoster (approved):
👌 nice!

PR #918, opened 2021-06-30
Fix enable synapse link error

victor-meng:
Set the `content-type` to `application/json` in the request header when enabling synapse.


src/Utils/arm/request.ts
```diff
@@ -70,6 +73,7 @@ export async function armRequestWithoutPolling<T>({
     method,
     headers: {
       Authorization: userContext.authorizationToken,
+      [HttpHeaders.contentType]: contentType || "text/plain",
```
southpolesteve, line 76: Should we default to `json` content type? Aren't all ARM APIs JSON?

southpolesteve:
I'm very surprised this is needed. Any idea why this isn't a bug for all existing update APIs? Collection/DB create, update, etc. Seems like they are all happy without having the content type

src/Utils/arm/request.ts
southpolesteve: I updated this to default to application/json

PR #925, opened 2021-07-06
Sqlx estimated cost calculation

siddjoshi-ms:

Adding cost estimation for Dedicated Gateway Offering (SQLx)


siddjoshi-ms:
Display cost
image

Update cost on instance count change
image

No screenshot for update cost on sku change as only 1 sku supported right now, but this cost will update whenever sku is changed with the implementation in this PR.

Any error (request error, missing data from fetchPrices API for given regions etc.) leads to default string being shown (default string is what is currently shown)
image

Only number is set when cost is determined to avoid any need for localization in the future.

src/SelfServe/SelfServeTypes.ts
```diff
@@ -236,6 +236,7 @@ export enum DescriptionType {
 export interface Description {
   /**
    * Key used to pickup the string corresponding to the text to be shown as part of the description, from the strings JSON file.
+   * In case key is not present in strings JSON file, displays key as is.
```
srinathnarayanan, line 239: Can you remove this. This is not specific to the Description, it is for any translation key. I will add it elsewhere in the documentation

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
     return { isUpdateInProgress: false, updateInProgressMessageTKey: undefined };
   }
 };
+
+// Data model for Read Regions Query
```
srinathnarayanan, line 132: Remove comments here and everywhere else. I d say add comments only when doing something outside the norm, or if it a todo

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
...
+    if (response.result.location !== undefined) {
+      readRegions.push(response.result.location.replace(" ", "").toLowerCase());
+    } else {
+      for (let i = 0; i < response.result.locations.length; i++) {
+        readRegions.push(response.result.locations[i].locationName.replace(" ", "").toLowerCase());
+      }
+    }
+    return readRegions;
+  } catch (err) {
+    alert(err);
```
srinathnarayanan, line 170: remove alerting. just throw here, the self serve framework catches and displays error as a error message bar on top of the page, below the command bar

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
     return { isUpdateInProgress: false, updateInProgressMessageTKey: undefined };
   }
 };
+
+// Data model for Read Regions Query
+interface ReadRegionsResponse {
```
srinathnarayanan, line 133: move all data models to SqlxTypes file

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
...
+      path: getGeneralPath(userContext.subscriptionId, userContext.resourceGroup, userContext.databaseAccount.name),
+      method: "GET",
+      apiVersion: "2021-04-01-preview",
+    });
+
+    // Parse ARM Response to determine read regions
+    if (response.result.location !== undefined) {
+      readRegions.push(response.result.location.replace(" ", "").toLowerCase());
+    } else {
+      for (let i = 0; i < response.result.locations.length; i++) {
```
srinathnarayanan, line 164: prefer the "for (var location in response.result.locations)" syntax

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
...
+}
+
+// Query ARM Fetch Prices to get pricing
+export const getPriceMap = async (readRegions: Array<string>): Promise<Map<string, Map<string, number>>> => {
+  try {
+    // Initialize empty PriceMap
+    const priceMap = new Map<string, Map<string, number>>();
+
+    // Query ARM
+    for (let i = 0; i < readRegions.length; i++) {
```
srinathnarayanan, line 200: see above comment on syntax of For Loop

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
...
+        queryParams: {
+          filter:
+            "armRegionName eq " +
+            readRegions[i] +
+            " and serviceFamily eq 'Databases' and productName eq 'Azure Cosmos DB Dedicated Gateway - General Purpose'",
+        },
+      });
+
+      // Parse response to create price mapping for current region
+      for (let j = 0; j < response.result.Items.length; j++) {
```
srinathnarayanan, line 218: see above comment on syntax of For Loop

src/SelfServe/SqlX/SqlX.rp.ts
```diff
@@ -128,3 +128,105 @@ export const refreshDedicatedGatewayProvisioning = async (): Promise<RefreshResu
...
+        regionPriceMap.set(response.result.Items[j].skuName, response.result.Items[j].retailPrice);
+      }
+
+      // Insert region's PriceMap to overall PriceMap
+      priceMap.set(readRegions[i], regionPriceMap);
+    }
+
+    return priceMap;
+  } catch (err) {
+    alert(err);
```
srinathnarayanan, line 228: remove alert. refer getDedicatedGatewayResource for throwing on error after logging telemetry.

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -79,6 +86,12 @@ const onNumberOfInstancesChange = (
   } else {
     currentValues.set("warningBanner", undefined);
   }
+
+  // Update cost per hour based on new instance count
```
srinathnarayanan, line 90: Remove comments here and everywhere else

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -177,6 +190,27 @@ const NumberOfInstancesDropdownInfo: Info = {
   },
 };
 
+// Block-scoped Variables to store information for cost calculation
+let priceMap: Map<string, Map<string, number>>;
```
srinathnarayanan, line 194: why declare it here? Declare and use where it is initialized

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -339,7 +377,7 @@ export default class SqlX extends SelfServeBaseClass {
   instances: number;
 
   @Values({
-    labelTKey: "Cost",
+    labelTKey: "Approximate Cost Per Hour",
```
srinathnarayanan, line 380: Add entry for "Approximate Cost Per Hour" in SqlX.json file 

example 
{
...
...
"ApproximateCost": "Approximate Cost Per Hour",
...
...
}

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -177,6 +190,27 @@ const NumberOfInstancesDropdownInfo: Info = {
   },
 };
 
+// Block-scoped Variables to store information for cost calculation
+let priceMap: Map<string, Map<string, number>>;
+let readRegions: Array<string>;
+
+const calculateCost = (skuName: string, instanceCount: number): string | Description => {
```
srinathnarayanan, line 197: this only returns a description right?

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -53,7 +55,12 @@ const CosmosD16s = "Cosmos.D16s";
 
 const onSKUChange = (newValue: InputType, currentValues: Map<string, SmartUiInput>): Map<string, SmartUiInput> => {
   currentValues.set("sku", { value: newValue });
-  currentValues.set("costPerHour", { value: costPerHourValue });
+
+  // Update cost per hour based on new sku
+  currentValues.set("costPerHour", {
+    value: calculateCost(newValue as string, currentValues.get("NumberOfInstances").value as number),
```
srinathnarayanan, line 61: this should be "instances" right?
This scenario is broken :
Account does not have sqlx provisioned, go to Dedicated gateway blade, change the Sku type.

src/SelfServe/SqlX/SqlX.tsx
siddjoshi-ms: Need this to be available outside the function that initializes it, that's why have declared it outside the scope of the initalize function.

siddjoshi-ms:
Info bubble added

image

src/Localization/en/SqlX.json
```diff
@@ -40,7 +40,7 @@
     "CosmosD4Details": "General Purpose Cosmos Compute with 4 vCPUs, 16 GB Memory",
     "CosmosD8Details": "General Purpose Cosmos Compute with 8 vCPUs, 32 GB Memory",
     "CosmosD16Details": "General Purpose Cosmos Compute with 16 vCPUs, 64 GB Memory",
-    "Cost": "Cost",
+    "ApproximateCost": "Approximate Cost Per Hour",
```
timsander1, line 43: Could we mention somewhere that the cost is in USD? Right now, we just have the $ in the cost estimate which could be ambiguous.

siddjoshi-ms: Made it US $, does that work?

timsander1: yeah that works!

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -79,6 +84,11 @@ const onNumberOfInstancesChange = (
   } else {
     currentValues.set("warningBanner", undefined);
   }
+
+  currentValues.set("costPerHour", {
```
srinathnarayanan, line 88: set costPerHour in the onEnableDedicatedGatewayChange function as well, based on whether 'newValue' is true or false. calculate in the true case, hide in false case.

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -79,6 +84,11 @@ const onNumberOfInstancesChange = (
...
+    value: calculateCost(currentValues.get("sku").value as string, newValue as number),
```
srinathnarayanan, line 89: limit to 2 decimals

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -92,7 +102,19 @@ const onEnableDedicatedGatewayChange = (
   if (dedicatedGatewayOriginallyEnabled === newValue) {
     currentValues.set("sku", baselineValues.get("sku"));
     currentValues.set("instances", baselineValues.get("instances"));
-    currentValues.set("costPerHour", baselineValues.get("costPerHour"));
+
+    if (newValue === true) {
```
srinathnarayanan, line 106: there is already a 'if newValue === true' block. Move this there

src/SelfServe/SqlX/SqlX.tsx
```diff
@@ -274,12 +336,16 @@ export default class SqlX extends SelfServeBaseClass {
       hidden: true,
     });
 
+    // Get Read Regions and PriceMap
```
srinathnarayanan, line 339: remove comment

src/SelfServe/SqlX/SqlX.tsx
siddjoshi-ms: the path inside this if statement will not hit the other if newValue===true block, it returns at the end of the if statement.
I can reorganize the code to only check if newValue===true once, but I wasn't sure if that was acceptable, so just made minimal change to existing code.

srinathnarayanan: dedicatedGatewayOriginallyEnabled === newValue ====> set everything to the baseline. it is already computed on init. Set costPerHour to baseline as well. 

This newValue ===true logic is irrespective of whether dedicatedGatewayOriginallyEnabled === newValue or not. So move it to that section

srinathnarayanan (approved):

southpolesteve (approved):

PR #928, opened 2021-07-09
UX Changes for Default Partition Key A/B Experiment

t-tarabhatia:
This PR request is for the UX changes for the default partition key experiment. This A/B test will only affect SQL and Mongo DB API users.

The control group will be left as is, where the partition key input field will have a placeholder example like "e.g. /address/zipCode".

The treatment group will have /id prefilled as the partition key as well as subtext, indicating that /id is a suitable partition key for small workloads. Users will be able to change the default /id input to something else if they'd like during the create collection process. 

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/928)


src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -122,9 +122,33 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
+        enableDedicatedThroughput: false,
+        createMongoWildCardIndex: isCapabilityEnabled("EnableMongo"),
+        useHashV2: false,
+        enableAnalyticalStore: false,
+        uniqueKeys: [],
+        errorMessage: "",
+        showErrorDetails: false,
+        isExecuting: false,
+      }
+    }
```
southpolesteve, line 147: You probably don't need to replace this entire object. You can just set the specific property you want to change:

`this.state. partitionKey = "something"`

southpolesteve (changes requested):
Thanks @t-tarabhatia! I made one comment but there are two other issues that need to be resolved to merge this:

- **linting**: run `npm run lint` locally and fix any errors. Warnings do not need to be fixed
- **formatting**: run `npm run format`. This will automatically format your code

If you want help with either issue please ping me on teams and we can pair. I also highly recommend installing the vscode prettier and eslint extension which will surface these issues in vscode.

southpolesteve (approved):

PR #929, opened 2021-07-09
Use window messaging to pass sensitive data to terminal iframe

gitanuj:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/929)

Currently we're passing everything needed by terminal iframe using query params. This includes notebooks server token too. We should avoid passing tokens in URL query params because URLs can be logged/stored elsewhere.

This change uses postRobot window messaging to pass data instead.

Verfied that terminal, mongo shell and cassandra shell all work fine.


src/Explorer/Controls/Notebook/NotebookTerminalComponent.tsx
```diff
@@ -2,44 +2,64 @@
...
+    this.sendPropsToTerminalFrame();
+  }
+
   public render(): JSX.Element {
     return (
       <div className="notebookTerminalContainer">
         <iframe
           title="Terminal to Notebook Server"
-          src={NotebookTerminalComponent.createNotebookAppSrc(this.props.notebookServerInfo, this.getTerminalParams())}
+          onLoad={(event) => this.handleFrameLoad(event)}
```
vignesh-mr, line 33: Might want to register the callback as a private var instead, else it creates a new lambda on every reload.

vignesh-mr (approved):

src/Explorer/Controls/Notebook/NotebookTerminalComponent.test.tsx
```diff
@@ -1,154 +0,0 @@
-import * as DataModels from "../../../Contracts/DataModels";
```
southpolesteve, line 1: Any reason for deleting the whole test file? Would be nice to have at least one test for this component to know it renders.

src/Explorer/Controls/Notebook/NotebookTerminalComponent.tsx
```diff
@@ -2,44 +2,60 @@
...
+  }
+
   public render(): JSX.Element {
     return (
       <div className="notebookTerminalContainer">
-        <iframe
-          title="Terminal to Notebook Server"
-          src={NotebookTerminalComponent.createNotebookAppSrc(this.props.notebookServerInfo, this.getTerminalParams())}
-        />
+        <iframe title="Terminal to Notebook Server" onLoad={this.handleFrameLoad.bind(this)} src="terminal.html" />
```
southpolesteve, line 31: Prob out of scope for this PR, but why is this an `iframe` at all? Can't we load the component async like we do for other notebooks and Monaco? `await import(./notebookTerminal)`

southpolesteve (approved):

src/Explorer/Controls/Notebook/NotebookTerminalComponent.tsx
southpolesteve: FYI there is no performance benefit to doing as changed with `bind()`. It still allocates a new function.

In any case JS functions are extremely cheap and fast to create. Any performance that might be gained here isn't enough to matter.

src/Explorer/Controls/Notebook/NotebookTerminalComponent.tsx
gitanuj: @languy probably has more context on this. IMO as long as we lazy load JupyterLab we should be okay not using the `iframe`.

src/Explorer/Controls/Notebook/NotebookTerminalComponent.test.tsx
gitanuj: Earlier we used query params to pass properties to the terminal `iframe`. The tests made sure we correctly passed the params. However now we use typescript's strict compile time checks to make sure correct props are passed as window message so we don't need those tests. I'll push an update to include a basic snapshot render test.

PR #954, opened 2021-07-22
Add changes to Partition Key A/B Test

t-tarabhatia:
This PR adds a flight to the default partition key test. The new flight prefills the partition key box with pk instead of id. This is to study which prefilled option is preferred.


src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -815,10 +811,28 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
+    if (userContext.features.partitionKeyDefault2 && userContext.apiType === "SQL") {
+      const key = "/pk";
+      return key;
+    } else if (userContext.features.partitionKeyDefault2 && userContext.apiType === "Mongo") {
+      const key = "pk";
+      return key;
+    }
+    return "";
+  }
+
```
victor-meng, line 831:
```suggestion
  private getPartitionKey(): string {
	if (userContext.apiType !== "SQL" && userContext.apiType !== "Mongo") {
		return "";
	}

	if (userContext.features.partitionKeyDefault) {
		return userContext.apiType === "SQL" ? "/id" : "_id";
	}

	if (userContext.features.partitionKeyDefault2) {
		return userContext.apiType === "SQL" ? "/pk" : "pk";
	}

	return "";
  }

```

src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -815,10 +811,28 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
   private getPartitionKeySubtext(): string {
     if (
-      userContext.features.partitionKeyDefault &&
-      (userContext.apiType === "SQL" || userContext.apiType === "Mongo")
+      (userContext.features.partitionKeyDefault && userContext.apiType === "SQL") ||
+      (userContext.features.partitionKeyDefault && userContext.apiType === "Mongo")
```
victor-meng, line 835: This logic is exactly the same as the previous one.

victor-meng (approved):

PR #956, opened 2021-07-22
Migrate resource tree for resource token to react

victor-meng:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/956?feature.someFeatureFlagYouMightNeed=true)


config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "GITHUB_CLIENT_ID": "167ea4b09801db1de03d",
+  "GITHUB_CLIENT_SECRET": "e1208e29f4e03c5f9c6fc08a36375df8344ab01d"
+}
```
southpolesteve, line 4: Did you mean to check this in?

less/tree.less
```diff
@@ -1,272 +1,271 @@
 @import "./Common/Constants";
 
-
 .resourceTree {
+  height: 100%;
+  flex: 0 0 auto;
+  width: 20%;
```
southpolesteve, line 6: Verify that the collapsing the resource tree still works. I know this is something we broke earlier. The tabs wouldn't fill the space.

config.json
victor-meng: Nop! This is the config for testing github connection. I'll remove it. Thanks for catching this!

southpolesteve (approved):

less/tree.less
victor-meng: This will indeed break collapsing the resource tree. I've reverted the change.

PR #1005, opened 2021-08-30
Self serve for GraphApiCompute

MehaKaushik:


MehaKaushik:
Copied from https://github.com/Azure/cosmos-explorer/pull/985/files

src/Localization/en/GraphAPICompute.json
```diff
@@ -0,0 +1,57 @@
+{
+    "GraphAPIDescription": "Provision a Graph GraphAPI Compute for your Azure Cosmos DB account.",
```
srinathnarayanan, line 2: add "DedicatedGateway": "Dedicated Gateway"

src/SelfServe/GraphAPICompute/GraphAPICompute.tsx
```diff
@@ -0,0 +1,411 @@
...
+    defaults.set("connectionString", undefined);
+    defaults.set("metricsString", {
+      value: undefined,
+      hidden: true,
+    });
+
+    regions = await getReadRegions();
+    priceMap = await getPriceMap(regions);
+    const response = await getCurrentProvisioningState();
+    if (response.status && response.status !== "Deleting") {
```
srinathnarayanan, line 330: you should probably handle "Deleting", "Creating" and "Created" cases separately.
the Connection string and monitor usage sections should show up only if its "Created" and not in "Creating" state. Its ok to show it in "Deleting" state though (because that info is already shown to the user)

src/Localization/en/GraphAPICompute.json
MehaKaushik: dont need it anywhere

src/SelfServe/GraphAPICompute/GraphAPICompute.tsx
MehaKaushik: Done

srinathnarayanan (approved):

PR #1070, opened 2021-09-11
Fix throughput cost estimate in add collection panel

victor-meng:
Fixed incorrect cost estimate formula used in add collection panel:
- For multi-master, we should multiply the price by the number of regions instead of number of regions + 1
- Removed the multi-master multiplier from the formula. The base price for autoscale throughput already takes into account the multiplier. I added the multiplier into the base price for manual throughput.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1070?feature.someFeatureFlagYouMightNeed=true)


deborahc:
Thanks Victor! Can you take a look at:
- In create container/db blade, the tooltip content is not showing up
image
- In Scale and Settings, for a multi-region write (multi-master) account with >1 regions, the unit price for RU should be 0.00016.

image

deborahc:
Thanks Victor! Looks good to me.

deborahc (approved):

PR #1117, opened 2021-10-05
Fix Unit Test: Mock the class to its instance 

kcheekuri:
Download from gallery pop-up expects explorer.tsx class instance but a concrete object was passed. 

Fixed this.


notlaforge (approved):

notlaforge:
I'm surprised this works, but it does!

gitanuj (dismissed):

gitanuj (approved):

PR #1145, opened 2021-10-21
Setting isNotebooKEnabled to true by default

srinathnarayanan:
the isNotebookEnabled state was earlier being set based on an RP API that looked for the notebook workspace to see if it had been provisioned or not. Now that the resource is removed, this always returns false. Since we want to continue to show the banner that says "notebooks is temporarily down" and the diabled buttons in the command bar, we are setting isNotebookEnabled to true by default.

Future work : remove all references to isNotebookEnabled and isNotebooksEnabledForAccount as part of c;eaning up compute v2 code.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1145)

### DIFF
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -1,4 +1,5 @@
 import { Link } from "@fluentui/react/lib/Link";
+import { isPublicInternetAccessAllowed } from "Common/DatabaseAccountUtility";
 import * as ko from "knockout";
 import React from "react";
 import _ from "underscore";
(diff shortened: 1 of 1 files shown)


src/Explorer/Explorer.tsx
```diff
@@ -1232,13 +1232,9 @@ export default class Explorer {
...
-    let isNotebookEnabled = true;
-    if (!userContext.features.phoenix) {
-      isNotebookEnabled =
-        userContext.authType !== AuthType.ResourceToken &&
-        ((await this._containsDefaultNotebookWorkspace(userContext.databaseAccount)) ||
-          userContext.features.enableNotebooks);
-    }
+
+    // TODO: remove reference to isNotebookEnabled and isNotebooksEnabledForAccount
+    const isNotebookEnabled = true;
```
swviswan, line 1237: >const isNotebookEnabled = true; [](http://example.com/codeflow?start=4&length=31)

Not putting it behind a flight?

PR #1168, opened 2021-11-30
Added integration for new github client for test env

srinathnarayanan:
This PR
- Adds a new config context for github test env client id
- Adds logic to pick this client id to send for github OAuth call and to junoin case juno endpoint used is test/test2/test3

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1168)

### DIFF
--- a/src/Utils/GitHubUtils.ts
+++ b/src/Utils/GitHubUtils.ts
@@ -1,4 +1,9 @@
 // https://github.com/<owner>/<repo>/tree/<branch>
+
+import { JunoEndpoints } from "Common/Constants";
+import { configContext } from "ConfigContext";
+import { userContext } from "UserContext";
+
 // The url when users visit a repo/branch on github.com
 export const RepoUriPattern = /https:\/\/github.com\/([^/]*)\/([^/]*)\/tree\/([^?]*)/;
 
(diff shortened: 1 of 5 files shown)


src/Utils/GitHubUtils.ts
```diff
@@ -60,3 +65,15 @@ export function toContentUri(owner: string, repo: string, branch: string, path:
 export function toRawContentUri(owner: string, repo: string, branch: string, path: string): string {
   return `https://raw.githubusercontent.com/${owner}/${repo}/${branch}/${path}`;
 }
+
+export function GetGithubClientId(): string {
+  const junoEndpoint = userContext.features.junoEndpoint ?? configContext.JUNO_ENDPOINT;
```
gitanuj, line 70: Is this needed? I thought we already update the values in configContext somewhere whenever feature flags are used. Can you confirm?

gitanuj (approved):

src/Utils/GitHubUtils.ts
srinathnarayanan: we dont update config context itself, we do the exact same thing in JunoClient's getJunoEndpoint

gitanuj (approved):

PR #1187, opened 2022-01-11
FIxed bugs in JupyterLabAppFactory

srinathnarayanan:
Fixed bug with logic to close the terminal tab once the shell is closed. If the shell hung, the earlier logic would not work. With these changes, we instead check for the opening of the shell instead of the closure, and then infer that the shell is closed if we see a new terminal line.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1187)

### DIFF
--- a/src/Terminal/JupyterLabAppFactory.ts
+++ b/src/Terminal/JupyterLabAppFactory.ts
@@ -8,21 +8,33 @@ import { Panel, Widget } from "@phosphor/widgets";
 import { userContext } from "UserContext";
 
 export class JupyterLabAppFactory {
-  private isShellClosed: boolean;
+  private isShellStarted: boolean;
   private onShellExited: () => void;
-  private checkShellClosed: ((content: string | undefined) => boolean | undefined) | undefined;
+  private checkShellStarted: ((content: string | undefined) => void) | undefined;
+
+  private isShellExited(content: string | undefined) {
+    return content?.includes("cosmosuser@");
+  }
+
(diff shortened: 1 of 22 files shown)


src/Terminal/JupyterLabAppFactory.ts
```diff
@@ -34,11 +46,12 @@ export class JupyterLabAppFactory {
     const session = await manager.startNew();
     session.messageReceived.connect(async (_, message: IMessage) => {
       const content = message.content && message.content[0]?.toString();
-      if (this.checkShellClosed && message.type == "stdout") {
+      console.log(content);
```
languy, line 49: Remove

languy (dismissed):
Does it work as expected if shell takes a while to start or you need a bit of polling before deciding it's not started?

src/Terminal/JupyterLabAppFactory.ts
srinathnarayanan: so the new terminal line (cosmosuser@...) is not seen once the shell is launched. If is only seen next after it shuts down or crashes in between.

srinathnarayanan:
> Does it work as expected if shell takes a while to start or you need a bit of polling before deciding it's not started?

so the new terminal line (cosmosuser@...) is not seen once the shell is launched. It is only seen next after it shuts down or crashes in between.

languy (approved):

balalakshmin (approved):

PR #1196, opened 2022-01-21
Validate endpoints from feature flags

ato9000:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1196)


src/Utils/EndpointValidation.ts
```diff
@@ -0,0 +1,82 @@
...
+    `^https:\\/\\/[\\.\\w]*portal\\.azure\\.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*portal\\.microsoftazure.de$`,
+    `^https:\\/\\/[\\.\\w]*ext\\.azure\\.(com|cn|us)$`,
+    `^https:\\/\\/[\\.\\w]*\\.ext\\.microsoftazure\\.de$`,
+    `^https://cosmos-db-dataexplorer-germanycentral.azurewebsites.de$`,
+];
+
+export const allowedJunoOrigins: ReadonlyArray<string> = [
+    "https://juno-test.documents-dev.windows-int.net",
+    "https://juno-test2.documents-dev.windows-int.net",
```
gitanuj, line 31: Endpoint for juno test3 is missing. It should include `https://juno-test3.documents-dev.windows-int.net`

gitanuj: There is a JunoEndpoints constants file from where we can use these endpoints too.

src/Utils/EndpointValidation.ts
```diff
@@ -0,0 +1,40 @@
+export function validateEndpoint(endpointToValidate: string | undefined, allowedEndpoints: string[]): boolean {
+  if (!endpointToValidate) {
+    return true;
+  }
+  const originToValidate: string = new URL(endpointToValidate).origin;
+  const allowedOrigins: string[] = allowedEndpoints.map((allowedEndpoint) => new URL(allowedEndpoint).origin) || [];
```
languy, line 6: Consider optimizing: if `allowedEndpoint` is static, the origin could be computed once. (as in: `const allowedOrigins = ...`).
This said, it's probably not going to matter much either way.

languy (dismissed):

asierisayas (approved):

gitanuj (dismissed):

src/Utils/EndpointValidation.ts
ato9000: Let's discuss offline. I will leave as is for now.

asierisayas (dismissed):

src/Utils/EndpointValidation.ts
```diff
@@ -0,0 +1,64 @@
+import * as Logger from "../Common/Logger";
+
+export function validateEndpoint(endpointToValidate: string | undefined, allowedEndpoints: string[]): boolean {
```
victor-meng, line 3: Can the `allowedEndpoints` be `ReadyonlyArray<string> | string[]` so we don't have to convert the ReadonlyArrays to a regular array?

src/Explorer/Explorer.tsx
```diff
@@ -422,7 +417,13 @@ export default class Explorer {
     connectionStatus.status = ConnectionStatusType.Connected;
     useNotebook.getState().setConnectionInfo(connectionStatus);
     useNotebook.getState().setNotebookServerInfo({
-      notebookServerEndpoint: userContext.features.notebookServerUrl || connectionInfo.data.notebookServerUrl,
+      notebookServerEndpoint:
+        (validateEndpoint(
+          userContext.features.notebookServerUrl,
+          allowedNotebookServerUrls.map((endpoint) => endpoint)
```
victor-meng, line 423: Is there anyway we can avoid doing this conversion? Also we can use the spread operator to do this if we have to: `[...allowedNotebookServerUrls]`

victor-meng (approved):

gitanuj (approved):

victor-meng (dismissed):

PR #1229, opened 2022-02-24
Update autoscale throughput limits

victor-meng:
- changed minimum autoscale throughput value from 4000 to 1000
- changed default autoscale throughput value for free tier account from 4000 to 1000

### DIFF
--- a/src/Utils/AutoPilotUtils.ts
+++ b/src/Utils/AutoPilotUtils.ts
@@ -1,6 +1,6 @@
-export const minAutoPilotThroughput = 4000;
-
+export const minAutoPilotThroughput = 1000;
 export const autoPilotIncrementStep = 1000;
+export const autoPilotDefaultThroughput = 4000;
 
 export function isValidAutoPilotThroughput(maxThroughput: number): boolean {
   if (!maxThroughput) {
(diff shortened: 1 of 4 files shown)


senthamilsindhu (dismissed):

vchske (dismissed):

src/Utils/AutoPilotUtils.ts
```diff
@@ -1,6 +1,6 @@
-export const minAutoPilotThroughput = 4000;
-
+export const minAutoPilotThroughput = 1000;
```
ato9000, line 1: please rename to autoPilotThroughput1K

src/Utils/AutoPilotUtils.ts
```diff
@@ -1,6 +1,6 @@
...
 export const autoPilotIncrementStep = 1000;
+export const autoPilotDefaultThroughput = 4000;
```
ato9000, line 3: please rename to autoPilotThroughput4K

src/Utils/AutoPilotUtils.ts
victor-meng: I think we should keep the current name for this one since it represents the minimum throughput value for autoscale. If we rename it to autoPilotThroughput1K then it loses the minimum part which is important for this number.

src/Utils/AutoPilotUtils.ts
victor-meng: Done

PR #1262, opened 2022-05-02
Added publicgallery flight

srinathnarayanan:
This PR adds a flight to hide/show the public samples of the gallery
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1262)


asierisayas (dismissed):

victor-meng (approved):

src/Explorer/Controls/NotebookGallery/__snapshots__/GalleryViewerComponent.test.tsx.snap
```diff
@@ -8,95 +8,6 @@ exports[`GalleryViewerComponent renders 1`] = `
     onLinkClick={[Function]}
     selectedKey="OfficialSamples"
   >
-    <PivotItem
```
Mathos1432, line 11: Does this mean we're no longer testing the public gallery because it's behind a feature flag? If so, is there a way to set the feature flag for the test, or is this not worth testing?

srinathnarayanan: these snapshots are testing only the load state of the component. We dont have the whole component available now dude to the feature flags, so this is fine.

src/Explorer/Controls/NotebookGallery/GalleryViewerComponent.tsx
```diff
@@ -148,14 +149,17 @@ export class GalleryViewerComponent extends React.Component<GalleryViewerCompone
...
-    const tabs: GalleryTabInfo[] = [
-      this.createPublicGalleryTab(
-        GalleryTab.PublicGallery,
-        this.state.publicNotebooks,
-        this.state.isCodeOfConductAccepted
-      ),
-      this.createSamplesTab(GalleryTab.OfficialSamples, this.state.sampleNotebooks),
-    ];
+    const tabs: GalleryTabInfo[] = [];
+    if (userContext.features.publicGallery) {
```
gitanuj, line 153: Should we also hide Published Notebooks tab?

src/Platform/Hosted/extractFeatures.ts
```diff
@@ -37,6 +37,7 @@ export type Features = {
   phoenixNotebooks?: boolean;
   phoenixFeatures?: boolean;
   notebooksDownBanner: boolean;
+  publicGallery?: boolean;
```
gitanuj, line 40: Do you need to add support for this flag in the method below? Otherwise this can't be used as feature flag

gitanuj (approved):

src/Platform/Hosted/extractFeatures.ts
srinathnarayanan: No this is set through flighting, we dont want to expose this through feature flags otherwise it could be misused

asierisayas (approved):

PR #1263, opened 2022-05-03
Rename properties from notebookService to phoenixService in phoenix c…

Mathos1432:
# Why is this change being made?
As we move to host more services on phoenix, we need naming to be more generic. A change was made to rename most code in control plane to phoenix* instead of notebook* so now we need to update data explorer to use the new property before we remove the old one.

# What changed?
- Updated the communication with phoenix to use the updated phoenix properties instead of notebooks

# How was this validated?
- [x] Removed all the properties marked as obsolete in ControlPlane. Ran ControlPlane, ContainerGateway, Router, DataExplorer and a Notebooks container locally and validated that the Getting started with .net notebook worked.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1263?feature.phoenixnotebooks=true&feature.phoenixfeatures=true)


src/Contracts/DataModels.ts
```diff
@@ -433,7 +433,7 @@ export interface NotebookWorkspaceConnectionInfo {
 
 export interface ContainerInfo {
   durationLeftInMinutes: number;
-  notebookServerInfo: NotebookWorkspaceConnectionInfo;
```
srinathnarayanan, line 436: similar to controlplane changes, should we not have both versions here?
If this gets deployed to MPAC and we dont have the changes on stage, it will break.

src/Contracts/DataModels.ts
```diff
@@ -469,8 +469,16 @@ export interface IMaxUsersPerDbAccountExceeded extends IValidationError {
 }
 
 export interface IPhoenixConnectionInfoResult {
+  /**
```
srinathnarayanan, line 472: we can just remove them right? there is no backward compat issue in case its removed.

src/Explorer/Explorer.tsx
```diff
@@ -372,8 +372,8 @@ export default class Explorer {
         if (connectionInfo.status !== HttpStatusCodes.OK) {
           throw new Error(`Received status code: ${connectionInfo?.status}`);
         }
-        if (!connectionInfo?.data?.notebookServerUrl) {
-          throw new Error(`NotebookServerUrl is invalid!`);
+        if (!connectionInfo?.data?.notebookServerUrl && !connectionInfo?.data?.phoenixServiceUrl) {
```
srinathnarayanan, line 375: lets have dependency only on phoenix server url and remove the notebookserverurl

src/Explorer/Explorer.tsx
```diff
@@ -370,8 +370,11 @@ export default class Explorer {
         });
         useNotebook.getState().setIsAllocating(true);
         connectionInfo = await this.phoenixClient.allocateContainer(provisionData);
-        if (!connectionInfo?.data?.notebookServerUrl) {
-          throw new Error(`NotebookServerUrl is invalid!`);
+        if (connectionInfo.status !== HttpStatusCodes.OK) {
```
srinathnarayanan, line 373: this check is not needed as it is handled in the phoenix client. This wont be hit

kcheekuri (approved):

srinathnarayanan (approved):

src/Contracts/DataModels.ts
```diff
@@ -438,7 +438,7 @@ export interface NotebookWorkspaceConnectionInfo {
 
 export interface ContainerInfo {
   durationLeftInMinutes: number;
-  notebookServerInfo: NotebookWorkspaceConnectionInfo;
+  phoenixServerInfo: NotebookWorkspaceConnectionInfo;
```
gitanuj, line 441: Should we rename `NotebookWorkspaceConnectionInfo` too?

gitanuj (approved):

PR #1282, opened 2022-05-25
Fix Official Samples not loading when Gallery tab is opened

gitanuj:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1282)

### DIFF
--- a/src/GalleryViewer/GalleryViewer.tsx
+++ b/src/GalleryViewer/GalleryViewer.tsx
@@ -2,6 +2,7 @@ import { initializeIcons, Link, Text } from "@fluentui/react";
 import "bootstrap/dist/css/bootstrap.css";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
+import { userContext } from "UserContext";
 import { initializeConfiguration } from "../ConfigContext";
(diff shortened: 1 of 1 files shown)


src/GalleryViewer/GalleryViewer.tsx
```diff
@@ -25,7 +25,7 @@ const onInit = async () => {
 
   const props: GalleryAndNotebookViewerComponentProps = {
     junoClient: new JunoClient(),
-    selectedTab: galleryViewerProps.selectedTab || GalleryTab.PublicGallery,
+    selectedTab: galleryViewerProps.selectedTab || GalleryTab.OfficialSamples,
```
srinathnarayanan, line 28: Is this change needed cos public gallery is hidden currently? since public gallery is behind a flight, should we check for the same flight here to choose between public gallery/official samples?

gitanuj: Good point. I'll add a feature flag check

srinathnarayanan (approved):

victor-meng (approved):

PR #1290, opened 2022-06-07
Update allocate call

kcheekuri:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1290)

This PR:

-Updates the phoenix allocate call with poolId as default

### DIFF
--- a/src/Common/Constants.ts
+++ b/src/Common/Constants.ts
@@ -354,6 +354,10 @@ export enum ContainerStatusType {
   Disconnected = "Disconnected",
 }
 
+export enum PoolIdType {
+  DefaultPoolId = "default",
+}
+
 export const EmulatorMasterKey =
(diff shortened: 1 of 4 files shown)


gitanuj (approved):

src/Common/Constants.ts
```diff
@@ -354,6 +354,10 @@ export enum ContainerStatusType {
   Disconnected = "Disconnected",
 }
 
+export enum PoolIdType {
```
papatid, line 357: Shouldn't this be PoolIds?

papatid (approved):

src/Common/Constants.ts
kcheekuri: For now we are not sure if we want to have per feature a poolid or multiple poolid's per feature. So going with PoolIdType

PR #1295, opened 2022-06-27
Adding chatbot to data explorer

balalakshmin:
This PR:

- Adds chatbot component in data explorer
- Added the feature behind a flag
- New library botframework-webchat added

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1268?feature.enablechatbot=true)


src/Chatbot/ChatbotClient.ts
```diff
@@ -0,0 +1,26 @@
+import { HttpHeaders } from "../Common/Constants";
+import { configContext } from "../ConfigContext";
+import { userContext } from "../UserContext";
+import { getAuthorizationHeader } from "../Utils/AuthorizationUtils";
+
+export class ChatbotClient {
+  public async getConversationToken(): Promise<{ conversationId: string; token: string; expires_in: number }> {
```
gitanuj, line 7: You can define this type and use that instead

src/Chatbot/ChatbotClient.ts
```diff
@@ -0,0 +1,26 @@
...
+    const url = `${configContext.JUNO_ENDPOINT}/api/chatbot/bot${userContext.databaseAccount.id}/conversationToken`;
```
gitanuj, line 8: Is the path correct?

src/Explorer/Explorer.tsx
```diff
@@ -85,6 +86,12 @@ export default class Explorer {
   // Notebooks
   public notebookManager?: NotebookManager;
 
+  public conversationToken: ko.Observable<string>;
```
gitanuj, line 89: I think we were trying to migrate DE away from knockout. Any reason why we would want these properties to be ko?

src/Explorer/Controls/CommandButton/CommandButtonComponent.tsx
```diff
@@ -259,7 +260,18 @@ export class CommandButtonComponent extends React.Component<CommandButtonCompone
...
           <div className={contentClassName}>
-            <img className="commandIcon" src={this.props.iconSrc} alt={this.props.iconAlt} />
+            if (this.props.iconName){" "}
+            {
+              <div>
+                <Icon
+                  styles={{ root: { marginBottom: -3 } }}
+                  className="panelInfoIcon"
+                  iconName={this.props.iconName}
+                  ariaLabel="ChatBot"
```
gitanuj, line 270: `CommandButtonComponent` is a generic button class used for all kinds of buttons so having special handling for chatbot button seems odd. Can we generalize this instead?

src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.tsx
```diff
@@ -26,6 +26,7 @@ import { userContext } from "../../../UserContext";
 import { getCollectionName, getDatabaseName } from "../../../Utils/APITypeUtils";
 import { isRunningOnNationalCloud } from "../../../Utils/CloudUtils";
 import { CommandButtonComponentProps } from "../../Controls/CommandButton/CommandButtonComponent";
+import { SupportPaneComponent } from "../../Controls/SupportPaneComponent/SupportPaneComponent";
```
gitanuj, line 29: We should lazy import this component to make loading of DE as lightweight as possible in case Chatbot isn't available.

gitanuj (changes requested):

src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.tsx
gitanuj: Also we should explore loading chatbot component only when the button is clicked.

src/Chatbot/ChatbotClient.ts
balalakshmin: Yes

PR #1314, opened 2022-08-08
Hide connect button incase Phoenix Notebooks flight is false

kcheekuri:
Hide connect button id flight notebooks is false and show connect button status only if phoenix features are used"

### DIFF
--- a/src/Explorer/Notebook/useNotebook.ts
+++ b/src/Explorer/Notebook/useNotebook.ts
@@ -219,10 +219,10 @@ export const useNotebook: UseStore<NotebookState> = create((set, get) => ({
     };
     const gitHubNotebooksContentRoot = notebookManager?.gitHubOAuthService?.isLoggedIn()
       ? {
-          name: "GitHub repos",
-          path: "PsuedoDir",
-          type: NotebookContentItemType.Directory,
-        }
+        name: "GitHub repos",
+        path: "PsuedoDir",
+        type: NotebookContentItemType.Directory,
+      }
       : undefined;
 
     set({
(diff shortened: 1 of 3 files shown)


src/Explorer/Notebook/useNotebook.ts
```diff
@@ -312,8 +312,8 @@ export const useNotebook: UseStore<NotebookState> = create((set, get) => ({
 
       if (dbAccountAllowedInfo.status === HttpStatusCodes.OK) {
         if (dbAccountAllowedInfo?.type === PhoenixErrorType.PhoenixFlightFallback) {
-          isPhoenixNotebooks = isPublicInternetAllowed && userContext.features.phoenixNotebooks;
-          isPhoenixFeatures = isPublicInternetAllowed && userContext.features.phoenixFeatures;
+          isPhoenixNotebooks = isPublicInternetAllowed && userContext.features.phoenixNotebooks === true;
+          isPhoenixFeatures = isPublicInternetAllowed && userContext.features.phoenixFeatures === true;
```
Mathos1432, line 316: It's been a while since I've done some typescript, Is this additional check to make sure that phoenixNotebooks and phoenixFeatures is assigned a bool and not some other value that would also return true when cast to a bool?

kcheekuri: Yes, in case of "userContext.features.phoenixNotebooks" undefined, "isPhoenixNotebooks" also becomes undefined. This will eliminate the case.

Mathos1432 (dismissed):

gitanuj (approved):

Mathos1432 (approved):

PR #1318, opened 2022-08-22
Hide connect btn and only show when phoenix notebooks or features are used

kcheekuri:
Hide connect btn and only show when phoenix notebooks or features are used


gitanuj (approved):
Could there be scenarios where a user has terminal/notebook tab opened and container gets disconnected (due to session timeout or some other error)? Do we show connect button in that case?

kcheekuri:
> Could there be scenarios where a user has terminal/notebook tab opened and container gets disconnected (due to session timeout or some other error)? Do we show connect button in that case?



> Could there be scenarios where a user has terminal/notebook tab opened and container gets disconnected (due to session timeout or some other error)? Do we show connect button in that case?

Yes, We show either Reconnect button or Connection Failed. And this is expected as the user started using notebooks

srinathnarayanan (approved):

victor-meng (approved):

PR #1342, opened 2022-10-13
Add retries when checking if account is allowed to access phoenix

victor-meng:
When a free tier account is just created, it's possible that the permissions have not been propogated, causing the `getDbAccountAllowedStatus` call to fail. This fix adds a retry logic to the request with a maximum retry count of 10 and a 5 seconds retry interval.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1342?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Notebook/useNotebook.ts
+++ b/src/Explorer/Notebook/useNotebook.ts
@@ -1,6 +1,5 @@
 import { isPublicInternetAccessAllowed } from "Common/DatabaseAccountUtility";
 import { cloneDeep } from "lodash";
-import promiseRetry from "p-retry";
 import { PhoenixClient } from "Phoenix/PhoenixClient";
 import create, { UseStore } from "zustand";
 import { AuthType } from "../../AuthType";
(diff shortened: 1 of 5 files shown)


ato9000 (approved):

jawelton74 (dismissed):

src/Explorer/Notebook/useNotebook.ts
```diff
@@ -309,7 +310,11 @@ export const useNotebook: UseStore<NotebookState> = create((set, get) => ({
 
       const isPublicInternetAllowed = isPublicInternetAccessAllowed();
       const phoenixClient = new PhoenixClient(userContext?.databaseAccount?.id);
-      const dbAccountAllowedInfo = await phoenixClient.getDbAccountAllowedStatus();
+      const dbAccountAllowedInfo = await promiseRetry(() => phoenixClient.getDbAccountAllowedStatus(), {
+        retries: 10,
```
kcheekuri, line 314: Any reason for having 10 retries?

kcheekuri: Also how do we differentiate between 403 because of ARM Call or 403 because of not having ListKey Permissions/RegionNotServicable which are valid?

kcheekuri: May be you could exclude the known cases and abort the retry in those cases? Like if 403 is because of 
1)RegionNotServicable
2)SubscriptionNotAllowed
 You can do something like [throw new AbortError(error)], other cases you may retry?

kcheekuri (changes requested):

src/Explorer/Notebook/useNotebook.ts
victor-meng: Discussed offline with Karthik

kcheekuri (approved):

jawelton74 (approved):

vchske (approved):

PR #1353, opened 2022-11-30
Users/jawelton/support new auth tokens

jawelton74:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1353?feature.someFeatureFlagYouMightNeed=true)

This change introduces a new auth token format (primary & secondary tokens to support secret rotation) side-by-side with
the current format. We will deprecate the current format in a subsequent change to be coordinated with support in portal backend.

It also refactors some code where we generate request headers based on authentication.

### DIFF
--- a/src/Explorer/Tables/TableDataClient.ts
+++ b/src/Explorer/Tables/TableDataClient.ts
@@ -12,7 +12,7 @@ import * as HeadersUtility from "../../Common/HeadersUtility";
 import { configContext } from "../../ConfigContext";
 import * as ViewModels from "../../Contracts/ViewModels";
 import { userContext } from "../../UserContext";
-import { getAuthorizationHeader } from "../../Utils/AuthorizationUtils";
+import { getAuthorizationHeaders } from "../../Utils/AuthorizationUtils";
 import * as NotificationConsoleUtils from "../../Utils/NotificationConsoleUtils";
 import { logConsoleInfo, logConsoleProgress } from "../../Utils/NotificationConsoleUtils";
 import Explorer from "../Explorer";
@@ -523,8 +523,10 @@ export class CassandraAPIDataClient extends TableDataClient {
   }
 
   private setAuthorizationHeader: (xhr: XMLHttpRequest) => boolean = (xhr: XMLHttpRequest): boolean => {
-    const authorizationHeaderMetadata: ViewModels.AuthorizationTokenHeaderMetadata = getAuthorizationHeader();
-    xhr.setRequestHeader(authorizationHeaderMetadata.header, authorizationHeaderMetadata.token);
+    const authorizationHeaderMetadata: Headers = getAuthorizationHeaders();
+    authorizationHeaderMetadata.forEach((value, key) => {
+      xhr.setRequestHeader(key, value);
+    });
 
     return true;
   };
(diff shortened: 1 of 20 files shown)


src/Explorer/Tables/TableDataClient.ts
```diff
@@ -523,8 +523,10 @@ export class CassandraAPIDataClient extends TableDataClient {
   }
 
   private setAuthorizationHeader: (xhr: XMLHttpRequest) => boolean = (xhr: XMLHttpRequest): boolean => {
-    const authorizationHeaderMetadata: ViewModels.AuthorizationTokenHeaderMetadata = getAuthorizationHeader();
-    xhr.setRequestHeader(authorizationHeaderMetadata.header, authorizationHeaderMetadata.token);
+    const authorizationHeaderMetadata = getAuthorizationHeaders();
+    Object.entries(authorizationHeaderMetadata).forEach((value, index, array) => {
```
victor-meng, line 527: It looks like we are not using `index` and `array` so we should just remove them from the parameters.

src/Utils/AuthorizationUtils.ts
```diff
@@ -1,21 +1,27 @@
...
-export function getAuthorizationHeader(): ViewModels.AuthorizationTokenHeaderMetadata {
+export function getAuthorizationHeaders() {
   if (userContext.authType === AuthType.EncryptedToken) {
-    return {
-      header: Constants.HttpHeaders.guestAccessToken,
-      token: userContext.accessToken,
-    };
+    return userContext.accessToken ? getAccessTokenAuthorizationHeaders(userContext.accessToken) : {};
+  } else {
+    return { [Constants.HttpHeaders.authorization]: userContext.authorizationToken || "" };
```
victor-meng, line 14: This will change the returned object to:
`
{
 	authorization: "token value"
}
`

The returned object before was:
`
{
	header: authorization,
	token: "token value"
}
`

Is this intentional?

src/HostedExplorer.tsx
```diff
@@ -25,8 +25,7 @@ initializeIcons();
 
 const App: React.FunctionComponent = () => {
   // For handling encrypted portal tokens sent via query paramter
-  const params = new URLSearchParams(window.location.search);
-  const [encryptedToken, setEncryptedToken] = React.useState<string>(params && params.get("key"));
+  const [encryptedToken, setEncryptedToken] = React.useState<EncryptedAccessToken>();
```
victor-meng, line 28: Since we are no longer assigning an initial value for `encryptedToken`, in line 36, we should also remove the initial value for `authType` as the expression would always evaluate to `undefined`.

src/Explorer/Tabs/MongoShellTab/MongoShellTabComponent.tsx
```diff
@@ -135,7 +136,7 @@ export default class MongoShellTabComponent extends Component<
     const databaseId = this.props.collection.databaseId;
     const collectionId = this.props.collection.id();
     const apiEndpoint = configContext.BACKEND_ENDPOINT;
-    const encryptedAuthToken: string = userContext.accessToken;
+    const encryptedAuthToken: EncryptedAccessToken = userContext.accessToken;
```
victor-meng, line 139: Have you updated the mongo shell to accept an `EncryptedAccessToken` type? If not, then we should We should pass `encryptedAuthToken.primaryToken` to the iframe instead.

src/Utils/AuthorizationUtils.ts
jawelton74: Yes, this is intentional. I have changed the getAuthorizationHeaders return type to be more easily consumed by it's various callers.

PR #1359, opened 2022-12-09
Fixes issue where empty partition key is treated like nonexistent key…

vchske:
… for Tables API

### DIFF
--- a/src/Common/dataAccess/deleteDocument.ts
+++ b/src/Common/dataAccess/deleteDocument.ts
@@ -13,14 +13,7 @@ export const deleteDocument = async (collection: CollectionBase, documentId: Doc
     await client()
       .database(collection.databaseId)
       .container(collection.id())
-      .item(
-        documentId.id(),
-        documentId.partitionKeyValue === undefined
-          ? undefined
-          : documentId.partitionKeyValue.length === 0
-          ? ""
-          : documentId.partitionKeyValue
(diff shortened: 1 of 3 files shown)


src/Common/dataAccess/deleteDocument.ts
```diff
@@ -13,7 +13,14 @@ export const deleteDocument = async (collection: CollectionBase, documentId: Doc
     await client()
       .database(collection.databaseId)
       .container(collection.id())
-      .item(documentId.id(), documentId.partitionKeyValue?.length === 0 ? undefined : documentId.partitionKeyValue)
+      .item(
+        documentId.id(),
```
victor-meng, line 17: NIT: Can we simply this to `documentId.partitionKeyValue?.length === 0 ? "" :  documentId.partitionKeyValue`? Since like the only case we don't want to return `documentId.partitionKeyValue` is when it's an empty array.

victor-meng (approved):

src/Common/dataAccess/deleteDocument.ts
vchske: I thought of doing it this way but then went with the super clear way to illustrate the 3 options. But I like the smaller way better.

jawelton74 (approved):

victor-meng (approved):

PR #1377, opened 2023-01-19
Add hierachical partition key in add containers in SQL

sunghyunkang1111:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1377?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
+++ b/src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
@@ -123,7 +123,8 @@ describe("ContainerSampleGenerator", () => {
     await generator.createSampleContainerAsync();
   });
 
-  it("should create sample for Mongo API account", async () => {
+  it("should not create any sample for Mongo API account", async () => {
+    const experience = "Sample generation not supported for this API Mongo";
     updateUserContext({
       databaseAccount: {
         properties: {
@@ -132,7 +133,7 @@ describe("ContainerSampleGenerator", () => {
       } as DatabaseAccount,
     });
 
-    expect(ContainerSampleGenerator.createSampleGeneratorAsync(explorerStub)).resolves.toBeTruthy();
+    expect(ContainerSampleGenerator.createSampleGeneratorAsync(explorerStub)).rejects.toMatch(experience);
   });
 
   it("should not create any sample for Table API account", async () => {
--- a/src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
+++ b/src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
@@ -123,8 +123,7 @@ describe("ContainerSampleGenerator", () => {
     await generator.createSampleContainerAsync();
   });
 
-  it("should not create any sample for Mongo API account", async () => {
-    const experience = "Sample generation not supported for this API Mongo";
+  it("should create sample for Mongo API account", async () => {
     updateUserContext({
       databaseAccount: {
         properties: {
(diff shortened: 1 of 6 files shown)


src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -833,12 +909,20 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
   }
 
-  private getPartitionKeyPlaceHolder(): string {
+  private getPartitionKeyPlaceHolder(index?: number): string {
     switch (userContext.apiType) {
       case "Mongo":
         return "e.g., address.zipCode";
       case "Gremlin":
         return "e.g., /address";
+      case "SQL":
```
victor-meng, line 918: Can we combine the two conditional operators into one? It looks like the conditions are the same (index = 0, 1, and 2).

src/Explorer/DataSamples/ContainerSampleGenerator.test.ts
```diff
@@ -123,8 +123,7 @@ describe("ContainerSampleGenerator", () => {
     await generator.createSampleContainerAsync();
   });
 
-  it("should not create any sample for Mongo API account", async () => {
-    const experience = "Sample generation not supported for this API Mongo";
+  it("should create sample for Mongo API account", async () => {
```
victor-meng, line 126: Is there a reason for changing this test? It seems unrelated to your changes.

src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -88,10 +88,10 @@ export interface AddCollectionPanelState {
   collectionId: string;
   enableIndexing: boolean;
   isSharded: boolean;
-  partitionKey: string;
+  partitionKeys: string[];
```
victor-meng, line 91: Are the other APIs going to support hierachical partition keys in the future as well? If not, then I'm not a big fan of changing this to an array for other APIs too. If SQL is the only API that will have multiple partition keys, then we should create a separate property for SQL instead of forcing the partition keys of other APIs to be an array as well.

src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -546,6 +548,90 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
+                        className="panelTextField"
+                        autoComplete="off"
+                        placeholder={this.getPartitionKeyPlaceHolder(index)}
+                        aria-label={this.getPartitionKeyName()}
+                        pattern={".*"}
+                        title={""}
+                        value={subPartitionKey}
+                        onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+                          if (!this.state.subPartitionKeys[index] && !event.target.value.startsWith("/")) {
+                            const subPartitionKeys = this.state.subPartitionKeys.map(
```
victor-meng, line 580: It looks like we don't really need a `map` loop here right? We can directly update `this.state.subPartitionKeys[index]`:

const newSubPartitionKeys = [...this.state.subPartitionKeys];
newSubPartitionKeys[index] = !this.state.subPartitionKeys[index] && !event.target.value.startsWith("/") ? "/" + 		 
    event.target.value.trim() : event.target.value.trim();
this.setState({ subPartitionKeys: newSubPartitionKeys });

src/Explorer/Panes/AddCollectionPanel.tsx
```diff
@@ -1164,11 +1269,16 @@ export class AddCollectionPanel extends React.Component<AddCollectionPanelProps,
...
 
     const uniqueKeyPolicy: DataModels.UniqueKeyPolicy = this.parseUniqueKeys();
-    const partitionKeyVersion = this.state.useHashV2 ? 2 : undefined;
+    const partitionKeyVersion = this.state.useHashV1 ? undefined : 2;
     const partitionKey: DataModels.PartitionKey = partitionKeyString
       ? {
-          paths: [partitionKeyString],
-          kind: "Hash",
+          paths: [
+            partitionKeyString,
```
victor-meng, line 1276: Do we need to include the `partitionKeyString` in the `paths` even when sub partitionkey is enabled?

sunghyunkang1111: yeah, because subpartition keys are stored in separate array so need to concatenate

victor-meng (approved):

asierisayas (approved):

PR #1404, opened 2023-03-20
Sampath accessibility sev 2 2

MokireddySampath:
defects-2278347,2278096, 2280371, 2262636

### DIFF
--- a/less/documentDB.less
+++ b/less/documentDB.less
@@ -2576,6 +2576,10 @@ a:link {
 .querydropdown.placeholderVisible {
   font-style: italic;
 }
+.querydropdown.placeholderVisible::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
+  color: #000;
+  opacity: 1; 
+}
 
 .querydropdown:hover {
   background-color: @AccentLow;
@@ -3087,3 +3091,4 @@ a:link {
   background: white;
   height: 100%;
 }
+
--- a/src/Explorer/Controls/TreeComponent/treeComponent.less
+++ b/src/Explorer/Controls/TreeComponent/treeComponent.less
@@ -3,11 +3,12 @@
 .treeComponent {
   .nodeItem {
     &:focus {
-      outline: 1px dashed @AccentMedium;
+      outline: 2px @AccentMedium;
     }
 
     .treeNodeHeader {
-      padding: @SmallSpace 2px;
+      // padding: @SmallSpace 2px;
+      margin: 2px;
       cursor: pointer;
       display: flex;
 
--- a/images/imgarrowlefticon.svg
+++ b/images/imgarrowlefticon.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" ?>
 <!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
 <svg enable-background="new 0 0 256 256" height="256px" id="Layer_1" version="1.1" viewBox="0 0 256 256" width="256px" xml:space="preserve" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-<path stroke="white" stroke-width="0.5" fill="#000" d="M179.199,38.399c0,1.637-0.625,3.274-1.875,4.524l-85.076,85.075l85.076,85.075c2.5,2.5,2.5,6.55,0,9.05s-6.55,2.5-9.05,0  l-89.601-89.6c-2.5-2.5-2.5-6.551,0-9.051l89.601-89.6c2.5-2.5,6.55-2.5,9.05,0C178.574,35.124,179.199,36.762,179.199,38.399z"/>
+<path stroke="white" stroke-width="0.5" fill="#b5a3a3" d="M179.199,38.399c0,1.637-0.625,3.274-1.875,4.524l-85.076,85.075l85.076,85.075c2.5,2.5,2.5,6.55,0,9.05s-6.55,2.5-9.05,0  l-89.601-89.6c-2.5-2.5-2.5-6.551,0-9.051l89.601-89.6c2.5-2.5,6.55-2.5,9.05,0C178.574,35.124,179.199,36.762,179.199,38.399z"/>
 </svg>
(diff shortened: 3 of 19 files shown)


src/Explorer/Controls/TreeComponent/treeComponent.less
```diff
@@ -3,11 +3,12 @@
 .treeComponent {
   .nodeItem {
     &:focus {
-      outline: 1px dashed @AccentMedium;
+      outline: 2px @AccentMedium;
     }
 
     .treeNodeHeader {
-      padding: @SmallSpace 2px;
+      // padding: @SmallSpace 2px;
```
victor-meng, line 10: Remove?

src/Explorer/Controls/TreeComponent/treeComponent.less
```diff
@@ -3,11 +3,12 @@
...
+      margin: 2px;
```
victor-meng, line 11: It looks like this shrinks the height of the tree node header: https://ms.portal.azure.com/?dataExplorerSource=https%3A%2F%2Fcosmos-explorer-preview.azurewebsites.net%2Fcommit%2Ff27706bc3769d225daa72c2edb1a92552fb5bb60%2Fexplorer.html%3Ffeature.pr%3Dhttps%253A%252F%252Fgithub.com%252FAzure%252Fcosmos-explorer%252Fpull%252F1404%2523sampath-accessibility-sev-2-2#@microsoft.onmicrosoft.com/resource/subscriptions/220fc532-6091-423c-8ba0-66c2397d591b/resourceGroups/vimeng-rg/providers/Microsoft.DocumentDb/databaseAccounts/vimeng-sql-test/dataExplorer.

Is there a particular reason why we need to switch from padding to margin?

less/documentDB.less
```diff
@@ -2576,6 +2576,10 @@ a:link {
 .querydropdown.placeholderVisible {
   font-style: italic;
 }
+.querydropdown.placeholderVisible::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
+  color: #000;
```
victor-meng, line 2580: I don't think the placeholder text should be in black. It will be the same color as the actual text.
image
image

If the original color was too light, you can use a darker gray.

src/Explorer/Controls/TreeComponent/treeComponent.less
MokireddySampath: When a user uses keyboard, the outline around the tree-item is not visible entirely while we use padding, but with margin the outline will be visible which is a requirement for the accessibility.

less/documentDB.less
MokireddySampath: I will use color with less contrast compared to text color used and make relevant changes for placeholder color and make a commit.

less/documentDB.less
```diff
@@ -2577,7 +2577,7 @@ a:link {
   font-style: italic;
 }
 .querydropdown.placeholderVisible::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
-  color: #000;
+  color: #403f3f;
```
victor-meng, line 2580: I think this color is still too close to black:
image

victor-meng: Use #767474 instead:
image

It will meet the 4.5:1 contract ratio requirement.

src/Explorer/Tabs/Tabs.tsx
```diff
@@ -92,6 +92,7 @@ function TabNav({ tab, active, tabKind }: { tab?: Tab; active: boolean; tabKind?
         }
       }}
       className={active ? "active tabList" : "tabList"}
+      style={active ? { fontWeight: "bolder" } : {}}
```
victor-meng, line 95: Which accessibility bug does this fix? I checked all the bugs listed in the description and this doesn't seem to be related to any of them.

MokireddySampath: This resolves the defect reted to color contrast ratio. Here is the link
https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2264174/

src/Explorer/Controls/TreeComponent/treeComponent.less
```diff
@@ -7,8 +7,7 @@
     }
 
     .treeNodeHeader {
-      // padding: @SmallSpace 2px;
-      margin: 2px;
+      margin: 1px;
```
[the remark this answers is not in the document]
MokireddySampath: I think we can add margin: @SmallSpace 2px . It should fix our height issue.

less/documentDB.less
MokireddySampath: It also should have at least 3:1 contrast ratio in comparison with background. I will check both and make a commit for this.

src/Explorer/Controls/TreeComponent/treeComponent.less
MokireddySampath: Removed the commented lines

src/Explorer/Tabs/Tabs.tsx
victor-meng: Thanks Sampath. Even though this is a small fix, please make sure you create a separate PR instead of adding it to this PR which already has more than 5 different fixes.

src/Explorer/Controls/TreeComponent/treeComponent.less
```diff
@@ -7,7 +7,7 @@
     }
 
     .treeNodeHeader {
-      margin: 1px;
+      margin: @SmallSpace 2px;
```
victor-meng, line 10: I tested this and it still didn't look right. Please make sure you test these changes locally.

This is what it looks like with your change:
image

This is what it looks like currently:
image

As you can see, the blue background color no longer covers the entire item due to the margin.

I looked into the issue myself and the root cause of the the outline not being entirely visible is that the parent div is exactly the same size as the `treeNodeHeader` div, and when the outlines are shown, the height and the width of the `treeNodeHeader` div becomes two pixels longer due to the width of the outline being 1px.

An easy fix is to add 1px of padding to the parent div which is `databaseHeader`. I tested this fix locally and it seemed to work:
image

victor-meng: image

victor-meng (approved):

jawelton74 (approved):

PR #1437, opened 2023-04-20
Changes made for the screenreader to readout the label and suffix ass…

MokireddySampath:
…ociated with the input

### DIFF
--- a/src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
+++ b/src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
@@ -15,13 +15,13 @@ import {
 import {
   ChangeFeedPolicyState,
   GeospatialConfigType,
-  getSanitizedInputValue,
   IsComponentDirtyResult,
-  isDirty,
   TtlOff,
   TtlOn,
   TtlOnNoDefault,
   TtlType,
+  getSanitizedInputValue,
+  isDirty,
 } from "../SettingsUtils";
 import { ToolTipLabelComponent } from "./ToolTipLabelComponent";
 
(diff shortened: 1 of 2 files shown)


jawelton74 (approved):

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
```diff
@@ -197,17 +197,20 @@ export class SubSettingsComponent extends React.Component<SubSettingsComponentPr
...
-            styles={getTextFieldStyles(this.props.timeToLiveSeconds, this.props.timeToLiveSecondsBaseline)}
-            type="number"
-            required
-            min={1}
-            max={Int32.Max}
-            value={this.props.timeToLiveSeconds?.toString()}
-            onChange={this.onTimeToLiveSecondsChange}
-            suffix="second(s)"
-          />
+          <div>
```
victor-meng, line 200: Why do we need wrap the textfield around a div?

MokireddySampath: It has no use here. It is a change from the previous commit, I tried using the labels. I will remove the div and make a commit

jawelton74 (dismissed):

src/Explorer/Controls/Settings/SettingsSubComponents/SubSettingsComponent.tsx
```diff
@@ -207,6 +207,7 @@ export class SubSettingsComponent extends React.Component<SubSettingsComponentPr
             value={this.props.timeToLiveSeconds?.toString()}
             onChange={this.onTimeToLiveSecondsChange}
             suffix="second(s)"
+            ariaLabel={`Time to live on ${this.props.timeToLiveSeconds} seconds`}
```
vchske, line 210: I'm not sure what the intent is here but "Time to live on x seconds" doesn't make sense to me. A more grammatically correct sentence would be "Time to live is x seconds." However, I don't think the purpose of aria-label is to provide a name for the current control, i.e. its purpose rather than the value of it as I believe a screen reader reads the value. So my recommendation for this property would be "Time to live in seconds" without any dynamic value being read as part of its name.

vchske (approved):

sunghyunkang1111 (approved):

jawelton74 (approved):

PR #1439, opened 2023-04-21
Tabs have been changed to follow keyboard focus order

MokireddySampath:

### DIFF
--- a/src/Explorer/Controls/Tabs/TabComponent.tsx
+++ b/src/Explorer/Controls/Tabs/TabComponent.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import { AccessibleElement } from "../../Controls/AccessibleElement/AccessibleElement";
 import "./TabComponent.less";
 
 export interface TabContent {
@@ -24,6 +23,7 @@ interface TabComponentProps {
  * We assume there's at least one tab
  */
 export class TabComponent extends React.Component<TabComponentProps> {
+  tabRefs: any = {};
   public constructor(props: TabComponentProps) {
     super(props);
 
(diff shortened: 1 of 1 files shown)


src/Explorer/Controls/Tabs/TabComponent.tsx
```diff
@@ -33,11 +36,41 @@ export class TabComponent extends React.Component<TabComponentProps> {
       throw new Error(msg);
     }
   }
+  state = {
```
victor-meng, line 39: You don't need to define a state for the active tab index. You already have it in the props: `this.props.currrentTabIndex`

src/Explorer/Controls/Tabs/TabComponent.tsx
```diff
@@ -47,26 +80,32 @@ export class TabComponent extends React.Component<TabComponentProps> {
...
         className += " unselectedToggle";
         ariaselected = false;
+        tabindex = -1;
       }
 
       return (
         <div className="tab" key={index}>
-          <AccessibleElement
-            as="span"
+          <span
```
victor-meng, line 96: Why can't we keep using `AccessibleElement`?

src/Explorer/Controls/Tabs/TabComponent.tsx
```diff
@@ -47,26 +80,32 @@ export class TabComponent extends React.Component<TabComponentProps> {
...
             className={className}
             role="tab"
-            onActivated={() => this.setActiveTab(index)}
+            onClick={() => this.setActiveTab(index)}
+            onKeyDown={(event: React.KeyboardEvent<HTMLSpanElement>) => this.handlekeypress(event)}
+            onFocus={() => this.setState({ activeTabIndex: index })}
             aria-label={`Select tab: ${tab.title}`}
             aria-selected={ariaselected}
+            tabIndex={tabindex}
+            ref={(element) => (this.tabRefs[index] = element)}
```
victor-meng, line 105: First of all, it shouldn't be called `tabRefs` since this is just a ref to the tab title, not the entire tab or the tab content. It should be renamed to `tabTitleRefs` to avoid confusion.

Also, if I understand your code correctly, the only reason why you would store the tab title refs is just to focus on the tab title when it's selected. There are other ways you can achieve this. E.g. you can set the `autoFocus` property for the selected tab title.

MokireddySampath: I will redo the solution with the suggestions in the comments

PR #1456, opened 2023-05-23
Defect2280614

MokireddySampath:


jawelton74 (dismissed):

victor-meng:
@jawelton74 I think we should transfer this bug to Fluent UI. We shouldn't be fixing this ourselves. This is an issue with the Fluent UI's `Link` component.

PR #1475, opened 2023-06-06
defect2262580

MokireddySampath:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1475?feature.someFeatureFlagYouMightNeed=true)
defect-https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2262580

The column header has an empty cell in the "Scale" tab of "Scale& Settings". This has been filled with the name "Cost Type" to represent the type of costs that are under the column header's cell.


MokireddySampath:
The defect is not relevant since there are changes in the specific part of the Ui

PR #1487, opened 2023-06-15
Change websocket authentication from url token to getting token from payload

kcheekuri:
In the current implementation we are getting token for websocket authentication from the URL but as per dev sec recommendation we want to pass token from the payload and authenticate every message and close the websocket if the toket is invalid.
This Change is happening for Notebooks, All Terminals, Mongo Schema Analyzers.

Previous WS URL:
image

image

Current Implementation:
image

In current PR we would send token in both url and also from the payload and once server side changes gets through we will remove token in the URL


src/Terminal/JupyterLabAppFactory.ts
```diff
@@ -64,6 +69,14 @@ export class JupyterLabAppFactory {
       }
     }, this);
 
+    if (userContext.features.wsAuthByPayload) {
+      var internalSend = session.send;
```
victor-meng, line 73: Can we use `let` instead?

victor-meng (dismissed):

victor-meng (approved):

tarazou9 (approved):

PR #1505, opened 2023-06-27
Focus order not logical on invoking entities in table 

MokireddySampath:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1505?feature.someFeatureFlagYouMightNeed=true)

defect-https://msdata.visualstudio.com/CosmosDB/_workitems/edit/2278705

Under the "Entities" tab present under "Table database", the keyboard focus moves directly to the table which is not desirable. The focus was forced to moved to the table o rendering the page which has been removed now, so that the focus moves to the "Entities" tab at the top.


src/Explorer/Tables/DataTable/DataTableBindingManager.ts
```diff
@@ -1,15 +1,14 @@
 import * as ko from "knockout";
 import * as _ from "underscore";
 
+import QueryTablesTab from "../../Tabs/QueryTablesTab";
```
vchske, line 4: Let's keep an eye on this kind of reordering of imports. Unless you are getting a lint error, undo these kinds of changes as they produce code churn for no benefit. No need to fix it here. Just keep an eye on it in future PRs.

vchske (approved):

src/Explorer/Tables/DataTable/DataTableBindingManager.ts
MokireddySampath: I will keep an eye on these changes for future changes

jawelton74 (approved):

PR #1507, opened 2023-06-27
[Query Copilot] Add learning bubble to query copilot input if text

v-darkora:
- Add learning bubble on copilot input text if users has not clicked on it for 30 seconds 

[AB#2486830](https://msdata.visualstudio.com/ba574a88-a171-48e0-8fcb-5fef6d23739c/_workitems/edit/2486830)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1507?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/QueryCopilot/QueryCopilotTab.tsx
+++ b/src/Explorer/QueryCopilot/QueryCopilotTab.tsx
@@ -138,6 +138,7 @@ export const QueryCopilotTab: React.FC<QueryCopilotTabProps> = ({
   const [filteredSuggestedPrompts, setFilteredSuggestedPrompts] = useState<SuggestedPrompt[]>(suggestedPrompts);
 
   const handleUserPromptChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    inputEdited.current = true;
     const { value } = event.target;
     setUserPrompt(value);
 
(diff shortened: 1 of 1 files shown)


Klepic95 (approved):

Klepic95 (dismissed):

vchske (dismissed):

src/Explorer/QueryCopilot/QueryCopilotTab.tsx
```diff
@@ -230,6 +234,23 @@ export const QueryCopilotTab: React.FC<QueryCopilotTabProps> = ({
...
+  React.useEffect(() => {
+    if (!inputEdited.current) {
+      setTimeout(() => {
+        if (!inputEdited.current) {
+          toggleCopilotTeachingBubbleVisible();
+          inputEdited.current = true;
+        }
+      }, 30000);
+    }
+  });
```
victor-meng, line 252: I think we only need run this useEffect once the first time user opens the tab, so you should be able to combine this with the useEffect defined above on line 237 (with a dependency array of []).

src/Explorer/QueryCopilot/QueryCopilotTab.tsx
```diff
@@ -241,11 +262,37 @@ export const QueryCopilotTab: React.FC<QueryCopilotTabProps> = ({
           id="naturalLanguageInput"
           value={userPrompt}
           onChange={(_, newValue) => setUserPrompt(newValue)}
```
victor-meng, line 264: We should set `inputEdited.current` to true inside the onChange function too.

victor-meng (approved):

PR #1559, opened 2023-07-27
[Query Copilot] Maintain Query Copilot state when switching tabs

Klepic95:
This pull request contains implementation for Query Copilot Tab to maintain state when switching between active tabs.

Related tasks:
[AB#2531037](https://msdata.visualstudio.com/ba574a88-a171-48e0-8fcb-5fef6d23739c/_workitems/edit/2531037)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1559?feature.someFeatureFlagYouMightNeed=true)


v-darkora (dismissed):

v-darkora (approved):

src/Explorer/Tabs/Tabs.tsx
```diff
@@ -143,11 +166,13 @@ const CloseButton = ({
   active,
   hovering,
   tabKind,
+  queryCopilotState,
 }: {
   tab: Tab;
   active: boolean;
   hovering: boolean;
   tabKind?: ReactTabKind;
+  queryCopilotState: QueryCopilotState;
```
victor-meng, line 175: You don't need to pass the state into this function. All you have to do it call `useQueryCopilotState.getState().resetQueryCopilotStates()` on line 186.

src/Explorer/Tabs/Tabs.tsx
```diff
@@ -158,6 +183,7 @@ const CloseButton = ({
     onClick={(event: React.MouseEvent<HTMLSpanElement, MouseEvent>) => {
       event.stopPropagation();
       tab ? tab.onCloseTabButtonClick() : useTabs.getState().closeReactTab(tabKind);
+      tabKind === ReactTabKind.QueryCopilot && queryCopilotState.resetQueryCopilotStates();
```
victor-meng, line 186: As mentioned above, simply call `useQueryCopilotState.getState().resetQueryCopilotStates()` here

src/Explorer/QueryCopilot/QueryCopilotTab.tsx
```diff
@@ -56,7 +57,7 @@ interface SuggestedPrompt {
 }
 
 interface QueryCopilotTabProps {
-  initialInput: string;
+  queryCopilotState: QueryCopilotState;
```
victor-meng, line 60: I don't think you should pass the states via props. You can simply add a hook to import the states in this file.

src/Explorer/QueryCopilot/QueryCopilotTab.tsx
```diff
@@ -74,52 +75,32 @@ const promptStyles: IButtonStyles = {
...
-  const [showDeletePopup, setShowDeletePopup] = useState<boolean>(false);
-  const [showFeedbackBar, setShowFeedbackBar] = useState<boolean>(false);
-  const [showCopyPopup, setshowCopyPopup] = useState<boolean>(false);
-  const [showErrorMessageBar, setShowErrorMessageBar] = useState<boolean>(false);
 
   const sampleProps: SamplePromptsProps = {
-    isSamplePromptsOpen: isSamplePromptsOpen,
-    setIsSamplePromptsOpen: setIsSamplePromptsOpen,
-    setTextBox: setUserPrompt,
+    isSamplePromptsOpen: queryCopilotState.isSamplePromptsOpen,
```
victor-meng, line 85: Instead of calling `queryCopilotState.xxx` everywhere, you can just descontruct the `queryCopilotState` object at the beginning of this function: e.g.
`const { isSamplePromptsOpen, setIsSamplePromptsOpen, query, setQuery, ...} = queryCopilotState`.
This will make the code much cleaner

src/Explorer/Tabs/Tabs.tsx
```diff
@@ -27,6 +28,7 @@ interface TabsProps {
 
 export const Tabs = ({ explorer }: TabsProps): JSX.Element => {
   const { openedTabs, openedReactTabs, activeTab, activeReactTab, networkSettingsWarning } = useTabs();
+  const queryCopilotState = useQueryCopilotState();
```
victor-meng, line 31: You should move this hook inside `QueryCopilotTab.tsx`.

src/Explorer/Tabs/Tabs.tsx
```diff
@@ -47,15 +49,20 @@ export const Tabs = ({ explorer }: TabsProps): JSX.Element => {
...
+              <TabNav
+                key={ReactTabKind[tab]}
+                active={activeReactTab === tab}
+                tabKind={tab}
+                queryCopilotState={queryCopilotState}
+              />
             ))}
             {openedTabs.map((tab) => (
-              <TabNav key={tab.tabId} tab={tab} active={activeTab === tab} />
+              <TabNav key={tab.tabId} tab={tab} active={activeTab === tab} queryCopilotState={queryCopilotState} />
```
victor-meng, line 60: You don't need to pass the state to the `TabNav` component. The only thing you need from it is the `resetQueryCopilotStates` function, which you can call directly by doing: `useQueryCopilotState.getState().resetQueryCopilotStates()`

src/Explorer/Tabs/Tabs.tsx
Klepic95: getState() does not exist because useQueryCopilotState is a custom hook with states and functions. I will try to fix it in another way.

src/Explorer/Tabs/Tabs.tsx
Klepic95: Same as above, I cannot call getState() method.

src/Explorer/QueryCopilot/QueryCopilotTab.tsx
Klepic95: I descontruct  it from the passed prop.

src/Explorer/Tabs/Tabs.tsx
Klepic95: I don't think that would be possible due to above comment (or I am missing something)

victor-meng (approved):

ato9000 (approved):

PR #1560, opened 2023-07-27
[Query Copilot] Update feature flag after sample data connection info fetch

v-darkora:
- After the sample data info fetch is finished, update the feature flag to true or false, thus even if the feature flag is enabled it will not show query more faster with copilot, unless sample data can be reached

[AB#2573352](https://msdata.visualstudio.com/ba574a88-a171-48e0-8fcb-5fef6d23739c/_workitems/edit/2573352)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1560?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/hooks/useKnockoutExplorer.ts
+++ b/src/hooks/useKnockoutExplorer.ts
@@ -435,11 +435,6 @@ async function updateContextForSampleData(explorer: Explorer): Promise<void> {
   const sampleDataConnectionInfo = parseResourceTokenConnectionString(data.connectionString);
   updateUserContext({
     sampleDataConnectionInfo,
-    features: extractFeatures(
-      new URLSearchParams({
-        "feature.enableCopilot": sampleDataConnectionInfo ? "true" : "false",
-      })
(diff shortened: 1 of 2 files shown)


src/hooks/useKnockoutExplorer.ts
```diff
@@ -433,7 +433,14 @@ async function updateContextForSampleData(explorer: Explorer): Promise<void> {
 
   const data: SampledataconnectionResponse = await response.json();
   const sampleDataConnectionInfo = parseResourceTokenConnectionString(data.connectionString);
-  updateUserContext({ sampleDataConnectionInfo });
+  updateUserContext({
+    sampleDataConnectionInfo,
```
victor-meng, line 437: No need to overwrite the feature flag. Just add this condition when deciding whether to render the button or not.

victor-meng (approved):

Klepic95 (approved):

PR #1584, opened 2023-08-21
[Query Copilot] V2 Backend integration

Klepic95:
This pull request contains implementation related to Query Copilot version 2 and it's integration with the backend services for generating queries.

Related tasks:
[AB#2635248](https://msdata.visualstudio.com/ba574a88-a171-48e0-8fcb-5fef6d23739c/_workitems/edit/2635248)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1584?feature.someFeatureFlagYouMightNeed=true)


src/Explorer/QueryCopilot/V2/Footer/Footer.tsx
```diff
@@ -29,18 +38,60 @@ export const Footer: React.FC = (): JSX.Element => {
...
     if (userPrompt.trim() !== "") {
-      setChatMessages([...chatMessages, userPrompt]);
-      setUserPrompt("");
       setIsGeneratingQuery(true);
+      useTabs.getState().setIsTabExecuting(true);
+      useTabs.getState().setIsQueryErrorThrown(false);
+      setChatMessages([...chatMessages, { source: 0, message: userPrompt }]);
+      try {
+        if (shouldAllocateContainer) {
+          await explorer.allocateContainer();
```
v-darkora, line 53: I don't think we should do any http request inside this component. If we have a client for calling the api let's handle all the request calls and handling, there instead of putting it in the component.

Klepic95: Resolved

src/Explorer/QueryCopilot/V2/Footer/Footer.tsx
```diff
@@ -29,18 +37,55 @@ export const Footer: React.FC = (): JSX.Element => {
...
+        setIsGeneratingQuery(false);
+        useTabs.getState().setIsTabExecuting(false);
+      }
     }
   };
 
+  const startSentMessageProcess = async () => {
+    await handleSentQueryRequest();
+  };
+
```
v-darkora, line 88: I was also thinking on this part too, as why should the footer have any responsibility over this? My idea is to just call one method in the client and that's it. All of the useQuery set states to the client as well. If that makes sense?

Klepic95: Everyting from handleSentQueryRequest() is now moved to QueryCopilotClient.

v-darkora (approved):

ato9000 (approved):

PR #1607, opened 2023-09-12
Initial Fabric support

sevoku:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1607?feature.someFeatureFlagYouMightNeed=true)

This adds initial support for Fabric as a new platform.
Data Explorer will be loaded into an iframe inside our Fabric extension iframe.
* Adds new Platform.Fabric option
* Conditionally loads a different Fluent Theme and overrides some default styles
* Hides many advanced features not required in fabric such as Notebooks, Scripts, etc.
* Replaces the resource tree with a new Fluent implementation

PR #1614, opened 2023-09-15
Fix active tabs border on Fabric

sevoku:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1614?feature.someFeatureFlagYouMightNeed=true)

With my last fix in https://github.com/Azure/cosmos-explorer/pull/1607/commits/c8b533f0aab657b6da47305f70cd8e5ad9939948 for Portal, I broke the active Tab design for Fabric, sorry about that!

Bug:
image

Fixed:
image

### DIFF
--- a/less/documentDBFabric.less
+++ b/less/documentDBFabric.less
@@ -70,6 +70,10 @@ a:focus {
   border-bottom: 2px solid @FabricAccentMedium;
 }
 
+.nav-tabs>li.active>.tabNavContentContainer>.tab_Content>.tabNavText {
+  border-bottom: 0px none transparent;
+}
+
 .tabNavContentContainer {
   padding: @SmallSpace 0px @SmallSpace 0px;
 


less/documentDBFabric.less
```diff
@@ -70,6 +70,10 @@ a:focus {
   border-bottom: 2px solid @FabricAccentMedium;
 }
 
+.nav-tabs>li.active>.tabNavContentContainer>.tab_Content>.tabNavText {
```
languy, line 73: Shouldn't we be having the same element, but just switch color instead?

languy (approved):

less/documentDBFabric.less
sevoku: No because we let the parent container have the underline on Fabric, not just the text. It spans the icon and close button as well, making it more consistent with the hover effect.

ato9000 (approved):

PR #1647, opened 2023-10-06
Implement retrieval of authorization token for Fabric via iframe rpc

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1647?feature.someFeatureFlagYouMightNeed=true)

For Fabric, the authorization token is obtained from the iframe parent (Fabric extension client).
Implement the contract between Fabric and Data Explorer to request and receive the token via iframe post messages.


src/Common/MessageHandler.ts
```diff
@@ -22,7 +22,7 @@ export function handleCachedDataMessage(message: any): void {
   if (messageContent.error != null) {
     cachedDataPromise.deferred.reject(messageContent.error);
   } else {
-    cachedDataPromise.deferred.resolve(JSON.parse(messageContent.data));
```
languy, line 25: Note that `handleCachedDataMessage()` isn't used anywhere, so I'm modifying the behavior to return the message content data as is, rather than parsing it.

ato9000 (approved):

ato9000 (approved):

PR #1658, opened 2023-10-12
Explorer null handling

JustinKol:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1658?feature.someFeatureFlagYouMightNeed=true)

Data Explorer was not handling NULL values properly.

Issue 1: If a row had fields which where NULL, when opening the row to edit, the fields that were NULL did not show so the user could insert values in those fields. 

Issue 2: When attempting to insert a new row, when trying to leave non-string data types as empty, errors are thrown from backend but not caught and displayed properly in DE.

RESOLUTION: The backend requires a valid value for all types so it was advised to display a Warning on the edit panel that null values cannot be edited.

### DIFF
--- a/src/Explorer/Tables/TableDataClient.ts
+++ b/src/Explorer/Tables/TableDataClient.ts
@@ -19,9 +19,6 @@ import Explorer from "../Explorer";
 import * as TableConstants from "./Constants";
 import * as Entities from "./Entities";
 import * as TableEntityProcessor from "./TableEntityProcessor";
-import { getLocalDateTime } from "./QueryBuilder/DateTimeUtilities";
-import { v4 as uuidv4 } from "uuid";
-import moment from "moment";
 
 export interface CassandraTableKeys {
   partitionKeys: CassandraTableKey[];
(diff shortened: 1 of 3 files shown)


asierisayas (approved):

asierisayas (dismissed):

src/Explorer/Tables/TableDataClient.ts
```diff
@@ -208,6 +208,9 @@ export class CassandraAPIDataClient extends TableDataClient {
           !originalDocument[property] ||
           newEntity[property]._.toString() !== originalDocument[property]._.toString()
         ) {
+          if (newEntity[property]._.toString() === "" || newEntity[property]._.toString() === undefined) {
```
jawelton74, line 211: I believe that undefined.ToString() will produce "undefined". Maybe change that second part of the clause to newEntity[property]._ === undefined ?

JustinKol: removed .toString()

jawelton74 (approved):

asierisayas (approved):

PR #1673, opened 2023-10-20
Check for accounts that are restricted from connection string login.

jawelton74:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1673?feature.someFeatureFlagYouMightNeed=true)

This change adds a call to the portal backend to determine if an account has been restricted from connection string based login.
If account has been restricted, user is informed via an error tooltip.

### DIFF
--- a/src/Platform/Hosted/Components/ConnectExplorer.tsx
+++ b/src/Platform/Hosted/Components/ConnectExplorer.tsx
@@ -42,16 +42,6 @@ export const isAccountRestrictedForConnectionStringLogin = async (connectionStri
   return (await response.text()) === "True";
 };
 
-export const hideErrorDetails = () => {
-  window.document.getElementById("errorDetails").innerText = "";
-  window.document.getElementById("errorDetailsInfoTooltip").style.display = "none";
-};
-
-export const showErrorDetails = (errorMessage: string) => {
-  window.document.getElementById("errorDetails").innerText = errorMessage;
-  window.document.getElementById("errorDetailsInfoTooltip").style.display = "inline";
-};
-
 export const ConnectExplorer: React.FunctionComponent<Props> = ({
   setEncryptedToken,
   login,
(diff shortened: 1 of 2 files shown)


src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -30,6 +30,28 @@ export const fetchEncryptedToken = async (connectionString: string): Promise<str
...
+  const response = await fetch(url, { headers, method: "POST" });
+  if (!response.ok) {
+    throw response;
+  }
+
+  return (await response.text()) === "True";
+};
+
+export const hideErrorDetails = () => {
+  window.document.getElementById("errorDetails").innerText = "";
```
vchske, line 46: You don't want to edit the DOM directly like this. If logic is self contained within a component, use state (like line 62) is the way to go. There are tons of examples throughout this repo's *.tsx files showing various React-native ways to alter the UI.

vchske (changes requested):

vchske (approved):

sunghyunkang1111 (dismissed):

src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -53,12 +66,18 @@ export const ConnectExplorer: React.FunctionComponent<Props> = ({
               id="connectWithConnectionString"
               onSubmit={async (event) => {
                 event.preventDefault();
+                setErrorMessage("");
 
                 if (isResourceTokenConnectionString(connectionString)) {
```
ato9000, line 71: we should probably also block resource tokens. These connection strings should also work to on if account is restricted

ato9000 (dismissed):

vchske (approved):

sunghyunkang1111 (approved):

PR #1675, opened 2023-10-25
Adds restarting vcore mongo shell if user enters wrong password

vchske:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1675)

On vCore Mongo, if the user enters an incorrect password, the whole terminal window closes. This adds a check for a specific message in the stdout which indicates the shell didn't accept the password and in that case recreates the terminal.

One of two things could happen in the future:
1. Once Phoenix adds the feature to fully login without the need for user input, this change could be undone.
2. If this is useful for other scenarios, other checks could be created.

### DIFF
--- a/src/Terminal/index.ts
+++ b/src/Terminal/index.ts
@@ -11,6 +11,8 @@ import { JupyterLabAppFactory } from "./JupyterLabAppFactory";
 import { TerminalProps } from "./TerminalProps";
 import "./index.css";
 
+let session: ITerminalConnection | undefined;
+
 const createServerSettings = (props: TerminalProps): ServerConnection.ISettings => {
   let body: BodyInit | undefined;
   let headers: HeadersInit | undefined;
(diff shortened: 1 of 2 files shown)


src/Terminal/index.ts
```diff
@@ -59,28 +61,38 @@ const initTerminal = async (props: TerminalProps): Promise<ITerminalConnection |
...
+    window.parent.postMessage(
+      { type: MessageTypes.CloseTab, data: { tabId: props.tabId }, signature: "pcIframe" },
+      window.document.referrer,
+    );
+  }
 };
 
 const main = async (): Promise<void> => {
-  let session: ITerminalConnection | undefined;
+  //let session: ITerminalConnection | undefined;
```
ato9000, line 95: nit: delete commented line

jawelton74 (approved):

languy (approved):

PR #1678, opened 2023-10-25
Minor changes to vcore mongo quickstart based on feedback

vchske:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1678)

### DIFF
--- a/src/Explorer/Quickstart/VCoreMongoQuickstartGuide.tsx
+++ b/src/Explorer/Quickstart/VCoreMongoQuickstartGuide.tsx
@@ -11,11 +11,15 @@ import {
 } from "@fluentui/react";
 import { customPivotHeaderRenderer } from "Explorer/Quickstart/Shared/QuickstartRenderUtilities";
 import {
+  findAndSortCommand,
+  findAndSortCommandForDisplay,
+  findByPagesCommand,
+  findByPagesCommandForDisplay,
+  findOrwellCommand,
+  findOrwellCommandForDisplay,
   loadDataCommand,
   newDbAndCollectionCommand,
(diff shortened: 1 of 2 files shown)


jawelton74 (approved):

src/Explorer/Quickstart/VCoreMongoQuickstartGuide.tsx
```diff
@@ -190,17 +194,75 @@ export const VcoreMongoQuickstartGuide: React.FC = (): JSX.Element => {
...
+                  />
+                  <IconButton
+                    iconProps={{
+                      iconName: "Copy",
+                    }}
+                    onClick={() => onCopyBtnClicked("#findByPagesCommand")}
+                  />
+                </Stack>
+                <DefaultButton
+                  style={{ marginTop: 32, width: 110 }}
```
languy, line 254: How come some of the buttons have a marginTop of 32 and some 16? Can we define constants in this file for style and have the buttons use them?

languy (approved):

src/Explorer/Quickstart/VCoreMongoQuickstartGuide.tsx
vchske: Based on the layout there needed to be more space between the sets of queries.
I agree that, generally speaking, it is better to have constants. What would be best is if DE had a single, consistent, visual "library" that *every* FC and class used to govern its layout. This would help to create a visual language that was common across all the different workflows. But obviously that's a big undertaking at this point.
Having said all that, in the absence of a unified visual library, within a single component, especially one that is in flux like this one, I'm ok if individual stylings are applied without the use of constants.

PR #1680, opened 2023-10-26
Fabric: clean up RPC and support show/hide toolbar message

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1680)

* Now use Promise to post message to Fabric
* Add debounce when requesting allResourceToken to fast
* Simplify schedule request token
* Modify initialization sequence: first get the connectionId from Fabric, then request token
* Add `scope` to `sendCachedMessage` in order to make the posted messages unique and avoid receiving messages for different artifact when switching artifacts
* Refactored rpc contracts to keep track of versions


src/hooks/useKnockoutExplorer.ts
```diff
@@ -108,31 +105,19 @@ async function configureFabric(): Promise<Explorer> {
 
         switch (data.type) {
```
ato9000, line 106: Let's add support for version property

switch (`${data.type}_${data.version}`) {
    case "initialize_": {
        const fabricDatabaseConnectionInfo: FabricDatabaseConnectionInfo = {
            endpoint: data.message.endpoint,
            databaseId: data.message.databaseId,
            resourceTokens: data.message.resourceTokens as { [resourceId: string]: string },
            resourceTokensTimestamp: data.message.resourceTokensTimestamp,
        };
        explorer = await createExplorerFabric(fabricDatabaseConnectionInfo);
        resolve(explorer);

        explorer.refreshAllDatabases().then(() => {
            openFirstContainer(explorer, fabricDatabaseConnectionInfo.databaseId);
        });
        scheduleRefreshDatabaseResourceToken();
        break;
    }
    case "initialize_ fabric2": {
        explorer = createExplorerFabric(data.message);
        await scheduleRefreshDatabaseResourceToken(true);
        resolve(explorer);
        await explorer.refreshAllDatabases();
        openFirstContainer(explorer, userContext.fabricContext.databaseConnectionInfo.databaseId);
        break;
    }

ato9000 (changes requested):

src/hooks/useKnockoutExplorer.ts
languy: Added a new message and explictly keep track of the message versions.

src/hooks/useKnockoutExplorer.ts
```diff
@@ -101,40 +102,35 @@ async function configureFabric(): Promise<Explorer> {
...
           case "initialize": {
-            const fabricDatabaseConnectionInfo: FabricDatabaseConnectionInfo = {
-              endpoint: data.message.endpoint,
-              databaseId: data.message.databaseId,
-              resourceTokens: data.message.resourceTokens as { [resourceId: string]: string },
-              resourceTokensTimestamp: data.message.resourceTokensTimestamp,
-            };
-            explorer = await createExplorerFabric(fabricDatabaseConnectionInfo);
+            const fabricVersion = data.version;
```
ato9000, line 112: This should be an array of versions so we can make Fabric support backward versions

ato9000 (dismissed):

src/hooks/useKnockoutExplorer.ts
languy: This is the version of the fabric client that is currently talking to DataExplorer. The array containing the supported versions is hardcoded in the respective clients.

src/Common/MessageHandler.ts
```diff
@@ -54,6 +63,10 @@ export function sendMessage(data: any): void {
   });
 }
 
+/**
+ *
+ * @param data Overwrite the data property of the message
```
vchske, line 68: Do we need to add this comment? Doesn't look like the method accepts a param.

vchske (dismissed):

src/Common/MessageHandler.ts
languy: Looks like the comment was for the previous method. Thanks for the catch!

src/Contracts/MessageTypes.ts
```diff
@@ -37,10 +43,9 @@ export enum MessageTypes {
   DisplayNPSSurvey,
   OpenVCoreMongoNetworkingBlade,
   OpenVCoreMongoConnectionStringsBlade,
-
-  // Data Explorer -> Fabric communication
-  GetAuthorizationToken,
-  GetAllResourceTokens,
+  GetAuthorizationToken, // Data Explorer -> Fabric
```
jawelton74, line 46: Laurent - can we plan to split message types into Portal message types and Fabric message types in the near future ?

jawelton74 (approved):

src/Contracts/MessageTypes.ts
languy: The constraint is that we are leveraging the existing messaging infrastructure and its an enum (ie Fabric receives a number), so if we have a completely separate enum file for Fabric, we could get collisions.
Let's take this offline.

sunghyunkang1111 (approved):

PR #1717, opened 2024-01-02
Bug 2819674: [Keyboard navigation - Cosmos DB Query Copilot - Save query]: Keyboard focus loss is observed by invoking the "X" (Close) button which is present in the 'Save query' dialog.

MokireddySampath:
… dialog

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1717?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Panes/PanelContainerComponent.tsx
+++ b/src/Explorer/Panes/PanelContainerComponent.tsx
@@ -76,7 +76,42 @@ export class PanelContainerComponent extends React.Component<PanelContainerProps
     );
   }
 
+  public isFocusable(element: HTMLElement) {
+    return (
+      element.tabIndex >= 0 ||
+      (element instanceof HTMLAnchorElement && element.href) ||
+      (element instanceof HTMLAreaElement && element.href) ||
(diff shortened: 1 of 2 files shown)


asierisayas (approved):

asierisayas (approved):

sindhuba (approved):

src/Explorer/Panes/PanelContainerComponent.tsx
```diff
@@ -76,7 +76,44 @@ export class PanelContainerComponent extends React.Component<PanelContainerProps
     );
   }
 
+  public isFocusable(element: HTMLElement) {
```
vchske, line 79: isFocusable and findFocusableParent seem like they might be useful elsewhere? If you think so too, they should be moved to a utility file in a common folder.

vchske (approved):

src/Explorer/Panes/PanelContainerComponent.tsx
MokireddySampath: It is useul for all the instances using this dialog. I will move them to a common folder.

PR #1720, opened 2024-01-09
Fabric toolbar style and support to show/hide it

sevoku:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1720?feature.someFeatureFlagYouMightNeed=true)


src/hooks/useKnockoutExplorer.ts
```diff
@@ -131,6 +131,11 @@ async function configureFabric(): Promise<Explorer> {
             handleCachedDataMessage(data);
             break;
           }
+          case "setToolbarStatus": {
+            const cdeToolbar = document.getElementById("cde-commandBarContainer");
```
languy, line 135: It would be better to do it the React way and avoid manipulating the DOM. Can we set the status somewhere in some global context and inject the height as property into the react component instead?

sevoku: Resolved via https://github.com/Azure/cosmos-explorer/pull/1720/commits/2ae6392b7e11d8e2c72c3a012169b54f74238569, thanks!

languy (approved):

PR #1745, opened 2024-02-13
add feedback policies integration with copilot

sunghyunkang1111:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1745?feature.someFeatureFlagYouMightNeed=true)

Adds feedback policies settings into the user context. Will only work with portal.
Removes teaching bubbles and modals
updated staging endpoint for Phoenix to Prod

Manually tested locally by changing the policy settings in tenant admin settings.

### DIFF
--- a/src/Explorer/QueryCopilot/QueryCopilotPromptbar.tsx
+++ b/src/Explorer/QueryCopilot/QueryCopilotPromptbar.tsx
@@ -539,84 +539,90 @@ export const QueryCopilotPromptbar: React.FC<QueryCopilotPromptProps> = ({
       </Stack>
 
       {showFeedbackBar && (
-        <Stack style={{ backgroundColor: "#FFF8F0", padding: "2px 8px" }} horizontal verticalAlign="center">
-          <Text style={{ fontWeight: 600, fontSize: 12 }}>Provide feedback on the query generated</Text>
-          {showCallout && !hideFeedbackModalForLikedQueries && (
-            <Callout
-              role="status"
-              style={{ padding: 8 }}
-              target="#likeBtn"
-              onDismiss={() => {
-                setShowCallout(false);
-                SubmitFeedback({
-                  params: {
-                    generatedQuery: generatedQuery,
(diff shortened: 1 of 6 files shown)


vchske (approved):

src/Explorer/QueryCopilot/QueryCopilotPromptbar.tsx
```diff
@@ -539,84 +539,90 @@ export const QueryCopilotPromptbar: React.FC<QueryCopilotPromptProps> = ({
...
-              <Text>
-                Thank you. Need to give{" "}
-                <Link
-                  onClick={() => {
+        <Stack
+          style={{ backgroundColor: "#FFF8F0", padding: "2px 8px", minHeight: 32 }}
+          horizontal
+          verticalAlign="center"
+        >
+          {userContext.feedbackPolicies?.policyAllowFeedback && (
```
jawelton74, line 547: Does this mean that if we don't have feedbackPolicies in the userContext then we won't display feedback ?

sunghyunkang1111: Yeah. I think this will only apply if someone decides to use copilot in standalone DE.

jawelton74: Or what happens if the  customer doesn't have feedback policies defined for their tenant. Will it be null in that case ?

sunghyunkang1111: Ah, the feedback policy API from the portal SDK will return object with all True values if they dont have anything configured. So that shouldn't be a case.

jawelton74: Got it - cool. But we still need to support the standalone DE scenario ?

jawelton74 (approved):

vchske (approved):

PR #1746, opened 2024-02-16
Improve error handling when acquiring aad tokens

jawelton74:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1746?feature.someFeatureFlagYouMightNeed=true)

This change adds the following:
- Handling of cases where acquiring tokens requires interaction, by introducing a wrapper function which first attempts to acquire token silently and then interactively if the first attempt fails. Example scenario for this is when MFA has been enabled for user.
- Add a landing page for authorization token failures to make it more prominent to user. It also handles cases where popups are disabled by providing a link to retry (see comments in code for explanation).


src/hooks/useAADAuth.ts
```diff
@@ -61,24 +69,60 @@ export function useAADAuth(): ReturnType {
...
+    try {
+      const armToken = await acquireTokenWithMsal(msalInstance, {
+        authority: `${configContext.AAD_ENDPOINT}${tenantId}`,
+        scopes: [`${configContext.ARM_ENDPOINT}/.default`],
       });
+
+      setArmToken(armToken);
+      setAuthFailure(null);
+    } catch (error) {
+      if (error instanceof msal.AuthError && error.errorCode === msal.BrowserAuthErrorMessage.popUpWindowError.code) {
```
languy, line 86: Consider sending something to telemetry in all these `catch` cases.

languy (approved):

src/hooks/useAADAuth.ts
jawelton74: Good suggestion - added tracing in the main util function.

vchske (approved):

asierisayas (approved):

PR #1770, opened 2024-03-19
Migrate DocumentsTab to React and add bulk delete and column resize

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1770)
Replace the current `DocumentsTab` knockout component with a `DocumentsTabV2` React component.
This applies to NoSQL and Mongo (Gremlin, Cassandra and Tables use other tab types to display the documents).
The functionality should be the same as before plus the following:
* Ability to select multiple documents to delete
* Use Bulk Operations to perform multiple deletes at the same time (noSQL)
* Ability to resize column width of the documents list
* Import `@uiw/react-split` package to provide the vertical slider between the items and the document display, but patch the library to fix compile issues.
* Modify jest config to output `.lcov` file, so we can visualize test code coverage in vscode (see instructions below)
---
### Do this to display unit test coverage in source file in Visual Studio Code ###
1. Install the "Coverage Gutters" vscode extension
2. Open "Preferences: User Settings" (ctrl-shift-P and type "preferences")
3. Search for "coverage"
4. Under "Coverage gutter: Coverage Base Dir", set the value to: `coverage/**` (which will point to the `coverage/` directory where the `.lcov` file is generated by jest)
5. `npm t` to execute all unit tests and generate the coverage file (`coverage/coverage.lcov`)
6. Open the source file that you want to inspect
7. To display code coverage: execute the vscode command (Ctrl-Shift-P): Coverage Gutter: Display Coverage. You should see red and green blocks on the left margin next to the line numbers indicating whether the corresponding line is covered by a unit test.

image


asierisayas (approved):

less/documentDB.less
```diff
@@ -2372,7 +2371,7 @@ a:link {
   overflow-y: scroll;
 }
 
-.tabs {
+.t abs {
```
analogrelay, line 2374: Is this intentional?

src/Common/dataAccess/deleteDocument.ts
```diff
@@ -24,3 +25,58 @@ export const deleteDocument = async (collection: CollectionBase, documentId: Doc
...
+      }));
+
+      const promise = v2Container.items.bulk(operations).then((bulkResult) => {
+        const result: DocumentId[] = [];
+        documentIdsChunk.forEach((documentId, index) => {
+          if (bulkResult[index].statusCode === 204) {
+            result.push(documentId);
+          }
+        });
+        return result;
```
analogrelay, line 66: non-blocking style option: This could be a filter.

```suggestion
        return documentIdsChunk.filter((_, index) => bulkResult[index].statusCode === 204);
```

src/Common/dataAccess/deleteDocument.ts
```diff
@@ -24,3 +25,58 @@ export const deleteDocument = async (collection: CollectionBase, documentId: Doc
...
+      });
+      promiseArray.push(promise);
+    }
+
+    const allResult = await Promise.all(promiseArray);
+    const flatAllResult = Array.prototype.concat.apply([], allResult);
+    logConsoleInfo(`Successfully deleted ${getEntityName(true)}: ${flatAllResult.length} out of ${nbDocuments}`);
```
analogrelay, line 73: nit: There's an unnecessary plural here if you're deleting one item:

image

Could fix this with:

```suggestion
    logConsoleInfo(`Successfully deleted ${getEntityName(flatAllResult.length > 1)}: ${flatAllResult.length} out of ${nbDocuments}`);
```

src/Common/dataAccess/deleteDocument.ts
```diff
@@ -24,3 +25,58 @@ export const deleteDocument = async (collection: CollectionBase, documentId: Doc
...
+    // TODO: handle case result.length != nbDocuments
+    return flatAllResult;
+  } catch (error) {
+    handleError(error, "DeleteDocuments", `Error while deleting ${documentIds.length} ${getEntityName(true)}`);
```
analogrelay, line 77: Similar plural thing here:

```suggestion
    handleError(error, "DeleteDocuments", `Error while deleting ${documentIds.length} ${getEntityName(documentIds.length > 1)}`);
```

You could also choose to have `getEntityName` accept a `count` parameter and do the size check there instead of having it expect a boolean indicating if there is more than one element. Then the call sites would be `getEntityName(documentIds.length)` instead.

src/Explorer/Tabs/DocumentsTabV2/DocumentsTableComponent.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+describe("DocumentsTableComponent", () => {
+  const createMockProps = (): IDocumentsTableComponentProps => ({
+    items: [
+      { [ID_HEADER]: "1", [PARTITION_KEY_HEADER]: "pk1" },
+      { [ID_HEADER]: "2", [PARTITION_KEY_HEADER]: "pk2" },
+      { [ID_HEADER]: "3", [PARTITION_KEY_HEADER]: "pk3" },
+    ],
+    onItemClicked: (index: number): void => {
+      index;
```
analogrelay, line 17: What's the purpose of this function? It seems to be a no-op since you're not returning the `index` value. If so, it might be clearer as `() => {}` or `(_: number): void => {}` (if the parameter/return types are essential)

src/Explorer/Tabs/DocumentsTabV2/DocumentsTableComponent.test.tsx
analogrelay, line 1: I'm curious if this could use snapshot tests like some other components do. I'm a fan of snapshot tests for this kind of UI since they allow you to validate the entire UI without having to write detailed asserts. It does require validating the snapshot itself, but often you can get sufficient verification by recreating the test situation in the running app and validating it there. Once you've validated that the UI looks correct, the snapshot will keep the entire thing stable until you make an intentional change.

That's just a thought though. These tests look like they do validate the core scenarios and feel sufficient as well.

src/Explorer/Tabs/DocumentsTabV2/DocumentsTabV2Mongo.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+      wrapper = await waitForComponentToPaint(wrapper);
+    });
+
+    afterEach(() => {
+      wrapper.unmount();
+    });
+
+    it("renders by default the first document", async () => {
+      // expect(wrapper.findWhere((node) => node.text().includes(PROPERTY_VALUE)).exists()).toBeTruthy();
+    });
```
analogrelay, line 151: Unnecessary test? Or unintentionally commented out?

analogrelay (dismissed):
Nothing blocking, but I had some non-blocking ideas/thoughts/questions.

Looks great in the preview! I'm excited to get this in. Once I migrate the Resource Tree and Tab bar to the new design we'll have a large chunk of the app using Fluent UI 9 🎉

src/Common/dataAccess/deleteDocument.ts
languy: That would be wise to make it work for localization, since [some languages have multiple plural forms](https://phrase.com/blog/posts/pluralization/). For now, I'll keep the API as boolean to minimize the changes in non-related files (we may have to rework the mongo vs non-mongo terminology utility anyway).

less/documentDB.less
languy: No. Thanks for catching it.

src/Explorer/Tabs/DocumentsTabV2/DocumentsTabV2Mongo.test.tsx
languy: Unintentionally commented out.

src/Explorer/Tabs/DocumentsTabV2/DocumentsTableComponent.test.tsx
languy: I totally agree (I totally forgot about the snapshots). The targeted tests were good initially to protect the code that displayed the important data, but the dom queries for the assertions are often finicky or rely on fluent's stability. Snapshots protect everything including styling.

sevoku (approved):
Gave this another shot with Mongo, works really nicely!

jawelton74:
I'm seeing a bug where the Delete button is removed from the Command bar when selecting all documents  for Mongo API accounts. 
My repro:

- Apply a filter
- Clear filter
- Select all documents via top checkbox
- Delete button is not enabled

image

- Selecting individual documents enables the Delete button but selecting all documents now makes it disappear.

languy:
> I'm seeing a bug where the Delete button is removed from the Command bar when selecting all documents for Mongo API accounts. My repro:
> 
> * Apply a filter
> * Clear filter
> * Select all documents via top checkbox
> * Delete button is not enabled
> 
> image
> 
> * Selecting individual documents enables the Delete button but selecting all documents now makes it disappear.

Good catch: in the current implementation, the "Delete" button doesn't mean "delete current document" anymore, but means delete selected documents.

analogrelay (approved):

sunghyunkang1111 (approved):

Pilchie:
Yay!

PR #1772, opened 2024-03-19
Add container ids to tabs

JustinKol:
Adding container IDs to tabs in order to better navigate tabs (viewing, closing, etc)

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1772?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/Explorer/Tabs/TabsBase.ts
+++ b/src/Explorer/Tabs/TabsBase.ts
@@ -146,7 +146,7 @@ export default class TabsBase extends WaitsForTemplateViewModel {
     const coll = this.collection?.id();
     if (coll) {
       if (coll.length > 8){
-        return coll.substr(0,5) + '....' + options.title;
+        return coll.slice(0,5) + '....' + options.title;
       } else {
       return coll + "." + options.title;
     }
--- a/src/Explorer/Tabs/TabsBase.ts
+++ b/src/Explorer/Tabs/TabsBase.ts
@@ -158,7 +158,7 @@ export default class TabsBase extends WaitsForTemplateViewModel {
         return db + "." + options.title;
       }
     } else {
-      return "";
+      return options.title;
     }
   }
 
--- a/src/Explorer/Tabs/TabsBase.ts
+++ b/src/Explorer/Tabs/TabsBase.ts
@@ -144,7 +144,7 @@ export default class TabsBase extends WaitsForTemplateViewModel {
   }
 
   public getTitle(): string {
-    return (this.collection?.id()) || (this.database?.id());
+    return this.collection?.id() || this.database?.id();
   }
 
   /** Renders a Javascript object to be displayed inside Monaco Editor */
--- a/src/Explorer/Tabs/TabsBase.ts
+++ b/src/Explorer/Tabs/TabsBase.ts
@@ -42,8 +42,8 @@ export default class TabsBase extends WaitsForTemplateViewModel {
     this.tabKind = options.tabKind;
     this.tabTitle = ko.observable<string>(this.getTitle(options));
     this.tabPath =
-      (this.collection &&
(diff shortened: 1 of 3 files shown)


asierisayas (approved):

src/Explorer/Tabs/TabsBase.ts
```diff
@@ -40,7 +40,7 @@ export default class TabsBase extends WaitsForTemplateViewModel {
     this.database = options.database;
     this.rid = options.rid || (this.collection && this.collection.rid) || "";
     this.tabKind = options.tabKind;
-    this.tabTitle = ko.observable<string>(options.title);
+    this.tabTitle = ko.observable<string>(this.getTitle() + " - " + options.title);
```
sevoku, line 43: `options.title` already contains the collection ID, see https://github.com/Azure/cosmos-explorer/blob/00ffc2d5a950c62d2776ee56d9d0f00376bca95f/src/Explorer/Tree/Collection.ts#L319, so we end up with "[Collection ID] - [Collection ID] - Items" in the title.

sevoku (commented):
This doesn't look right:
image
What's the goal we want to achieve here? Also no more Path in the tooltip?

Without the change it was:
image
(note: database and container have the same title "test", sorry for the bad example)

src/Explorer/Tabs/TabsBase.ts
JustinKol: I do like the arrows.

I believe the ASK was worded a little incorrectly but I went with it. The user wants to be able to identify their tabs easier without having to click on each Collection again to find the one they want.

JustinKol: I have reverted the tooltip and corrected the logic.

src/Explorer/Tabs/TabsBase.ts
```diff
@@ -143,6 +142,26 @@ export default class TabsBase extends WaitsForTemplateViewModel {
     return (this.collection && this.collection.container) || (this.database && this.database.container);
   }
 
+  public getTitle(options: ViewModels.TabOptions): string {
+    const coll = this.collection?.id();
+    const db = this.database?.id();
+    if (coll) {
+      if (coll.length > 8) {
+        return coll.slice(0, 5) + "...." + options.title;
```
sevoku, line 150: An ellipsis has 3 dots and I'd recommend using the ellipsis character instead of actual 3 dots: "…" (Alt+0133)

now:
image

with an ellipsis (character):
image

But overall looks much nicer now.

JustinKol: Thanks, I just noticed the 4 vs 3 dots myself.

sevoku (dismissed):
LGTM, but we might want to look at some corner cases:
image
Sometimes I still see some unnecessary double ellipses (see second tab) and I'd apply some trimming before ellipsizing (see the space between "New" and "…" in the third tab)

src/Explorer/Tabs/TabsBase.ts
```diff
@@ -143,6 +142,26 @@ export default class TabsBase extends WaitsForTemplateViewModel {
...
+        return coll + "." + options.title;
+      }
+    } else if (db) {
+      if (db.length > 8) {
+        return db.slice(0, 5) + "…" + options.title;
+      } else {
+        return db + "." + options.title;
+      }
+    } else {
+      return "";
```
jawelton74, line 161: Should this return options.title if coll and db are undefined ?

JustinKol: You are correct. Fixed.

jawelton74 (approved):

sevoku (approved):

PR #1779, opened 2024-03-27
Remove notebooks UI

sindhuba:
This change removes the notebooks component UI in Portal Frontend as part of the feature deprecation.
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1779?feature.someFeatureFlagYouMightNeed=true)


src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.test.ts
```diff
@@ -368,62 +366,6 @@ describe("CommandBarComponentButtonFactory tests", () => {
     });
   });
 
-  describe("GitHub buttons", () => {
```
sevoku, line 371: Why are we commenting out the other changes but fully removing this one?

sevoku (changes requested):
It'd be good to add written comments to each change, about why we comment out the code (as simple as "//deprecated notebooks UI"). And also either comment out the code as whole or remove it, mixing is confusing when reviewing code afterwards (without the need to read git history)

src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.test.ts
sindhuba: I got rid of the tests but in the code - I will ultimately check to see if there are any dependencies/references to the commented portion so I can remove them easily all together. So, it'd be better for me to track it this way.

sindhuba:
> It'd be good to add written comments to each change, about why we comment out the code (as simple as "//deprecated notebooks UI"). And also either comment out the code as whole or remove it, mixing is confusing when reviewing code afterwards (without the need to read git history)

Got it - removed all the code to avoid confusion in the future. Thanks.

jawelton74 (approved):

sevoku (commented):
Hhhm, thnking twice... shouldn't we remove the actual notebooks code as well? Isn't it all dead code now sitting in the codebase?

PR #1780, opened 2024-03-27
fix opening collections

sunghyunkang1111:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1780?feature.someFeatureFlagYouMightNeed=true)

**Description**
The open database or collections action will not work if the action is fired too early, before databases are loaded.
OR no collections are loaded.  This is only apparent when the actions are triggered from portal such as ExplorerContracts.ActionContracts.OpenCollectionTab
e.g clicking on the collection row on the overview blade

So adding a subscription listener to the list of databases that calls subscription listener to collections.

**How Has This Been Tested?**
By clicking on the row of the collection item from overview blade, should open data explorer then the items tab

### DIFF
--- a/src/Explorer/OpenActions/OpenActions.tsx
+++ b/src/Explorer/OpenActions/OpenActions.tsx
@@ -49,104 +49,104 @@ function openCollectionTab(
       return;
     };
     const databasesSubscription = useDatabases.subscribe(databaseActionHandler, (state) => state.databases);
-  }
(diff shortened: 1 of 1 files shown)


vchske (approved):

asierisayas (approved):

sevoku (approved):
verified with Fabric

PR #1785, opened 2024-04-02
[Task #3061771] Correct render order issues on undo

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1785?feature.someFeatureFlagYouMightNeed=true)

Undo/redo is another feature that's built in to the Monaco editor so we _should_ be getting it for free. However, we aren't, so something must be going wrong. After a bunch of digging, the root cause of the problem is the way we're tracking the content of the editor itself. We hook a 'content changed' event on the editor, then use that to update the internal state tracking the content of the editor. When we re-render, we use that internal state to set the content of the editor. That causes a few problems:

1. Setting the content using `setValue` will destroy the undo/redo stack.
2. Unnecessarily setting the content of the editor to it's existing content will create new, confusing, undo stack entries.
3. If the ordering of events from the editor isn't what we expect, our internal state may get out of sync with the editor's actual content.

This PR fixes all three of those issues:

First, when we **do** need to externally change the content of the editor (such as when Copilot is used), we use `executeEdits` instead of `setValue`. This has the effect of emulating a "Select All, Paste New Content" action, and properly updates the undo/redo stack.

Second, in `EditorReact`, when comparing the content in our provided props to see if we should update the Monaco editor content, we compare against the actual current value (`this.editor.getModel().getValue()`) instead of our previous props (`previous.content`). This way, if the editor updates itself (such as in an undo operation), we're still comparing against the right value.

Finally, instead of hooking `this.editor.getModel().onDidChangeContent` to trigger content changed events, we hook `this.editor.onDidChangeModelContent`. Of the three changes, this is the most... guess-y 😅 . I did a bunch of tracking of event ordering and noticed that for some user operations, particularly **undo** and **redo** operations, the `this.editor.onDidChangeCursorSelection` event fires **before** `this.editor.getModel().onDidChangeContent`. This causes us to trigger the `onContentSelected` event **before** `onContentChanged`. The `QueryTabComponent` updates it's state on `onContentSelected`, which causes a re-render. However, it _has not seen the new content yet_, which means it tries to push the **old** content into the editor. This destroys the undo operation. However, I did find that `this.editor.onDidChangeModelContent` **does** consistently fire before `this.editor.onDidChangeCursorSelection`. Switching to that event fixes the issues with undo/redo without breaking other scenarios as far as I can find.

I'm not super excited to be this coupled to event ordering, but without a more significant dive in to reworking our data model for the editor, I think it's the best option to get the immediate benefit of making undo/redo work.

Here's a video demo showing the new changes in action!

https://github.com/Azure/cosmos-explorer/assets/7574/7dba991a-5255-42b9-9bbf-95bb45c33b7d

### DIFF
--- a/src/Explorer/Controls/Editor/EditorReact.tsx
+++ b/src/Explorer/Controls/Editor/EditorReact.tsx
@@ -48,15 +48,12 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
 
   public componentDidUpdate() {
     if (!this.editor) {
-      // Nothing special to do if the editor doesn't exist yet
       return;
     }
 
     const existingContent = this.editor.getModel().getValue();
 
-    // Compare the new content to the editor's existing content
     if (this.props.content !== existingContent) {
-      // If it's different, push an edit to the editor, respecting the undo stack
       this.editor.pushUndoStop();
       this.editor.executeEdits("", [
         {
--- a/src/Explorer/Controls/Editor/EditorReact.tsx
+++ b/src/Explorer/Controls/Editor/EditorReact.tsx
@@ -46,7 +46,7 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
     }, 100);
   }
 
-  public componentDidUpdate(previous: EditorReactProps) {
+  public componentDidUpdate() {
     if (!this.editor) {
       // Nothing special to do if the editor doesn't exist yet
       return;
--- a/src/Explorer/Controls/Editor/EditorReact.tsx
+++ b/src/Explorer/Controls/Editor/EditorReact.tsx
@@ -47,8 +47,22 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
   }
 
   public componentDidUpdate(previous: EditorReactProps) {
-    if (this.props.content !== previous.content) {
-      this.editor?.setValue(this.props.content);
+    if (!this.editor) {
+      // Nothing special to do if the editor doesn't exist yet
+      return;
+    }
+
+    const existingContent = this.editor.getModel().getValue();
+
+    // Compare the new content to the editor's existing content
+    if (this.props.content !== existingContent) {
+      // If it's different, push an edit to the editor, respecting the undo stack
+      this.editor.executeEdits("", [
+        {
+          range: this.editor.getModel().getFullModelRange(),
+          text: this.props.content,
+        },
+      ]);
     }
   }
 
@@ -71,9 +85,14 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
 
   protected configureEditor(editor: monaco.editor.IStandaloneCodeEditor) {
     this.editor = editor;
-    const queryEditorModel = this.editor.getModel();
     if (!this.props.isReadOnly && this.props.onContentChanged) {
-      queryEditorModel.onDidChangeContent(() => {
+      // Hooking the model's onDidChangeContent event because of some event ordering issues.
+      // If a single user input causes BOTH the editor content to change AND the cursor selection to change (which is likely),
+      // then there are some inconsistencies as to which event fires first.
+      // But the editor.onDidChangeModelContent event seems to always fire before the cursor selection event.
+      // (This is NOT true for the model's onDidChangeContent event, which sometimes fires after the cursor selection event.)
+      // If the cursor selection event fires first, then the calling component may re-render the component with old content, so we want to ensure the model content changed event always fires first.
+      this.editor.onDidChangeModelContent(() => {
         const queryEditorModel = this.editor.getModel();
         this.props.onContentChanged(queryEditorModel.getValue());
       });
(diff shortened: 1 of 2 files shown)


src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
```diff
@@ -544,7 +544,7 @@ export default class QueryTabComponent extends React.Component<IQueryTabComponen
     useCommandBar.getState().setContextButtons(this.getTabsButtons());
   }
 
-  public setEditorContent(): string {
+  public getEditorContent(): string {
```
analogrelay, line 547: I was getting confused by this method name since it _seems_ to be a `get` method rather than a `set` method, so I went ahead and renamed it. Happy to revert if I'm misunderstanding!

sunghyunkang1111 (approved):

sunghyunkang1111:
Quick question about "This has the effect of emulating a "Select All, Paste New Content" action, and properly updates the undo/redo stack".
What would be the best way to implement, if we wanted to programmatically change the content but everytime we do it, we append it to the existing content in editor? because Copilot is supposed to append everytime it generate a query but then the entire new content we set becomes an undo/redo stack item.

analogrelay:
> What would be the best way to implement, if we wanted to programmatically change the content but everytime we do it, we append it to the existing content in editor?

@sunghyunkang1111 That's definitely possible! The `executeEdits` method takes a list of "edits" to make, and each edit has the range of text to "replace" (which can be zero in length if you're just inserting) and the text to put at that range (which can be a different size than the source range). Right now, the range we pass is `[start_of_document  ... end_of_document]`, but we could also pass `[end_of_document ... end_of_document]` which would have the effect of inserting the text at the end.

jawelton74 (approved):

src/Explorer/Controls/Editor/EditorReact.tsx
```diff
@@ -46,9 +46,24 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
     }, 100);
   }
 
-  public componentDidUpdate(previous: EditorReactProps) {
-    if (this.props.content !== previous.content) {
-      this.editor?.setValue(this.props.content);
+  public componentDidUpdate() {
+    if (!this.editor) {
+      // Nothing special to do if the editor doesn't exist yet
```
vchske, line 51: This is a style preference, but I think this line is self-documenting and makes the comment here superfluous. Generally, DE adopted the attitude that a comment is a code smell unless an underlying component/logic is just hard to understand and too easy to mess up without a comment.

src/Explorer/Controls/Editor/EditorReact.tsx
```diff
@@ -46,9 +46,24 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
...
+      return;
+    }
+
+    const existingContent = this.editor.getModel().getValue();
+
+    // Compare the new content to the editor's existing content
```
vchske, line 57: Same for these comments

src/Explorer/Controls/Editor/EditorReact.tsx
```diff
@@ -71,9 +86,14 @@ export class EditorReact extends React.Component<EditorReactProps, EditorReactSt
 
   protected configureEditor(editor: monaco.editor.IStandaloneCodeEditor) {
     this.editor = editor;
-    const queryEditorModel = this.editor.getModel();
     if (!this.props.isReadOnly && this.props.onContentChanged) {
-      queryEditorModel.onDidChangeContent(() => {
+      // Hooking the model's onDidChangeContent event because of some event ordering issues.
```
vchske, line 90: This is when it is probably good to have a note as we are dependent on an undocumented order of events. So I like these comments here.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
```diff
@@ -496,13 +496,16 @@ export default class QueryTabComponent extends React.Component<IQueryTabComponen
   };
 
   public onChangeContent(newContent: string): void {
+    // The copilot store's active query takes precedence over the local state,
```
vchske, line 499: I would remove these comments as well.

vchske (approved):

src/Explorer/Controls/Editor/EditorReact.tsx
analogrelay: Yep, you're right, happy to tidy these up.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
analogrelay: I did think this was a useful one because the fact that the copilot store takes precedence is not clear when reading _this_ method. In fact, it took some debugging to realize that it was this state update ordering that was causing one of the problems I was seeing. So without the comment, someone might reorder the state updates and introduce a difficult-to-detect bug. Happy to remove it though, if folks still think it's unnecessary.

src/Explorer/Controls/Editor/EditorReact.tsx
analogrelay: Yep, also superfluous.

vchske (approved):

sunghyunkang1111 (approved):

jawelton74 (approved):

PR #1787, opened 2024-04-03
Post preview link as PR comment

analogrelay:
*Draft PR because the only way to test GitHub Actions is to open a PR* 😅

This change updates our GitHub Action to post the preview link as a PR comment, after the preview site is deployed. As a result, you no longer have to remember to update the link in the template, and it only gets posted when it will work (so you don't get Blob Not Found errors from clicking it too "early").

If you need to set feature flags in the preview link, there are two options:

1. I believe contributors can edit the comment GitHub Actions posted, so you can add the feature flag there
2. You can also always still post your own link

Because this is now automated, I removed the PR template, but I'm happy to keep it around if folks want it. Using a comment means the link appears down in the conversation instead of up in the description, which may not be ideal, but I think automating it is very helpful. Editing the PR description is _much harder_ to automate (though I think it's possible).


analogrelay:
Ah darn, this requires a PAT. You can't post a comment with the standard GitHub Actions token :(

PR #1788, opened 2024-04-03
add support for react dev tools to the editor

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1788?feature.someFeatureFlagYouMightNeed=true)

I was going to use the [React Dev Tools extension](https://react.dev/learn/react-developer-tools) to get some information about the Components hierarchy in the explorer. Unfortunately, since we host the explorer inside an iframe, it doesn't Just Work ™️ . To make it work, [we need to inject a little code to copy the `__REACT_DEVTOOLS_GLOBAL_HOOK__ ` value from the parent frame to the child frame](https://github.com/facebook/react-devtools/issues/76#issuecomment-128091900). This has to be done _before_ React is loaded, so it takes a little bit of trickery.

I noticed that this code already exists in `ReactDevTools.ts`, but it isn't imported anywhere. So, I've added it to `Main.tsx`, right at the top. This means it'll always run before the main explorer UI loads, even in production builds. That should be safe though since the React Dev Tools don't give any more information than a user couldn't already acquire through the standard browser dev tools. It's just easier to view.

Once this is in place, the dev tools show the full Component hierarchy!

Screenshot 2024-04-03 112003

### DIFF
--- a/src/Main.tsx
+++ b/src/Main.tsx
@@ -1,5 +1,5 @@
 // Import this first, to ensure that the dev tools hook is copied before React is loaded.
-import './ReactDevTools';
+import "./ReactDevTools";
 
 // CSS Dependencies
 import { initializeIcons, loadTheme } from "@fluentui/react";
--- a/src/Main.tsx
+++ b/src/Main.tsx
@@ -1,3 +1,6 @@
+// Import this first, to ensure that the dev tools hook is copied before React is loaded.
+import './ReactDevTools';
+
 // CSS Dependencies
 import { initializeIcons, loadTheme } from "@fluentui/react";
 import { QuickstartCarousel } from "Explorer/Quickstart/QuickstartCarousel";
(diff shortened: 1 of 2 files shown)


sevoku:
Wanted to test this but https://cosmos-explorer-preview.azurewebsites.net/pull/1788 (pls don't forget to update it in the description after creating the PR) doesn't load Data Explorer at all. Not sure if it's due to this change or some weird coincidence.

analogrelay:
@sevoku Yep, it's a problem with how the iframe acts when embedded in the portal. Looking in to it now. If possible, I'll just have it no-op when in the portal.

analogrelay:
I added a `try...catch` around the logic to copy the react global hook. That seems to be the most consistent way to detect cross-origin security issues like this. Because the Azure portal is coming from a different origin, the iframe can't read values from the parent. There's no way around that (it's an intentional, and good, security feature), so for this case, the dev tools just won't work.

analogrelay:
Confirmed that after that fix, data explorer loads in the portal

sunghyunkang1111 (approved):

sevoku (approved):

PR #1789, opened 2024-04-04
[Task 3061766] Global Keyboard Shortcuts, implemented through the Command Bar

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1789?feature.someFeatureFlagYouMightNeed=true)

**This PR introduces a new dependency on `tinykeys`** I'm not 100% clear on what process, if any, is needed for that but I'm happy to follow whatever is needed.

This PR adds Keyboard Shortcuts by allowing Command Bar Buttons to specify a keyboard shortcut that should be used to activate the command. Upon pressing the keyboard shortcut, the existing `onCommandClick` handler will be executed. The following shortcuts are bound:

| Shortcut | Context | Action |
| - | - | - |
| `Ctrl+J`/`Cmd+J` [^1] | A collection is selected | Open a new Query Tab for the selected collection |
| `Shift+Enter` | A query tab is active | Execute the query (or the selection, if there is one) in the current query tab |
| `Esc` | A query is executing | Cancel the currently-executing query |
| `Ctrl+S`/`Cmd+S` | A query tab is active | Open the Save Query side-panel |
| `Ctrl+O`/`Cmd+O` | Anywhere the Open Query commands appear | Open the Open Query side-panel |
| `Ctrl+Shift+O`/`Cmd+Shift+O` | Anywhere the Open Query commands appear | Open the Open Query from Disk side-panel |
| `Ctrl+S`/`Cmd+S` | Editing a Document, SPROC, UDF, or Trigger | Saves the relevant item |
| ~~`Esc`~~ | ~~Editing a Document, SPROC, UDF, or Trigger~~ | ~~Discard changes to the relevant item~~ **Removed for this PR, still looking at finding the right keybind for this** |
| `Shift+Enter` | Editing a SPROC | Open the Execute SPROC side panel |

The shortcuts are implemented through `tinykeys`. At the root of the app, we have a new `KeyboardShortcutRoot` component. In this component, we define a mapping of an "Action Name" to a list of shortcuts that can trigger it (multiple shortcuts can be registered for a single action, and any of the registered shortcuts will work). For example:

```typescript
const bindings: Record<KeyboardAction, string[]> = {
  [KeyboardAction.NEW_QUERY]: ["$mod+J"],
  [KeyboardAction.EXECUTE_ITEM]: ["Shift+Enter"],
  // ...
};
```

*Note: The syntax `$mod` in tinykeys is replaced with `Control` on Windows/Linux and `Command` on macOS, matching user expectations for these kinds of shortcuts*

The `KeyboardAction` enum defines all the actions we specify in the app. When the root component is mounted, it will register a `keydown` handler on it's root element that handles each of these shortcuts (using tinykeys). When any of the shortcuts are pressed, the named "Action" will be triggered.

However, the Actions are not always available and their handlers change based on the context of the app. For example, `EXECUTE_ITEM` on a Query Tab executes the query. But on a SPROC tab it opens the side panel to execute the SPROC. To handle this, we have a new Zustand global state store[^2] called `useKeyboardActionHandlers`. Whenever a keyboard action is triggered, we check a map of `KeyboardAction` to handler function. If there's a handler registered for an action, we run it (and it usually stops propagation of the event). If not, we no-op and the event will propagate as normal.

Right now, the only place that updates handlers is the `CommandBarComponentAdapter`. The buttons we create can designate an option `KeyboardAction` that can be used to trigger the button. **If** the button is present and enabled, the adapter registers it's handler. If it's not present or disabled, the handler is not registered. This allows us to use all our existing logic for managing those command bar buttons to also manage their keyboard actions. Once we have actions that don't map directly to command bar buttons, we'll need a bit of refactoring here, but I have an idea how that will work and don't want to overcomplicate things in this PR since all the shortcuts added here are from the command bar [^3].

This PR completes [ADO task #3061766](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/3061766/)

[^1]: It's not possible to bind Ctrl+N in the browser, so we can't use that.
[^2]: I could also use a React Context here, but it seems our existing practice is to use Zustand so I'm happy to do so.
[^3]: Basically, the problem is that the `CommandBarComponentAdaptor` replaces the entire handler map every time. This is necessary so it doesn't have to worry about removing old handlers when the button disappears. I intend to fix this by storing multiple sets of handler mappings and allowing each part of the application that provides handlers to replace it's set entirely without affecting the others.

### DIFF
--- a/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
+++ b/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
@@ -7,6 +7,7 @@ import {
   IDropdownStyles,
 } from "@fluentui/react";
 import { useQueryCopilot } from "hooks/useQueryCopilot";
+import { KeyboardHandlerMap } from "KeyboardShortcuts";
 import * as React from "react";
 import _ from "underscore";
 import ChevronDownIcon from "../../../../images/Chevron_down.svg";
@@ -233,3 +234,28 @@ export const createConnectionStatus = (container: Explorer, poolId: PoolIdType, 
     onRender: () => <ConnectionStatus container={container} poolId={poolId} />,
   };
 };
+
+export function createKeyboardHandlers(allButtons: CommandButtonComponentProps[]): KeyboardHandlerMap {
+  const handlers: KeyboardHandlerMap = {};
+
+  function createHandlers(buttons: CommandButtonComponentProps[]) {
+    buttons.forEach((button) => {
+      if(button.disabled !== true && button.keyboardAction) {
+        handlers[button.keyboardAction] = (e) => {
+          button.onCommandClick(e);
+
+          // If the handler is bound, it means the button is visible and enabled, so we should prevent the default action
+          return true;
+        }
+      }
+
+      if (button.children && button.children.length > 0) {
+        createHandlers(button.children);
+      }
+    });
+  }
+
+  createHandlers(allButtons);
+
+  return handlers;
+}
--- a/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
+++ b/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
@@ -240,7 +240,7 @@ export function createKeyboardHandlers(allButtons: CommandButtonComponentProps[]
 
   function createHandlers(buttons: CommandButtonComponentProps[]) {
     buttons.forEach((button) => {
-      if (button.disabled !== true && button.keyboardAction) {
+      if (!button.disabled && button.keyboardAction) {
         handlers[button.keyboardAction] = (e) => {
           button.onCommandClick(e);
 
--- a/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
+++ b/src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
@@ -240,13 +240,13 @@ export function createKeyboardHandlers(allButtons: CommandButtonComponentProps[]
 
   function createHandlers(buttons: CommandButtonComponentProps[]) {
     buttons.forEach((button) => {
-      if(button.disabled !== true && button.keyboardAction) {
+      if (button.disabled !== true && button.keyboardAction) {
         handlers[button.keyboardAction] = (e) => {
           button.onCommandClick(e);
 
           // If the handler is bound, it means the button is visible and enabled, so we should prevent the default action
           return true;
-        }
+        };
       }
 
       if (button.children && button.children.length > 0) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -106,6 +106,7 @@
         "styled-components": "5.0.1",
         "swr": "0.4.0",
         "terser-webpack-plugin": "5.3.9",
+        "tinykeys": "2.1.0",
         "underscore": "1.9.1",
         "utility-types": "3.10.0",
         "zustand": "3.5.0"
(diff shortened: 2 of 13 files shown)


vchske (approved):

vchske (approved):

jawelton74 (approved):

analogrelay:
Update: While experimenting with multi-key sequences (for example `alt+n p`, which would mean to press Alt+N, then release and press P), I've discovered a [pretty blocking bug in react-hotkeys](https://github.com/greena13/react-hotkeys/issues/219). I also noticed that react-hotkeys is no longer actively maintained, so we can't really expect a fix for that. That leaves us with two options:

1. Use a different library
2. Don't use multi-key sequences

Right now, our plans call for some actions to use multi-key sequences, so I'm doing a little investigation in to option 1. The [react-hotkeys-hook](https://www.npmjs.com/package/react-hotkeys-hook) library appears to be more actively maintained, with a release only 2 months ago, so I'm investigating that and will report back and update this PR thread. If it fixes our issues, taking on the new dependency feels like the right approach here to me.

sevoku (dismissed):
Good catch, blocking for now so this doesn't get merged by accident. Maybe let's convert back to draft.

analogrelay:
> Maybe let's convert back to draft.

Good idea, done.

Unfortunately, `react-hotkeys-hook` also doesn't support multi-key sequences. However, I've identified [tinykeys](https://www.npmjs.com/package/tinykeys) as a possible candidate for replacement. It requires a little more binding logic on our side (since it's not directly integrated with React) but since we generally just want global keyboard shortcuts, it's pretty straightforward.

analogrelay:
Ok, this is ready to review again. Instead of `react-hotkeys`, I'm using `tinykeys`. See the PR description for an update on how I've set things up. The table of keyboard shortcuts hasn't changed, but the rest is fairly new.

analogrelay:
Small update. During testing I noticed that some situations cause focus to go to the body. Specifically, when clicking in the blank space on the Home tab.

You can test this yourself by running this in the browser console in DE:

```javascript
const handle = setInterval(() => console.log("active element", document.querySelector("#explorerMenu").contentDocument.activeElement), 1000);
```

This will print the active focused element in the DE IFrame every second. Then click on blank space in the Home tab and you'll see this:

image

*You can stop the printing with `clearInterval(handle)` or just F5 to refresh the page*

Currently, we bind keyboard shortcuts to the root element in the React component tree, which is a descendant of the `body`. So when focus is on the body, keyboard events go straight there without bubbling up and we miss them.

Since we want our keyboard shortcuts to be global (we enable/disable their _behavior_ based on context, but the events can and should fire no matter your focus), we might as well bind them to the body.

src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
```diff
@@ -233,3 +234,28 @@ export const createConnectionStatus = (container: Explorer, poolId: PoolIdType,
     onRender: () => <ConnectionStatus container={container} poolId={poolId} />,
   };
 };
+
+export function createKeyboardHandlers(allButtons: CommandButtonComponentProps[]): KeyboardHandlerMap {
+  const handlers: KeyboardHandlerMap = {};
+
+  function createHandlers(buttons: CommandButtonComponentProps[]) {
+    buttons.forEach((button) => {
+      if (button.disabled !== true && button.keyboardAction) {
```
languy, line 243: nit:
```suggestion
      if (!button.disabled && button.keyboardAction) {
```

languy (dismissed):
Nicely done!

languy:
Question: this doesn't suppress all the accessibility shortcuts that are already implemented, right? mainly tab to navigate through the elements and spacebar to click on the buttons.

src/Explorer/Menus/CommandBar/CommandBarUtil.tsx
analogrelay: I initially had this but was wary of the impact of `disabled` being undefined. But you're right, it's fine because we want `undefined` to be `false` (as it usually is).

analogrelay:
@languy Good catch. It should not impact those, but I'll do a quick test to validate that assumption.

analogrelay:
@languy Confirmed, those accessibility shortcuts _are unaffected_ and continue to work as expected. As long as we don't try to bind something to _just_ `Tab` or `Spacebar`, they will continue to work.

jawelton74 (dismissed):

JustinKol:
Could we add the Save hotkey to the saveNewDocumentButton in the DocumentsTab?

analogrelay:
@JustinKol Yeah, I noticed that was missing and actually have it in a follow-up branch. I can pull it in here.

JustinKol (approved):

PR #1817, opened 2024-04-23
Add button (and keyboard shortcut) to download query

analogrelay:

Adds a new button to the toolbar in the Query Editor, "Download Query". Looks like we already had an SVG for an icon for this 🎉 (specifically, it's used to download "Per-partition query metrics" on the Query Stats tab of the results panel)

image

Pressing this button, or pressing the keyboard shortcut `Ctrl+Shift+S`, will trigger the browser to download the text of the current query window:

image


src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
```diff
@@ -225,6 +226,20 @@ export default class QueryTabComponent extends React.Component<IQueryTabComponen
     }
   };
 
+  public onDownloadQueryClick = (): void => {
+    const text = this.getCurrentEditorQuery();
+    const queryFile = new File([text], `SavedQuery.sql`, { type: "text/plain" });
```
jawelton74, line 231: Suggest using a .txt file extension. The query language depends on the account API, and the open file from disk dialog defaults to .txt files.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
```diff
@@ -419,6 +434,16 @@ export default class QueryTabComponent extends React.Component<IQueryTabComponen
...
     }
 
+    buttons.push({
+      iconSrc: DownloadQueryIcon,
+      iconAlt: "Download Query",
+      keyboardAction: KeyboardAction.DOWNLOAD_ITEM,
+      onCommandClick: this.onDownloadQueryClick,
+      commandButtonLabel: "Download Query",
+      ariaLabel: "Download Query",
+      hasPopup: false,
```
jawelton74, line 444: Can we disable the button if query window is empty ?

analogrelay: Yep, that should be possible.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
analogrelay: Good point. I had it as `txt` originally but then noticed we talk about it as a "SQL Query" else where. I think `txt` does make more sense here though, especially given that we default the open dialog to that.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
analogrelay: Done. What I actually did was tie the Download button in to the existing Save button state. That button wasn't disabled when the query was empty though, so I went ahead and added that logic since it seems appropriate to me. So now **both** the Save Query and Download Query buttons are disabled when the query editor is completely empty.

src/Explorer/Tabs/QueryTab/QueryTabComponent.tsx
analogrelay: Done!

jawelton74 (approved):

sevoku (approved):

PR #1818, opened 2024-04-24
[Accessibility-3100036]: Close button is nested under 'Home' tab control under 'Data Explorer' pane.

satya07sb:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1818?feature.someFeatureFlagYouMightNeed=true)

In the previous implementation within the 'Tabs' section, we provided full clickable access, allowing users to click on tabs containing a close button.

In this update, we've enhanced the user experience by differentiating the 'Home' section and the close button section. We achieved this by enclosing the 'Home' section within its own wrap, while the close button remains separate. As a result, all Home tab controls now exclusively belong to the 'Home' section, providing a more streamlined navigation experience. Additionally, the close button maintains its distinct position, ensuring clarity and intuitive interaction for users.


asierisayas (dismissed):

sevoku (changes requested):
Unfortunately this breaks the design in Fabric:
image

Fabric specific style is here: https://github.com/Azure/cosmos-explorer/blob/77883e7b6693bc7abd31d68339c70f19638691c0/less/documentDBFabric.less#L69-L118

sevoku (dismissed):
This works, thanks!

sunghyunkang1111 (approved):

vchske (changes requested):
This breaks being able to close at least one tab (Scale and Settings). Notice how far out the close button is:
image
Although you can't see the pointer I am hovering over where the close button should be here:
image
But because the close button was outside the tab, the tab to the right (Items) now covers the close button.

vchske (dismissed):
Change fixed the close button. Thanks!

One more thing, when using the tab key to navigate the tabs, when I tab to, say "....Items", I can hit the spacebar and the Items tab is now focused. But when I tab back to the "Home" tab and I hit the spacebar, the Home tab is not brought to focus, even though the text of tab, "Home" is underlined:
image
I'm not sure if this is problematic enough to hold this checkin but wanted to point it out.

satya07sb:
> Change fixed the close button. Thanks!
> 
> One more thing, when using the tab key to navigate the tabs, when I tab to, say "....Items", I can hit the spacebar and the Items tab is now focused. But when I tab back to the "Home" tab and I hit the spacebar, the Home tab is not brought to focus, even though the text of tab, "Home" is underlined: image I'm not sure if this is problematic enough to hold this checkin but wanted to point it out.

You are correct, but this is an existing issue. The code change is not affecting this problem, and it is out of scope for this ticket. However, it will not take much effort to fix it, so let me address it as well.

vchske (approved):

sevoku (approved):
Retested with Fabric and everything looks fine.

PR #1825, opened 2024-04-26
Update UPAPI for Dedicated Gateway in Self Serve Portal

tarazou9:
# Description
This is for updating today's UPAPI with the new one for Dedicated Gateway in Self Serve Portal. 

# Changes
Replacing today's UPAPI calls with two API calls. 
- First API call is to get Offering Ids for each sku by service name and location. Here is the reference link: https://learn.microsoft.com/en-us/rest/api/marketplacecatalog/dataplane/skus/list?view=rest-marketplacecatalog-dataplane-2023-05-01-preview&tabs=HTTP
- Second API call is making to the new UPAPI to use Offering Id for each sku to retrieve prices. 

# How Has This Been Tested?
Test locally from end to end.


asierisayas (dismissed):

src/Common/Constants.ts
```diff
@@ -249,6 +249,7 @@ export class HttpHeaders {
   public static partitionKey: string = "x-ms-documentdb-partitionkey";
   public static migrateOfferToManualThroughput: string = "x-ms-cosmos-migrate-offer-to-manual-throughput";
   public static migrateOfferToAutopilot: string = "x-ms-cosmos-migrate-offer-to-autopilot";
+  public static xAPIKey: string = "X-API-Key";
```
jawelton74, line 252: should the value be lower case ?

src/ConfigContext.ts
```diff
@@ -42,6 +42,9 @@ export interface ConfigContext {
   ARM_API_VERSION: string;
   GRAPH_ENDPOINT: string;
   GRAPH_API_VERSION: string;
+  CATALOG_ENDPOINT: string;
```
jawelton74, line 45: Please add a comment about what this endpoint is

src/Common/Constants.ts
tarazou9: @jawelton74 Here is the API document link, it looks like it's using "X-API-Key" in the doc. https://learn.microsoft.com/en-us/rest/api/marketplacecatalog/dataplane/skus/list?view=rest-marketplacecatalog-dataplane-2023-05-01-preview&tabs=HTTP#request-headers

src/ConfigContext.ts
tarazou9: @jawelton74 Comment added, thank you for pointing this out!

jawelton74 (approved):

asierisayas (approved):

PR #1826, opened 2024-04-28
add capacityMode

sunghyunkang1111:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1826?feature.someFeatureFlagYouMightNeed=true)

Description
with a breaking change in RP for api version 2024-05-15-preview, contracts need to be updated in the portal.
Adding capacityMode property in database account resource where the account capacitymode is defined rather than in the capabilities.
Currently standalone DE is using fixed 2024-02-15-preview version,
the problem is when in-portal DE loads the databaseaccount from the portal with the latest API and passed into the DE, which breaks the contract for serverless accounts. 

How Has This Been Tested?
Tested locally with sideloaded portal where the databaseaccount is fetched from 2024-05-15-preview API

### DIFF
--- a/src/Contracts/DataModels.ts
+++ b/src/Contracts/DataModels.ts
@@ -1,4 +1,4 @@
-import { ConnectionStatusType, ContainerStatusType } from "../Common/Constants";
+import { CapacityMode, ConnectionStatusType, ContainerStatusType } from "../Common/Constants";
 
 export interface ArmEntity {
   id: string;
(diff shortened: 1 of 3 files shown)


src/Contracts/DataModels.ts
```diff
@@ -35,6 +35,7 @@ export interface DatabaseAccountExtendedProperties {
   ipRules?: IpRule[];
   privateEndpointConnections?: unknown[];
   capacity?: { totalThroughputLimit: number };
+  capacityMode?: string;
```
languy, line 38: Can this be of type `CapacityMode`?

jawelton74 (approved):

PR #1851, opened 2024-05-30
Implement ctrl-shift click to select multiple documents

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1851)

This change implements the ability to select multiple documents by clicking with Ctrl and/or Shift. It will select/deselect individual items while keeping the other currently selected items, or select/deselect a range of documents. The behavior follows Windows File Explorer.

The mouse cursor is also changed to a pointer.

For single selection toggling, it's Command-click on Mac and Ctrl-Click on other platforms.

When selection is disabled (Fabric read-only), revert to just click behavior.

2024-05-30 shift ctrl selection


src/Explorer/Tabs/DocumentsTabV2/DocumentsTableComponent.tsx
```diff
@@ -121,24 +123,58 @@ export const DocumentsTableComponent: React.FC<IDocumentsTableComponentProps> =
     [columnHeaders],
   );
 
-  const onIdClicked = useCallback((index: number) => onSelectedRowsChange(new Set([index])), [onSelectedRowsChange]);
+  const [selectionStartIndex, setSelectionStartIndex] = React.useState<number>(undefined);
+  const onTableCellClicked = useCallback(
+    (e: React.MouseEvent, index: number) => {
+      const result = selectionHelper(selectedRows as Set<number>, index, e.shiftKey, e.ctrlKey, selectionStartIndex);
```
sevoku, line 129: For MacOS it should be `e.metaKey` instead of `e.ctrlKey` if I'm not mistaken.

sevoku (commented):
LGTM works like a charm, but let's don't forget about MacOS.

src/Explorer/Tabs/DocumentsTabV2/DocumentsTableComponent.tsx
languy: That is correct. Turns out we already had code to detect platform and check ctrl vs meta for Tables API which I moved to a more common place.

sevoku (approved):
Nice, verified on MacOS as well.

jawelton74 (approved):

PR #1854, opened 2024-06-03
[accessibility-2280341]:[Keyboard Navigation - Azure Cosmos DB - Graph]: Keyboard focus order is not logical after selecting any tab control.

satya07sb:
Description:
Previously, navigating through tab controls posed challenges with illogical focus order after selection. We've addressed this issue by implementing Fluent Tab (Pivot). This enhancement ensures smooth navigation for all users, regardless of their preferred method of interaction. With these updates, navigating through Azure Cosmos DB - Graph is now more intuitive and efficient than ever before.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1854?feature.someFeatureFlagYouMightNeed=true)

https://github.com/Azure/cosmos-explorer/assets/107645008/dded84ce-0d74-4e6e-bae9-b7ef362f08cf


src/Common/tabNavigationUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param event The keyboard event triggering the tab navigation.
+ * @param currentTabIndex The index of the currently focused tab.
+ * @param tabs An array of tab objects representing the tabs in the component.
+ * @param navigateTab A function to navigate to a specific tab by index.
+ * @param activateTab A function to activate the currently focused tab basically onClick function.
+ */
+import { NormalizedEventKey } from "Common/Constants";
+import { Tab } from "Explorer/Controls/Tabs/TabComponent";
+
+export const handleTabNavigation = (
```
sevoku, line 15: This file should probably be in `src/Explorer/Controls/Tabs/` since it's specific to that component.

sevoku (commented):
When an inactive tab is focused, pressing SHIFT+TAB will move focus to the current active Tab rather then to the previous component.
image
Focus will go to "JSON" here.

Also wouldn't it make more sense to replace the custom tabs with Fluent tabs (same as we did for Query results) and get the right behavior initially?

analogrelay:
> Also wouldn't it make more sense to replace the custom tabs with Fluent tabs (same as we did for Query results) and get the right behavior initially?

I think we could probably migrate this over to Fluent tabs without too much effort, but it's not something currently on my radar. In general, the Fluent controls have much better default accessibility features than our custom stuff, so the more we can standardize on those, the better.

src/Explorer/Controls/Tabs/TabComponent.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { AccessibleElement } from "../../Controls/AccessibleElement/AccessibleElement";
+import { Pivot, PivotItem } from "@fluentui/react/lib/Pivot";
```
analogrelay, line 2:
```suggestion
import { Pivot, PivotItem } from "@fluentui/react";
```

You should be able to import these from the root module of `@fluentui/react`, which is better since it doesn't rely on the internal file structure of the module.

analogrelay (approved):
I'm good with this 👍🏻 . We're moving several things over to Fluent UI 9 (the `@fluentui/react-components` library) but there are significant differences there that would require a little re-working of the tabs component and I think it's fine to stick with Fluent UI 8 for this initially since it gets us the accessibility benefits we need without a major rewrite of the page.

analogrelay (approved):
Looks good to me!

sevoku (approved):
Works fine!

PR #1856, opened 2024-06-03
Add form and validation for vector search

sunghyunkang1111:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1856?feature.someFeatureFlagYouMightNeed=true)

This PR addresses the concern for invalid form submission for vector search parameters.
So adding a UI form with validations.

image


analogrelay (commented):
First, some small suggestions as comments/suggestions inline, non-blocking.

Initially, not having much background on the vector search feature, I tried to configure Vector store policies on a new container and hit a few confusing things. I was able to work out what I was doing wrong, but I think there are some things we could do to clarify the experience a bit:

1. It felt a little awkward to have the vector policies and the indexing policies as separate UI. From my understanding of the docs, every vector **property** can have _zero or one_ **indexing policies**. So rather than managing them as separate items in separate lists, it seems like it would be better if they were managed in a single list of Vector properties. Something like this:

image

The "Index Type" dropdown would contain all the index types *and* the value 'none' (or something similar) to indicate that not index should be applied to that property. Using a single list would also solve some of the jarring validations (like flagging the dimension count up in the policies list when you select an index type you've selected requires a smaller dimension count).
 
2. I didn't realize that you can't have two indexing policies for a single path, because our UI doesn't do anything to block that, or validate it. I had to wait for the save operation to fail before getting feedback that I had done something wrong (at which point I saw in the docs that you can't have two different indexes for a single path). We could detect this case and show an error. Taking suggestion 1 above would make this irrelevant though, since they wouldn't be separate lists that could get out of sync. If there are other reasons to keep them as two lists, we should put some client-side validation in place (or even better, limit the Paths you can select to only the ones that don't already have indexing policies).

sunghyunkang1111:
> First, some small suggestions as comments/suggestions inline, non-blocking.
> 
> Initially, not having much background on the vector search feature, I tried to configure Vector store policies on a new container and hit a few confusing things. I was able to work out what I was doing wrong, but I think there are some things we could do to clarify the experience a bit:
> 
> 1. It felt a little awkward to have the vector policies and the indexing policies as separate UI. From my understanding of the docs, every vector **property** can have _zero or one_ **indexing policies**. So rather than managing them as separate items in separate lists, it seems like it would be better if they were managed in a single list of Vector properties. Something like this:
> 
> image
> The "Index Type" dropdown would contain all the index types _and_ the value 'none' (or something similar) to indicate that not index should be applied to that property. Using a single list would also solve some of the jarring validations (like flagging the dimension count up in the policies list when you select an index type you've selected requires a smaller dimension count).   2. I didn't realize that you can't have two indexing policies for a single path, because our UI doesn't do anything to block that, or validate it. I had to wait for the save operation to fail before getting feedback that I had done something wrong (at which point I saw in the docs that you can't have two different indexes for a single path). We could detect this case and show an error. Taking suggestion 1 above would make this irrelevant though, since they wouldn't be separate lists that could get out of sync. If there are other reasons to keep them as two lists, we should put some client-side validation in place (or even better, limit the Paths you can select to only the ones that don't already have indexing policies).

I think this is awesome insight. Never occured to be in one form since we had two separate free form monaco editor before. Will address this and also run it by James 👍

asierisayas (approved):

jawelton74 (approved):

vchske (approved):

analogrelay (approved):

PR #1857, opened 2024-06-03
Update Playwright, improve E2E test reliability, add scripts to deploy test resources

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1857?feature.someFeatureFlagYouMightNeed=true)

This is a set of a few changes to make the E2E test experience a little smoother:

1. There are now Bicep files describing test accounts suitable for running all the tests. You can deploy these automatically using the `test\resources\deploy.ps1` script. More details are provided in the `test\README.md` file.
2. The tests are no longer hard-coded to certain resources. You can use `DE_TEST_` environment variables to point the tests at different resources.
3. End-To-End tests are more reliable. They use `data-test` attributes more, and have a set of helpers to easily find context menus, expand tree nodes, etc. In general, I've found them to be much more reliable, though there's always still a risk of flakiness with this kind of E2E test.
4. Updated to the latest version of Playwright, which includes a built-in test framework instead of using jest-playwright. This version also provides a great UI for running/debugging tests:

image

5. Expand Playwright testing to include all three major browser engines: Chromium, Webkit, Firefox.

6. Updated the CI build to use Playwright's built-in sharding and report merging support. We split the test run across 8 shards, and Playwright automatically distributes tests across each shard in a deterministic way. Then, a final GitHub Actions job collects the reports and merges them into a single artifact. This is based on the [guide in the Playwright docs](https://playwright.dev/docs/test-sharding#github-actions-example).

7. Removed the notebooks-related tests as they aren't currently being run (and the feature is scheduled for removal I believe)

8. A few small fixes to the app itself. Restructuring some `data-test` attributes, and adding a few new ones. Also, updated language on the Confirm Database Delete side panel to use the API-specific term for a database ("keyspace", etc.) instead of always saying "database" (aligning it better with Confirm Collection Delete)


languy (approved):

jawelton74 (approved):
Awesome !!!

PR #1861, opened 2024-06-05
stop tree nodes from expanding to full height, causing overlap

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1861?feature.someFeatureFlagYouMightNeed=true)

This fixes an issue discovered after deploying the new resource tree (#1841) to MPAC. The tree nodes were set to `height: 100%`, and the containing Fluent Provider was set to `overflow: hidden`. A FluentUI Tree is a nested set of `Tree`s containing `TreeItem`s which contain `Tree`s containing `TreeItem`s, etc. Setting a `TreeItem` to `height: 100%` forced them to expand to fit all their children. Removing this avoids the overlap. However, because the containing Fluent Provider is `overflow: hidden`, doing that means overflow off the bottom is truncated, so we just need to switch the container to `overflow: auto` to get scrollbars back.

Here's the _broken_ behavior in `master`. I'm shrinking the browser quite a bit to make it happen, but it gets much worse as you have more and more items in a tree node, so if you have a DB with a lot more containers (I don't) then this would happen at "normal" vertical window sizes too:

https://github.com/Azure/cosmos-explorer/assets/7574/16d03f8c-0208-439b-b51d-731c3d91d30d

With these changes, there's no overlap in similar situations:

https://github.com/Azure/cosmos-explorer/assets/7574/6a0ebd4a-e9b4-437c-bf3d-e94b5c3006b0

I have plans to build an E2E test to protect against other overlap issues, but it's blocked on #1857 so I'll work on that separately and bring it in a separate PR.

I've also verified this when hosted in the portal, and it's fixed there too.

### DIFF
--- a/src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
+++ b/src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
@@ -150,7 +150,6 @@ export const TreeNodeComponent: React.FC<TreeNodeComponentProps> = ({
       data-test={`TreeNodeContainer:${treeNodeId}`}
       value={treeNodeId}
       itemType={isBranch ? "branch" : "leaf"}
-      style={{ height: "100%" }}
       onOpenChange={onOpenChange}
     >
       <TreeItemLayout
--- a/src/Explorer/Controls/TreeComponent/__snapshots__/TreeNodeComponent.test.tsx.snap
+++ b/src/Explorer/Controls/TreeComponent/__snapshots__/TreeNodeComponent.test.tsx.snap
@@ -5,11 +5,6 @@ exports[`TreeNodeComponent does not render children if the node is loading 1`] =
   data-test="TreeNodeContainer:root"
   itemType="branch"
   onOpenChange={[Function]}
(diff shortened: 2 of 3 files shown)


analogrelay:
> Setting a TreeItem to height: 100% forced them to expand to fit all their children.

As a side note, you can see this happening in my first video. Note that the `DATA` node (which is the root node) remains on screen throughout, and the scrolling is happening _within_ that node. That's because the `DATA` node is stretched to 100% height. This effect then _cascades_ (CSS pun absolutely intended 😅 ) across all the child nodes, which also stretch to 100% height.

analogrelay:
Oop, hang tight, snapshots need to be updated.

analogrelay:
Ok, test snapshots have been updated. This is ready for review now!

jawelton74 (dismissed):

analogrelay:
Rebased on the latest master and resolved a conflict. No other changes.

languy (approved):

sevoku (approved):

PR #1865, opened 2024-06-06
Redesign resource tree

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1865?feature.someFeatureFlagYouMightNeed=true)

This PR takes on redesigning the Resource Tree, and making some small design nudges to the new Documents Tab to line up with those changes.

Here's a quick preview, but I recommend opening the preview site up and playing around!

image

There's a slight difference in the layout if you enable the Sample Database for Copilot, since we want to clearly mark _your_ data and the sample data. This is similar to what we do in the existing design:

image

This PR **does not** include reworking the Command Bar to move it down into the tab bar. I wanted to split the redesign work somewhere so that it wasn't too big in a single PR. I think it's reasonable to roll this out all the way to production even though it doesn't completely match our intended design, but if we want to find a way to hold/stage it until the rest of the design work is done, I'm OK with that too. A separate PR will come after this to move the command bar down into the tab body.

A rough overview of the changes:

1. Moved the Home button off the command bar and to the top node in the resource tree
2. Moved the New Collection/Database buttons off the command bar and in to a new "Global Commands" split button just above the resource tree
3. Redesigned the Resource Tree according to design mockups
4. Refactored tree-related styles into a `Styles.ts` file.
5. Trimmed the top padding on the tab bar so that it had the height of a single "Layout Row" (see below)
6. Adjusted the height of the MessageBars used above the Tab bar to match the height of a "Layout Row"
7. Refactored the Documents Tab to use [Griffel styles](https://griffel.js.org/) (the `makeStyles` and `className={styles.className}` pattern) to align with how Fluent UI handles theming and styles.
8. Adjusted the Filter row in the Documents Tab so that it had the height of a single "Layout Row"
9. Adjusted the table header and row heights to match the "Layout Row" height
10. Moved the "Edit Filter" button to the far right (Open Question: This wasn't entirely clear in the design mockups, we can easily move it back to being just to the right of the existing query if we want)
11. Adjusted the "Load More" row at the bottom to match the "Layout Row" height and added a top border to line up with the items in the table.

#### Layout Rows

The design mockups feature a lot of things aligned on a grid. Since the components are spread across a lot of CSS layout contexts, they can't easily be put in a single CSS grid. So, there's a `LayoutConstants` object that stores the height of a "Layout Row" (`36px`, from the design mockup). A lot of elements use this height as a basis for their own height. For example, the Global Commands box is `2 * LayoutConstants.layoutRowSize` pixels high, aligning it with rows in the tab bar and documents tab.

#### TODO

A few things still need to be done before this can be undrafted and ready for final review.

* [x] Resolve conflicts with master
* [x] I'm getting occasional `ResizeObserver loop completed with undelivered notifications.` errors on the Documents Tab. I need to investigate that further.
* [x] Test Snapshots are completely broken and need to be updated
* [x] Do full round of verification across API types and Hosted, Platform, Emulator, and Fabric hosting environments.


asierisayas (approved):

src/Explorer/Tree/treeNodeUtil.ts
```diff
@@ -234,8 +227,8 @@ export const buildCollectionNode = (
     },
     onExpanded: async () => {
       // Rewritten version of expandCollapseCollection
-      useSelectedNode.getState().setSelectedNode(collection);
-      useCommandBar.getState().setContextButtons([]);
+      // useSelectedNode.getState().setSelectedNode(collection);
+      // useCommandBar.getState().setContextButtons([]);
```
languy, line 231: Intentional? Remove?

languy (approved):

package.json
```diff
@@ -48,6 +48,7 @@
     "@types/node-fetch": "2.5.7",
     "@uiw/react-split": "5.9.3",
     "@xmldom/xmldom": "0.7.13",
+    "allotment": "1.20.2",
```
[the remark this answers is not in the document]
analogrelay: Yep, I'll do that

src/Explorer/Tree/treeNodeUtil.ts
analogrelay: Yep, this is intentional because it changes the target collection for New SQL Query/etc. when you expand a node even though the tree selection does not change. I want to do a quick second pass on that behavior before actually deleting it though. I'll do that and then delete these lines and push another iteration.

analogrelay: Actually, in re-testing this behavior, I restored these lines. If we don't change the selected container on expand, the user has no way to change the selected container without opening a tab like Items or Settings.

sunghyunkang1111 (approved):

languy (approved):

PR #1876, opened 2024-06-19
[accessibility-2262594]:[Keyboard Navigation - Azure Cosmos DB - Data…

satya07sb:
… Explorer]: Keyboard focus is not retaining back to 'more' button after closing 'Delete container' dialog.

### DIFF
--- a/src/Explorer/ContextMenuButtonFactory.tsx
+++ b/src/Explorer/ContextMenuButtonFactory.tsx
@@ -52,12 +52,15 @@ export const createDatabaseContextMenu = (container: Explorer, databaseId: strin
   if (userContext.apiType !== "Tables" || userContext.features.enableSDKoperations) {
     items.push({
       iconSrc: DeleteDatabaseIcon,
-      onClick: () =>
+      onClick: (lastFocusedElement?: React.RefObject<HTMLElement>) =>
         useSidePanel
           .getState()
           .openSidePanel(
             "Delete " + getDatabaseName(),
-            <DeleteDatabaseConfirmationPanel refreshDatabases={() => container.refreshAllDatabases()} />,
+            <DeleteDatabaseConfirmationPanel
+              lastFocusedElement={lastFocusedElement}
+              refreshDatabases={() => container.refreshAllDatabases()}
+            />,
           ),
       label: `Delete ${getDatabaseName()}`,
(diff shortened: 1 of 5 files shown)


vchske (approved):

sevoku (approved):
I think this is obsolete with https://github.com/Azure/cosmos-explorer/pull/1865 and unfortunately I see the same issue there. cc @analogrelay 

But maybe let's merge this still, in case we need the old tree for some reason @jawelton74 ?

jawelton74:
> I think this is obsolete with #1865 and unfortunately I see the same issue there. cc @analogrelay
> 
> But maybe let's merge this still, in case we need the old tree for some reason @jawelton74 ?

I'm fine with still merging this, thanks !

analogrelay:
> I think this is obsolete with https://github.com/Azure/cosmos-explorer/pull/1865 and unfortunately I see the same issue there. cc @analogrelay

I'll take a look at that today, and see if I can make a similar fix in the new tree.

analogrelay:
Actually, just took another look at this PR. The fix here _is_ in the new tree, it was just done before the redesign (which was largely cosmetic, since I had already ported the tree to Fluent UI 9 earlier). If this branch is updated to latest `master` and conflicts are resolved, it should resolve the issue in the latest version of the tree. The conflicts look fairly straightforward (mostly just merging some of these changes with some style changes made to the same elements).

One question I have though, is this limited to the Delete Container dialog? What about other dialogs like New Container (from the context menu for a Database)? Do we perhaps need a more holistic approach to handle this in `useSidePanel` for _any_ time that a side-panel is opened and then dismissed?

satya07sb:
> Actually, just took another look at this PR. The fix here _is_ in the new tree, it was just done before the redesign (which was largely cosmetic, since I had already ported the tree to Fluent UI 9 earlier). If this branch is updated to latest `master` and conflicts are resolved, it should resolve the issue in the latest version of the tree. The conflicts look fairly straightforward (mostly just merging some of these changes with some style changes made to the same elements).
> 
> One question I have though, is this limited to the Delete Container dialog? What about other dialogs like New Container (from the context menu for a Database)? Do we perhaps need a more holistic approach to handle this in `useSidePanel` for _any_ time that a side-panel is opened and then dismissed?

sure, I will look into it.

PR #1887, opened 2024-06-27
Error rendering improvements

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1887?feature.someFeatureFlagYouMightNeed=true)

This PR improves how we display errors in the query tab. Here's a quick preview:

image

We're parsing errors out of the message returned by the CosmosDB backend, and then using that information to render a compiler-style error list including the error code, line number and message. In addition, we use that information to render a "red squiggly" under the text referenced by the error, which has a tooltip showing the error itself. You can also use standard Monaco keyboard shortcuts to navigate between these "problem" markers (F8 / Shift+F8).

**A note about how we are getting rich error details**: The REST API returns a raw string with the error message. For example:

```
Message: {"errors":[{"severity":"Error","location":{"start":9,"end":14},"code":"SC2005","message":"'glarb' is not a recognized built-in function name."},{"severity":"Error","location":{"start":32,"end":37},"code":"SC2005","message":"'blarg' is not a recognized built-in function name."}]}
ActivityId: db54d090-b952-4c1a-b2a7-40e0cf645b5c, Microsoft.Azure.Documents.Common/2.14.0
```

This string is provided, as a fully escaped single JSON string, in the response body when executing the query. So, the exact error JSON format isn't part of the object model of the response. Therefore, we're doing some hacky parsing here, by extracting the text after `Message: ` through to the end of that line and expecting it to be a JSON object. It's fairly well protected to ensure it doesn't completely fail. It may fail to decode the error properly, but it should fall back to treating the message as a single error if that's the case.

Rarely, I've seen accounts start to return a `BadRequest` error with no details. When this happens, we can't parse individual errors out and fall back to rendering the error as received (see below). ~~I haven't yet tracked down the reason for that but I'm investigating and~~ ~~I'm investigating this with the query team~~ The query team has identified the issue and is working on a fix, which shouldn't require any additional work on our side. It happens on `master` as well, so it's not directly related to this PR.

image


analogrelay:
Compilation errors are fixed by upgrading typescript, similar to #1884, so this is blocked on merging that.

jawelton74:
The red squigglies are incorrect when using 'Execute Selection'  

image

analogrelay:
> The red squigglies are incorrect when using 'Execute Selection'

Ah! Good catch. Should be straightforward to fix. I'll revert to draft while I work on that and undraft (and ping) when done.

jawelton74 (approved):

asierisayas (approved):

PR #1895, opened 2024-07-03
Fix issue with fetching data for Tables account

sindhuba:


sunghyunkang1111 (approved):

src/Common/CosmosClient.ts
```diff
@@ -81,6 +81,12 @@ export const tokenProvider = async (requestInfo: Cosmos.RequestInfo) => {
     }
   }
 
+  let retryAttempts: number = 50;
```
sunghyunkang1111, line 84: We already do have a library "promiseRetry" that facilitates retry operations with retry options like number of retries, exponential backoff and etc. like here, maybe we should use it here as well

https://github.com/Azure/cosmos-explorer/blob/dfe79b20f5f71247143bed46f4ec2c2731726578/src/Phoenix/PhoenixClient.ts#L49

sunghyunkang1111 (changes requested):

vchske (approved):

PR #1903, opened 2024-07-08
Fix vCoreMongo/PostGres quickstart bug

sindhuba:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1903?feature.someFeatureFlagYouMightNeed=true)


asierisayas (dismissed):

src/hooks/useKnockoutExplorer.ts
```diff
@@ -519,7 +521,9 @@ async function configurePortal(): Promise<Explorer> {
               useDataPlaneRbac.setState({ dataPlaneRbacEnabled: dataPlaneRbacEnabled });
             }
           } else {
-            await fetchAndUpdateKeys(subscriptionId, resourceGroup, account.name);
+            if (userContext.apiType !== "Postgres" && userContext.apiType !== "VCoreMongo") {
```
vchske, line 524: I'm wondering if there's a way we can clean up this whole block. One thing I think can cause many unwanted bugs (or at least make debugging more difficult) is too many nested ifs/elses. At the very minimum this else should be changed to
} else if (api != postgres && api != vcoremongo) {
    await fetchAndBlah;
}
But take some time today or tomorrow morning and see if there's not a more concise way to do things from line 497 to 526.

sindhuba: Got it. Fixed it. Thanks!

asierisayas (approved):

vchske (approved):

PR #1919, opened 2024-07-22
Save and restore DocumentsTab state to local storage

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1919)

* Save the state of columns width, filter and middle separator per container.
* Add a button to clear filter next to the filter input
* Add a button to clear all the states in the main settings pane:
image
(For Fabric, the Settings pane is enabled, but the RBAC settings are hidden)

### Implementation
The state of each component is saved in local storage inside a single object under the "AppState" key.
The AppState object is a hashmap where each key is a unique combination of `componentName/subComponentName/accountName/databaseName/containerName`.
All path segments are optional except `componentName`. This relies on "/" not being part of the name. Normally, if the parent isn't defined, children should not be defined (for example "containerName" cannot be defined if "databaseName" isn't defined), but as long as the entire path string is unique, the code allows for it.

(The other option would have been to save it in a nested object, which would have removed some duplication, but turned out to complicate the saving and retrieval. A flatter organization seemed simpler and more compact.)

Example:
image

If number of entries reaching a max (now set to 100k) when saving state, we remove the oldest one by timestamp in order to insert the new entry.


src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -829,6 +832,26 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
...
           </div>
         )}
+        <div className="settingsSection">
+          <div className="settingsSectionPart">
+            <DefaultButton
+              onClick={() => {
+                useDialog
+                  .getState()
+                  .showOkCancelModalDialog(
+                    "Clear your settings and history",
```
sevoku, line 842:
```suggestion
                    "Clear local settings and history",
                    "This action will remove all custom configurations saved in your browser, such as filter history, personalized tab arrangements, and column settings in tables.",
                    "Clear",
                    () => deleteAllStates(),
                    "Cancel",
                    undefined,
                  );
              }}
            >
              Clear local settings and history
```
image

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -829,6 +832,26 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
...
+                    "Are you sure?",
+                    "Clear",
+                    () => deleteAllStates(),
+                    "Cancel",
+                    undefined,
+                  );
+              }}
+            >
+              Clear your settings and history
```
sevoku, line 851:
```suggestion
                    "Clear local settings and history",
                    "This action will remove all custom configurations saved in your browser, such as filter history, personalized tab arrangements, and column settings in tables.",
                    "Clear",
                    () => deleteAllStates(),
                    "Cancel",
                    undefined,
                  );
              }}
            >
              Clear local settings and history
```
image

asierisayas (dismissed):

sevoku (dismissed):

asierisayas (dismissed):

src/Explorer/Tabs/DocumentsTabV2/DocumentsTabStateUtil.ts
```diff
@@ -0,0 +1,96 @@
+// Definitions of State data
+
+import { deleteState, loadState, saveState, saveStateDebounced } from "Shared/AppStatePersistenceUtility";
+import { userContext } from "UserContext";
+import * as ViewModels from "../../../Contracts/ViewModels";
+import { Action } from "../../../Shared/Telemetry/TelemetryConstants";
+import * as TelemetryProcessor from "../../../Shared/Telemetry/TelemetryProcessor";
+
+const componentName = "DocumentsTab";
+export type SubComponentName = "ColumnSizes" | "FilterHistory" | "MainTabDivider";
```
analogrelay, line 10: Maybe use string enums for these?

src/Shared/AppStatePersistenceUtility.ts
```diff
@@ -0,0 +1,103 @@
...
+];
+
+/**
+ * /componentName/subComponentName/globalAccountName/databaseName/containerName/
+ * Any of the path segments can be "" except componentName
+ * Export for testing purposes
+ * @param path
+ */
+export const createKeyFromPath = (path: StorePath): string => {
+  let key = `/${path.componentName}`; // ComponentName is always there
```
analogrelay, line 90: If we're not going to escape `/` characters (which I think is fine for now), we should assert the requirement that path segments cannot have `/` in them.

```suggestion
  if (path.componentName.includes("/")) {
      throw new Error(`Invalid component name: ${path.componentName}`);
  }
  let key = `/${path.componentName}`; // ComponentName is always there
```

src/Shared/AppStatePersistenceUtility.ts
```diff
@@ -0,0 +1,103 @@
...
+  orderedPathSegments.forEach((segment) => {
+    const segmentValue = path[segment as keyof StorePath];
```
analogrelay, line 92:
```suggestion
    const segmentValue = path[segment as keyof StorePath];
    if (segmentValue.includes("/")) {
        throw new Error(`Invalid setting path segment: ${segment}`);
    }
```

src/Explorer/Tabs/DocumentsTabV2/DocumentsTabStateUtil.ts
languy: I like union type for its simplicity: it's a simple check at compile time, but enum could be useful to enumerate all values.

sevoku (approved):

analogrelay (approved):

PR #1922, opened 2024-07-31
Add Word Wrap menu option to Editor React monaco context menu in Documents Tab

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1922)

The monaco editor has a context menu invoke with mouse right-click.
* Add a "Toggle" word wrap option in that menu. The change isn't persisted, yet.

image


jawelton74 (dismissed):

sevoku (approved):
Can we change the position of the item? First on top seems to be a bit intrusive, especially for those who use the menu often (mechanical memory) and this is a kind of option you usually use once in a while (if at all once we persist its value). In VSCode that option is at the very bottom of the view menu.

Pilchie:
Thanks @languy!

jawelton74 (approved):

sevoku (approved):
Nice!
image

PR #1926, opened 2024-08-09
Fix resource tree styling

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1926)

In some cases, the "Create Container" button is not present, so the 2nd element (resource tree) incorrectly takes the size of the button.

image

The `makestyles()` method seems to create styles statically (I couldn't pass an argument to conditionally create a style for each option).
So this fix overrides the css `gridTemplateRows` property at runtime based on whether the "Create Container" button is displayed.

### DIFF
--- a/src/Explorer/Sidebar.tsx
+++ b/src/Explorer/Sidebar.tsx
@@ -249,9 +249,11 @@ export const SidebarContainer: React.FC<SidebarProps> = ({ explorer }) => {
     setLoading(false);
   }, [setLoading]);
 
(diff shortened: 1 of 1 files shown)


asierisayas (approved):

jawelton74 (approved):

analogrelay (approved):
Great catch. I did test the tree on Fabric but made some late changes to the global commands layout and didn't review it there. Thanks so much for jumping in!

PR #1927, opened 2024-08-12
[accessibility-1217621]: Keyboard focus gets lost on the page which opens after activating "Data Explorer" menu item present under 'Overview' page.

satya07sb:
This PR addresses an issue where keyboard focus is lost when navigating to the Data Explorer page from the Overview page in Azure Cosmos DB. The problem was that after activating the "Data Explorer" menu item, the page did not correctly set focus on the expected element, impairing keyboard navigation.

**Note**: We are Using `setTimeout `with `0ms `ensures that the code runs asynchronously after the current execution context completes. Without it, the code is not working as expected due to JavaScript's synchronous execution model.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1927?feature.someFeatureFlagYouMightNeed=true)

https://github.com/user-attachments/assets/a1528b42-fb2b-4349-9065-a8755904ac47

### DIFF
--- a/src/Explorer/Sidebar.tsx
+++ b/src/Explorer/Sidebar.tsx
@@ -26,7 +26,7 @@ import { getCollectionName, getDatabaseName } from "Utils/APITypeUtils";
 import { Allotment, AllotmentHandle } from "allotment";
 import { useSidePanel } from "hooks/useSidePanel";
 import { debounce } from "lodash";
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from "react";
 
 const useSidebarStyles = makeStyles({
   sidebar: {
@@ -109,6 +109,7 @@ interface GlobalCommand {
   icon: JSX.Element;
   onClick: () => void;
   keyboardAction?: KeyboardAction;
+  ref?: React.RefObject<HTMLButtonElement>;
 }
 
 const GlobalCommands: React.FC<GlobalCommandsProps> = ({ explorer }) => {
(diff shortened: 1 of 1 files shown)


sevoku:
Is there a specific reason we want to focus the refresh button initially? The first focusable element would rather be the first button in the toolbar. However from usability perspective it'd make more sense to focus the big "New Container" button, users who don't have created any container yet, can just proceed from there and others need only to hit Tab once to reach their data. Thoughts?

satya07sb:
> Is there a specific reason we want to focus the refresh button initially? The first focusable element would rather be the first button in the toolbar. However from usability perspective it'd make more sense to focus the big "New Container" button, users who don't have created any container yet, can just proceed from there and others need only to hit Tab once to reach their data. Thoughts?


As per your suggestion it will make more sense; However, the ticket specifies that the focus should be on the refresh button when opening the Data Explorer.
Ticket: [](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/1217621 )

jawelton74:
> > Is there a specific reason we want to focus the refresh button initially? The first focusable element would rather be the first button in the toolbar. However from usability perspective it'd make more sense to focus the big "New Container" button, users who don't have created any container yet, can just proceed from there and others need only to hit Tab once to reach their data. Thoughts?
> 
> As per your suggestion it will make more sense; However, the ticket specifies that the focus should be on the refresh button when opening the Data Explorer. Ticket: [accessibility-1217621](https://msdata.visualstudio.com/CosmosDB/_workitems/edit/1217621)

I agree with Sevo. Let's seek clarification from the accessibility team.

vchske (approved):

sevoku:
There is a new comment that they closed the bug which is unfortunate.

satya07sb:
> There is a new comment that they closed the bug which is unfortunate. Let's just go ahead and focus the "New Container" button as we discussed?

I’ve discussed this with the accessibility team, but I haven’t received any feedback from them yet. Let’s go ahead with your suggestion for the "New Container" button, as it looks good from every angle.

sevoku (approved):

jawelton74 (approved):

PR #1928, opened 2024-08-13
Removing check for value when populating partition key values

vchske:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1928)

There was a check where, when updating a document, it would only populate a partition key value if there was a non-empty value for that field in the document to begin with. This is incorrect behavior as a document in a container with hierarchical partitions can have an empty value for one of those partition key fields. The result of this would be the document would be sent with x number of partition key fields but only x - 1 partition key values which would fail backend validation.

This PR simply allows whatever was in that field to be passed on to the "partitionKeyValues" property. If that value is "" then so be it. Updating documents with empty values for partition key fields should now work as expected.

### DIFF
--- a/src/Utils/QueryUtils.test.ts
+++ b/src/Utils/QueryUtils.test.ts
@@ -108,6 +108,7 @@ describe("Query Utils", () => {
       },
       Elevation: 3742,
       Type: "Stratovolcano",
+      Category: "",
       Status: "Tephrochronology",
       "Last Known Eruption": "Last known eruption from A.D. 1-1499, inclusive",
       id: "9e3c494e-8367-3f50-1f56-8c6fcb961363",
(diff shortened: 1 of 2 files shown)


sunghyunkang1111 (dismissed):

src/Utils/QueryUtils.test.ts
sunghyunkang1111, line 1: Why not change the test case to check for empty string?

vchske: The test case that was removed was testing for a partition key field that is not present in the partition key definition. But I don't think this is a valid use case as we use the definition returned by the backend.
I suppose I could create a test that has an empty value for one of the fields?

sunghyunkang1111 (approved):

jawelton74 (approved):

PR #1940, opened 2024-08-20
Improve Entra ID token acquisition logic

sevoku:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1940)

This PR improves the Entra ID / AAD token acquisition logic a little bit and is also groundwork to make Entra ID work in Fabric. 

Improvements:
* Avoid Sign-In Popups as much as possible with an optional `silent` flag
  * With a `loginHint` MSAL is able to sign in with the RBAC scope using `msal.ssoSilent` [without showing any account if the user is already signed in to the host environment (Portal/Fabric)](https://learn.microsoft.com/en-us/entra/identity-platform/msal-js-sso#sso-between-different-apps)
  * Only fall back to [`msal.loginPopup` with `prompt: none`](https://learn.microsoft.com/en-us/entra/identity-platform/msal-js-prompt-behavior#interactive-requests-with-promptnone) if necessary
* Try to silently acquire the RBAC token on startup. If successful, don't show the Entra ID button and users should be able to load data right away.

Further Improvements / Thoughts:
* [x]  Previous logic was always calling `msal.loginPopup` first, so there was always a Popup shown and in most cases auto-closed since MSAL was signed-in already. Was this behaviour intentional? Are there some corner cases where customers might need to sign in with a different account for the RBAC scope?
**EDIT: After thinking more about it, it doesn't make much sense to sign in with a different account. Hence checking first if MSAL already has an account simply avoids unnecessary Sign-Ins.**
* [x] "cachedTenantId" from Local Store is only used with the Hosted environment, are there any reasons we want to set/read these values when running elsewhere? The previous logic was storing the Tenant ID in local store but was not reusing it initially. We could though keep storing it and try filter known accounts by Tenant ID i.e. for debugging purposes, I'm just not sure how useful this is.
**EDIT: I don't think we need this at all, will remove, also see next point:**
* [ ] Portal should send a user name when initializing DE. With this PR `acquireMsalTokenForAccount` will pick it up and do a silent sign-in for RBAC. If the user is already signed in and eventually with different tenants, the user name will help selecting the matching account and acquiring the RBAC token silently.
* [x] Update Documentation: https://github.com/MicrosoftDocs/azure-databases-docs/pull/25


jawelton74:
> > image
> 
> @jawelton74 Maybe rephrase it like this: image
> 
> Or we can simply remove the notice, since in most cases it should not be an issue (and there is documentation otherwise)?
> 
> Oh and we should try to sign in silently when people change that setting and hit Apply. I'll add that.

I think we can try removing it now with this change. We do emit an error in the notifications pane if user tries to run an operation and we don't have the token, prompting them to click on the button.

And yes, we'll need to update this documentation (which currently says that user must click on the Login button) :

https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-setup-rbac#use-data-explorer

sevoku:
@jawelton74 I have tested this whole scenario with a clean browser session and here is a quick update on the requirement to Re-SignIn in DE:

In case a user has never signed in to DE through MSAL, MSAL will not be able to sign in silently, however since the user is signed in to Portal/Fabric, MSAL doesn't ask me for a password but 2FA only.

This means for all clean browser sessions the Login button will be shown and therefore I'd suggest to keep the notice, otherwise it'll be confusing for new users.

sevoku:
@sindhuba @jawelton74 this is ready now. With the last commit I don't get any popups at all with Fabric (Fabric side is WIP still, will open a separate PR for that when it's ready). Please have a look at the last task in the description, we'd need Portal to send a userName to DE for a silent sign-in E2E. This PR should not be blocked on that though.

languy (dismissed):

sevoku:
Here is the documentation update: https://github.com/MicrosoftDocs/azure-databases-docs/pull/25

sindhuba (approved):

jawelton74 (approved):

languy (approved):

PR #1945, opened 2024-08-23
Improve SettingsPane

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1945)

The current SettingsPane has too many options.
Move all items into an `<Accordion>`. Only one option can be opened at a time.
Move the Info tips (i) text into the accordion section.

image


sevoku (approved):

analogrelay:
FYI: The Playwright failures are the resource token issue, which appears to be unrelated to this PR (and are being investigated separately).

PR #1948, opened 2024-08-26
Migrate Copilot local persistence (toggle and prompt history) to new local storage infrastructure

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1948)
Move the copilot toggle status and prompt history into the `AppState` local storage, in order to manage it like the other user preference data. With this change, the "Clear History" button in the general settings pane also clear the copilot data.


sunghyunkang1111 (approved):

src/Explorer/Tabs/QueryTab/QueryTabComponent.test.tsx
```diff
@@ -16,6 +18,24 @@ import React from "react";
...
+    return undefined;
+  }
+};
+
+jest.mock("Shared/AppStatePersistenceUtility", () => ({
+  loadState,
+  AppStateComponentNames: {
+    QueryCopilot: "QueryCopilot",
+  },
+}));
```
analogrelay, line 37: I don't think you need to do it now, but I'm noticing a pattern for testing state loading/saving that might be worth refactoring into a helper 🤔

analogrelay (dismissed):
I verified that the playwright failures are the unrelated resource token failure, so I'm OK with merging even with them failing.

src/Explorer/Tabs/QueryTab/QueryTabComponent.test.tsx
languy: Turns out we don't need to mock `loadState`, we can use the real api to persist the state with the right value before the test (and cleanup afterwards). So I replaced this code with the actual call to set the state in the tests.

sunghyunkang1111 (approved):

PR #1951, opened 2024-08-27
Reduce extra spacing in the new tree and items tab

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1951)

After some user feedback that there's too much blank space in the new tree view, I went back in and looked for some things we could do to better fit DE into a small screen environment.

This PR is the result of that. Here's a video toggling between the old and new layouts. The screen size here is 1366x768, which is our standard "desktop" (as in not-mobile, laptops _are_ included) minimum size. You can see that there's quite a bit of space savings from these changes:

https://github.com/user-attachments/assets/d4eca67e-145b-48d2-88df-67a54cf55169

In addition, I've added some icons back to the tree, after discussing with Doug and Meredith. We don't have icons everywhere, but we've added icons to the Database, Collection, and Database Scale nodes. We intentional avoided putting icons on the child nodes of a Collection for now. The tree layout works best when all nodes at a given level have an icon and we wanted to focus on distinguishing databases and collections, adding an icon to the "Scale" node primarily so that it aligns with Collection nodes.

The layout changes I've made:

* Reduce the row size used for the tree, items tab, and other layout elements (for example, the tab bar and message bars have this height) from 36px to 32px
* Reduce the base font size from 14px to 13px, primarily so that it aligns with the font size used in the sidebar on the Portal.

I've made these changes in a way that starts to lay the groundwork for allowing users to toggle layout sizes. The designs we're working from include this, it just hasn't been implemented yet. There are plans for three sizes: Compact, Cozy, and Roomy. I've designated what we're doing here the "Compact" design. Later on, we can add the Cozy and Roomy designs.

| Platform | Screenshot (1366x768) |
| - | - |
| **Hosted** | image |
| **Portal** | image |
| **Fabric** | image |

### DIFF
--- a/src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
+++ b/src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
@@ -149,15 +149,19 @@ export const TreeNodeComponent: React.FC<TreeNodeComponentProps> = ({
 
   // We use the expandIcon slot to hold the node icon too.
   // We only show a node icon for leaf nodes, even if a branch node has an iconSrc.
+  const treeIcon = node.iconSrc === undefined ?
+    undefined
+  : typeof node.iconSrc === "string" ? (
+    <img src={node.iconSrc} className={treeStyles.nodeIcon} alt="" />
+  ) : (
+    node.iconSrc
+  );
+
   const expandIcon = isLoading ? (
     <Spinner size="extra-tiny" />
-  ) : !isBranch ? (
-    typeof node.iconSrc === "string" ? (
-      <img src={node.iconSrc} className={treeStyles.nodeIcon} alt="" />
-    ) : (
-      node.iconSrc
-    )
-  ) : openItems.includes(treeNodeId) ? (
+  ) : !isBranch ?
+      undefined
+  : openItems.includes(treeNodeId) ? (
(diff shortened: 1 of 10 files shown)


src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
```diff
@@ -174,7 +175,7 @@ export const TreeNodeComponent: React.FC<TreeNodeComponentProps> = ({
       <TreeItemLayout
         className={mergeClasses(
           treeStyles.treeItemLayout,
-          expandIcon ? undefined : treeStyles.treeItemLayoutNoIcon,
+          //expandIcon ? undefined : treeStyles.treeItemLayoutNoIcon,
```
vchske, line 178: Should we keep this?

vchske (dismissed):

src/Explorer/Controls/TreeComponent/TreeNodeComponent.tsx
analogrelay: Oop, no we shouldn't, I'll remove it.

asierisayas (approved):

sevoku (approved):

analogrelay:
Confirmed the Playwright failures are only on the resource token related tests, which are broken for an unrelated reason.

PR #1953, opened 2024-08-29
fix rendering of global commands menu

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1953?feature.someFeatureFlagYouMightNeed=true)

I was opening the split button "Global Commands" menu and I noticed something:

image

Hey menu, what are you doing way up there?

Well, it turns out that when you put two DOM elements under a `<MenuTrigger>`, it gets a little confused and does all the positioning based on the _last_ element. In our case, the last element is only visible on small screens (we swap between the first and last element based on container width using CSS container queries). As a result, that messes with things when the first element (the SplitButton) is the one that's visible.

This PR wraps those elements in a new `div`. Unfortunately, that's not entirely enough, we also need the Menu to know to anchor the menu off that div. We do that using a React state value.

With this change, the menu is back where it belongs:

image

And still works fine in small screens:

image

### DIFF
--- a/src/Explorer/Sidebar.tsx
+++ b/src/Explorer/Sidebar.tsx
@@ -10,7 +10,7 @@ import {
   SplitButton,
   makeStyles,
   mergeClasses,
-  shorthands
+  shorthands,
 } from "@fluentui/react-components";
 import { Add16Regular, ArrowSync12Regular, ChevronLeft12Regular, ChevronRight12Regular } from "@fluentui/react-icons";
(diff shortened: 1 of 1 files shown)


vchske (dismissed):

asierisayas (dismissed):

analogrelay:
Playwright test failures are all in the resource token tests.

sevoku (approved):

jawelton74 (approved):

PR #1954, opened 2024-08-30
Better handling throttling error in bulk delete

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1954)

A very common problem is bulk delete operations leading to a throttling error: the default minimum throughput of a container typically doesn't handle bulk deleting hundreds of documents at a time. Only a few dozens will be deleted, while the rest returns a 429 (and a delay to retry).

#### NoSQL
For NoSql, each individual document bulk delete operation returns its own status code, allowing us to retry the failed throttled operations. During the deleting process a modal dialog is presented with a progress and status. If throttling errors arise, they are automatically retried and a warning is displayed:
deleting
If other failures arise, the dialog will display the number of errors.

The user can choose to abort at any time. Any failure (including un-retried 429's) will be displayed in the dialog:
aborted

Otherwise, once all documents have been deleted, the dialog displays the result (including failure count if present).
successful

Clicking "Close" dismisses the dialog with no other effect.

#### Mongo
Our MongoProxy uses the mongo client which does not return individual bulk operation result. However, the remote call for bulk delete does return a 400 (Bad Request) status code with a message body from Cosmos DB: "A write operation resulted in an error. Error=16500, RetryAfterMs=4, Details='Batch write error.".
This allows us to present a different error message with more explanation and suggestion to mitigate.
failed_mongo

In both API's there is a [Learn More](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-request-rate-too-large) link to point the user to our doc about throttling errors.


sevoku:
> Any failure (including un-retried 429's) will be displayed in the dialog:
> deleted_aborted

Do we log the actual error in the notification area? It might be useful for the user to know what the actual error was.

> Otherwise, once all documents have been deleted, the dialog displays the result (including failure count if present).
> deleted completed

If the operation was successful with retries, I'd rephrase the warning. Right now it sounds like something failed, although it didn't but was slower than it could be. Maybe something like "**Warning** Some delete requests were throttled (429). For faster bulk-deletes in the future, consider increasing the throughput on your container or database."

analogrelay (approved):

sevoku (approved):

src/Explorer/Tabs/DocumentsTabV2/DocumentsTabV2.tsx
```diff
@@ -1915,6 +1917,26 @@ export const DocumentsTabComponent: React.FunctionComponent<IDocumentsTabCompone
...
+   * If completed and all documents were deleted, display: all documents deleted.
+   * @returns 429 warning message
+   */
+  const get429WarningMessageNoSql = (): string => {
+    let message = 'Some delete requests failed due to a "Request too large" exception (429)';
+
+    if (bulkDeleteOperation.count === bulkDeleteProcess.successfulIds.length) {
+      message += ", but were successfully retried.";
+    } else if (bulkDeleteMode === "inProgress" || bulkDeleteMode === "aborting") {
+      message += "(429). Retrying now.";
```
analogrelay, line 1931: I don't think you need the `(429)` prefix here, it's already in the base message.

```suggestion
      message += ". Retrying now.";
```

sevoku (approved):

asierisayas (approved):

PR #1961, opened 2024-09-05
Allow slashes in persistence keys

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1961)

Users might lose their settings if one of the segments contains uri character that gets encoded (e.g. a `/` in the name, since we use `/` as a separator).

To keep the code simple and free from legacy logic, I did not implement any migration from the old persistence.
Previous settings will disappear for some users who have `/` in the name (it should be the subset who use resource token). Persistence has been introduced only recently and for a few preferences, so the amount of lost user preferences should be very limited.


analogrelay (commented):
Looks like a great approach to me. URL encode/decode is built into the browser and specifically designed to allow `/` in path segments (among other things). I like this much better than trying to design our own encoding scheme.

jawelton74 (approved):

vchske (approved):

PR #1975, opened 2024-09-19
Add config files for Codespaces

analogrelay:

*This change doesn't affect the app functionality at all*

This PR adds a few config files to provide a standardized configuration for launching a GitHub Codespace or VS Code Dev Container (locally) for this repo. It doesn't require the use of codespaces at all, but makes it possible for a user to quickly open the repo up in a Codespace that is preconfigured to immediately run the app.

I've been using Codespaces more for development and trying it out on this repo (among others I'm working with, in particular the Azure SDKs) and it provides a useful alternative to having to install all the prereqs on my local box.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,32 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "Azure Cosmos DB Explorer",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"onCreateCommand": ".devcontainer/oncreate",
+	"features": {
+		"ghcr.io/devcontainers/features/azure-cli:1": {
+			"version": "latest"
+		},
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+	"name": "Azure Cosmos DB Explorer",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"onCreateCommand": ".devcontainer/oncreate",
```
Pilchie, line 9: Weird that one of these is relative to current dir, and the other is to repo root.  I assume this works though?

Pilchie (approved):

.devcontainer/devcontainer.json
analogrelay: It does work, yeah. There's a reason for it, but it's definitely a little obscure. `build.dockerfile` specifies the value for the `-f` argument to `docker build`. So, this is effectively saying: `docker build -f Dockerfile`. In order to avoid the image pulling in the contents of the entire repo, that command is run **from the `.devcontainer`** directory.

jawelton74 (approved):

PR #1978, opened 2024-09-24
[Keyboard Navigation - Azure Cosmos DB - Data Explorer]: Keyboard focus is not retaining back to 'more' button after closing 'Delete container' dialog.

satya07sb:
This PR addresses an issue in the Azure Cosmos DB Data Explorer where keyboard focus does not return to the 'More' button after closing the 'Delete Container' dialog. This affects keyboard navigation and overall accessibility for users relying on keyboard inputs.

And This PR implements a holistic method for focusing on custom elements. The `useSidePanel` hook will now accept an optional `getRef` property as required. Whenever we call `useSidePanel`, we can pass an additional ref property, allowing it to focus on that specific element. We are using a `setTimeout` of 300 ms to accommodate delays in component rendering.

[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1978?feature.someFeatureFlagYouMightNeed=true)

https://github.com/user-attachments/assets/aa7803bd-2960-4500-ad99-411491d2455d


src/hooks/useSidePanel.ts
```diff
@@ -7,12 +7,18 @@ export interface SidePanelState {
...
   isOpen: false,
   panelWidth: "440px",
   openSidePanel: (headerText, panelContent, panelWidth = "440px") =>
     set((state) => ({ ...state, headerText, panelContent, panelWidth, isOpen: true })),
-  closeSidePanel: () => set((state) => ({ ...state, isOpen: false })),
+  closeSidePanel: () => {
+    set((state) => ({ ...state, isOpen: false }));
+    const lastFocusedElement = useSidePanel.getState().getRef;
+    setTimeout(() => {
+      lastFocusedElement?.current?.focus();
```
jawelton74, line 21: Do we need to clear the getRef state after setting the focus there ?

vchske (approved):

jawelton74 (approved):

PR #1979, opened 2024-09-24
Ensure the "Ctrl+Alt+["/"Ctrl+Alt+]" shortcuts don't get triggered on "AltGr+8"/"AltGr+9"

analogrelay:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1979)

We currently support using `Ctrl+Alt+[` and `Ctrl+Alt+]` for swapping to the left/right tab in the query editor. However, the sequence `AltGr+8`/`AltGr+9` on many European keyboards, used to type the characters `[` and `]`, gets reported in many browsers as `Ctrl+Alt+[`/`Ctrl+Alt+]`. Because we have our keyboard bindings, this means you can't type `[` or `]` in the query editor on these keyboards and have to swap to another text editor to type these and copy-paste them in. We have a fair amount of customer feedback about this, so this PR just removes those bindings (**Update below**, I was able to fix the bindings so they still work on US keyboards). We still have `Ctrl+F6` and `Ctrl+Shift+F6` to swap between tabs.

**I could use some folks with European keyboards to validate this. I'm looking to see if there are options for me to emulate one myself, but I don't have a physical one to validate that the previous experience is broken and is now fixed by this PR**


analogrelay:
I was able to validate this by installing a Norwegian keyboard on my local machine and using the On-Screen Keyboard. Prior to this change, it was not possible to type `[` or `]` in to the query editor, as it would switch tabs. After this change, this sequence now correctly types a `[` or `]`.

I also used a [Keyboard Event Viewer](https://w3c.github.io/uievents/tools/key-event-viewer.html) page to see if I could find a way to make `Ctrl+Alt+[` continue to work on US keyboards, and I did find a way! When on a Norwegian keyboard, pressing `AltGr+8` yields the highlighted `keyup` event:

image

Note that while `event.key` is `[`, the `event.code` value is `Digit8`. The `code` here is a *more accurate* representation of the exact keys being pressed. When combined with `event.getModifierState()`, you get `AltGraph + Digit8`. Pressing `Ctrl+Alt+[` on a Norwegian keyboard results in the _same_ event.

On a US keyboard, typing `Ctrl+Alt+[` yields the highlighted `keyup` event below:

image

Note again that while `event.key` is still `[`, the `event.code` is now `BracketLeft`. This gives us something to distinguish the two. The tinykeys library we're using [can match against *either `key` or `code`*](https://www.npmjs.com/package/tinykeys#keybinding-syntax). We were previously writing the shortcut as `$mod+Alt+[`, which mean tinykeys was checking for `event.key` to be `[`. If, instead, we use `$mod+Alt+BracketLeft`, we cause tinykeys to compare against `event.code` **instead** of `event.key`. This allows us to distinguish between the two cases and keep the keyboard shortcut functioning for US keyboards while treating `AltGr+8` properly as typing a `[` and thus not triggering the shortcut.

I validated this behavior locally. When on a US keyboard, I can use either `Ctrl+Alt+[` **or `Ctrl+Shift+F6` to move to the left tab. When on a Norwegian keyboard, I can use `AltGr+8` to type a `[` **and** use `Ctrl+Shift+F6` to move to the left tab.

analogrelay:
NOTE: I would still appreciate it if anyone can provide validation on a physical keyboard. I mostly trust the Windows on-screen keyboard to properly emulate the keyboard, but better safe than sorry here.

vchske (approved):

jawelton74 (approved):

PR #1998, opened 2024-10-28
Improve DocumentsTab filter input

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/1998)
 * Always keep filter editable
 * Single "Apply filter"/"Cancel"
 * Fluent UI-based dropdown for saved and default filters with link to learn more. 
 * Clear button on the right clear filter
 * Cleaned up old style and code (for example don't focus with outline)
 * Input and dropdown navigation with tab for a11y compliance
 * The input and dropdown menu is a React component.

image

#### Behavior of the dropdown and keyboard navigation
* Dropdown will show when input gets focus
* Dropdown shows only when input has no text
* Clear button at the end of the input box becomes down button to reveal dropdown if input has no text. 
* Generally first dropdown item gets autofocus when dropdown opens (when input doesn't have focus)
* Arrow up/down move focus in the dropdown item. If last item, we focus back to the input.
* Tab or shift tab also work. If last item, tab goes to the Apply button.

### DIFF
--- a/src/Explorer/Controls/InputDataList/InputDataList.tsx
+++ b/src/Explorer/Controls/InputDataList/InputDataList.tsx
@@ -19,8 +19,6 @@ import { NormalizedEventKey } from "Common/Constants";
 import { tokens } from "Explorer/Theme/ThemeUtil";
 import React, { FC, useEffect, useRef } from "react";
 
-const theme = getTheme();
-
 const useStyles = makeStyles({
   container: {
     padding: 0,
@@ -40,7 +38,6 @@ const useStyles = makeStyles({
     width: "100%",
     fontSize: tokens.fontSizeBase300,
     fontWeight: 600,
-    color: theme.palette.themePrimary,
     padding: `${tokens.spacingVerticalM} 0 0 ${tokens.spacingVerticalM}`,
   },
   dropdownStack: {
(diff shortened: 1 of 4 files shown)


sunghyunkang1111 (approved):

sevoku:
This looks great, however I see 2 issues:
1. a11y: there is no way to enter the dropdown with suggestions/history using the keyboard, when entering the text entry the dropdown is shown, however I can't select anything using the arrow keys or tab into it (without the change I can use up/down arrow to select something)
2. when the entry is empty, the "X" button remains. I think it should change to a "Down Arrow" to reveal the history (and also might be the alternative entry point for keyboard accessibility):
image

languy:
> This looks great, however I see 2 issues:
> 
> 1. a11y: there is no way to enter the dropdown with suggestions/history using the keyboard, when entering the text entry the dropdown is shown, however I can't select anything using the arrow keys or tab into it (without the change I can use up/down arrow to select something)
> 2. when the entry is empty, the "X" button remains. I think it should change to a "Down Arrow" to reveal the history (and also might be the alternative entry point for keyboard accessibility):
>    image

Good points. Fixed and added more description about keyboard navigation.

sevoku (approved):
Thanks! Much better!

jawelton74 (approved):

PR #2005, opened 2024-11-08
Update styles for Fabric

languy:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/2005)

The new Fabric extension code already takes care of these paddings.

### DIFF
--- a/less/documentDBFabric.less
+++ b/less/documentDBFabric.less
@@ -26,7 +26,6 @@ a:focus {
 
 #divExplorer {
   background-color: #f5f5f5;
-  padding: @FabricBoxMargin;
 }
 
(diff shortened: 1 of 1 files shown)


less/documentDBFabric.less
```diff
@@ -26,7 +26,6 @@ a:focus {
 
 #divExplorer {
   background-color: #f5f5f5;
-  padding: @FabricBoxMargin;
```
languy, line 29: This is the only functional change. Everything else is reformatting.

jawelton74 (approved):

sevoku (approved):

asierisayas (approved):

PR #2009, opened 2024-11-14
Vector Embedding and Full Text Search

vchske:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/2009)


src/Contracts/DataModels.ts
```diff
@@ -199,11 +200,19 @@ export interface IndexingPolicy {
   compositeIndexes?: any[];
   spatialIndexes?: any[];
   vectorIndexes?: VectorIndex[];
+  fullTextIndexes?: FullTextIndex[];
 }
 
 export interface VectorIndex {
   path: string;
   type: "flat" | "diskANN" | "quantizedFlat";
+  diskANNShardKey?: string;
```
Pilchie, line 209: Shard key applies to both diskANN and quantizedFlat.

Also, we aren't shipping this next week because we don't have it deployed.

vchske: In the specs I have the field is still called "diskANNShardKey". I will leave this until the feature is released and edit if necessary. This field isn't visible anywhere in the ux yet.

sunghyunkang1111 (dismissed):

PR #2027, opened 2025-01-14
Removed NPS Survey from DE since it has been moved to the Overview Blade

JustinKol:
[Preview this branch](https://cosmos-explorer-preview.azurewebsites.net/pull/2027?feature.someFeatureFlagYouMightNeed=true)

### DIFF
--- a/src/hooks/useKnockoutExplorer.ts
+++ b/src/hooks/useKnockoutExplorer.ts
@@ -53,7 +53,7 @@ import {
 } from "../Utils/AuthorizationUtils";
 import { isInvalidParentFrameOrigin, shouldProcessMessage } from "../Utils/MessageValidation";
 import { getReadOnlyKeys, listKeys } from "../Utils/arm/generatedClients/cosmos/databaseAccounts";
-
+import { applyExplorerBindings } from "../applyExplorerBindings";
 // This hook will create a new instance of Explorer.ts and bind it to the DOM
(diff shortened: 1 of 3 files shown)


asier-isayas (approved):

src/hooks/useKnockoutExplorer.ts
```diff
@@ -95,13 +94,6 @@ export function useKnockoutExplorer(platform: Platform): Explorer {
     effect();
   }, [platform]);
 
-  useEffect(() => {
-    if (explorer) {
-      applyExplorerBindings(explorer);
```
jawelton74, line 100: We still need to keep this, don't we ?

src/hooks/useKnockoutExplorer.ts
```diff
@@ -89,19 +88,17 @@ export function useKnockoutExplorer(platform: Platform): Explorer {
           restoreOpenTabs();
         }
 
+        useEffect(() => {
```
jawelton74, line 91: Why move this ? I think it would just remain where it was, with the openNPSSurveyDialog() call removed ?

JustinKol: Yes, sorry. I added it back in the wrong place.

jawelton74 (approved):

asier-isayas (approved):

PR #2037, opened 2025-01-29
Manual Region Selection

bogercraig:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2037?feature.someFeatureFlagYouMightNeed=true)

image

## Overview and Impact
This PR adds the option for NoSQL accounts to select the region Data Explorer (DE) uses for all operations on the cosmos client.  Selecting a read region will disable all CRUD operations.  This only applies when users use AAD/Entra auth.

Impacted Data Explorer sections include Documents tab, query editor, and conflict resolution with multi region write accounts.

Stored procedures, triggers, and UDF should not be impacted even when read only region is selected.  CRUD operations with these are ARM calls.

## How Do Regions Swap?
After auth with Entra/AAD, Data Explorer retrieves all read and write endpoints from the Cosmos DB RP.  If a region besides global is selected, the client will use a regional endpoint to initialize the client and change `enableEndpointDiscovery` to false.  This prevents the client from automatically finding an optimal region based on location while the location is manually set.

If another region is selected, the client is disposed of and recreated with the newly selected endpoint.  If the global endpoint is selected, the client should operate the same as before this change.

## Testing
Set CosmosClient diagnostics level to `debugUnsafe` and in both local and PR preview instances, manually selected region further from the client.  Checked that the client's contacted region matched the endpoint proxied to in the requestPlugin.  Then tracked the activity ID returned by the client to the gateway contacted in the Cosmos DB backend using FrontendEvent table.  From client, to proxy, to backend, request was tracked to the selected region.

This ensured that whether using the `requestPlugin` in a test environment or the client directly in production, it would reach out to the designated region.

Errors are returned rather than automatically selecting a write regions when write requests are made on a read endpoint.  These operations should, however, all be disabled when a read region is selected.

## Code Changes

- CosmosClient.ts
For instances of DE that run locally or in PR preview, the endpoint selection function was changed to proxy requests directly to the regional endpoint.  When running in prod, the cosmos client uses the regional endpoint and then disables endpoint discovery if the user selects a non-global endpoint.
- StorageUtility.ts
Add `SelectedRegionalEndpoint` so user's region selection persists between sessions.
- UserContext.ts
`selectedRegionalEndpoint` and `writeEnabledInSelectedRegion` added for configuring cosmos client based on selected region.  Absence of these in the user context results in the client defaulting to the current behavior.
- useKnockoutExplorer.ts
`checkAndUpdateSelectedRegionalEndpoint` introduced to retrieve region selected from previous sessions and configure `selectedRegionalEndpoint` and `writeEnabledInSelectedRegion`.  Ensures selected endpoint still exists on the account.
- SettingsPane.tsx
Adds section to settings menu for region selection.  Configures client based on settings update and changes `clientWriteEnabled` for CRUD operation buttons.
- DocumentsTabV2.tsx
Add `clientWriteEnabled` zustand state for react elements.  For immediately updating CRUD UI elements state.
- QueryTabComponent.tsx
Disable new query creation and update when read only region selected.

### DIFF
--- a/src/Common/CosmosClient.ts
+++ b/src/Common/CosmosClient.ts
@@ -171,7 +171,7 @@ let _client: Cosmos.CosmosClient;
 
 export function client(): Cosmos.CosmosClient {
   console.log(`Client Call`);
-  console.log(`Client Call: userContext.hasDataPlaneRbacSettingChanged: ${userContext.hasDataPlaneRbacSettingChanged}`);
+  console.log(`Client Call: userContext.refreshCosmosClient: ${userContext.refreshCosmosClient}`);
   if (_client) {
     if (!userContext.refreshCosmosClient) {
       return _client;
--- a/src/Common/CosmosClient.ts
+++ b/src/Common/CosmosClient.ts
@@ -128,8 +128,8 @@ export const endpoint = () => {
     return configContext.EMULATOR_ENDPOINT || location.origin;
   }
   return (
-    userContext.endpoint ||
     userContext.selectedRegionalEndpoint ||
+    userContext.endpoint ||
     userContext?.databaseAccount?.properties?.documentEndpoint
   );
 };
--- a/src/Common/CosmosClient.ts
+++ b/src/Common/CosmosClient.ts
@@ -120,21 +120,6 @@ export const requestPlugin: Cosmos.Plugin<any> = async (requestContext, diagnost
   console.log(`REQUEST CONTEXT PROXY: ${JSON.stringify(requestContext.headers["x-ms-proxy-target"])}`);
   console.log(`WRITE ENABLED ON ENDPOINT: ${userContext.writeEnabledInSelectedRegion}`);
   return await next(requestContext);
-
-  // try {
-  //   return await next(requestContext);
-  // } catch (error) {
-  //   throw {
-  //     code: error?.code || undefined,
-  //     message: error.message,
-  //   };
-  // }
-
-  // try {
-  //   return await next(requestContext);
-  // } catch (error) {
-  //   console.log(error.code);
-  // }
 };
 
 export const endpoint = () => {
@@ -265,25 +250,6 @@ export function client(): Cosmos.CosmosClient {
       },
     },
   };
-
-  // // Account details from userContext.
-  // // console.log(`userContext details: ${JSON.stringify(userContext)}`);
-  // console.log(`userContext.databaseaccount details: ${JSON.stringify(userContext.databaseAccount)}`);
-  // console.log(
-  //   `userContext?.databaseAccount?.properties?.documentEndpoint details: ${JSON.stringify(
-  //     userContext?.databaseAccount?.properties?.documentEndpoint,
-  //   )}`,
-  // );
-  // console.log(
-  //   `userContext?.databaseAccount?.properties?.readLocations details: ${JSON.stringify(
-  //     userContext?.databaseAccount?.properties?.readLocations,
-  //   )}`,
-  // );
-  // console.log(
-  //   `userContext?.databaseAccount?.properties?.writeLocations details: ${JSON.stringify(
-  //     userContext?.databaseAccount?.properties?.writeLocations,
-  //   )}`,
-  // );
 
   if (configContext.PROXY_PATH !== undefined) {
     (options as any).plugins = [{ on: "request", plugin: requestPlugin }];
--- a/src/Common/CosmosClient.ts
+++ b/src/Common/CosmosClient.ts
@@ -118,13 +118,21 @@ export const requestPlugin: Cosmos.Plugin<any> = async (requestContext, diagnost
   console.log(`REQUEST CONTEXT ENDPOINT: ${JSON.stringify(requestContext.endpoint)}`);
   requestContext.headers["x-ms-proxy-target"] = endpoint();
   console.log(`REQUEST CONTEXT PROXY: ${JSON.stringify(requestContext.headers["x-ms-proxy-target"])}`);
+  // return await next(requestContext);
+
+  // try {
+  //   return await next(requestContext);
+  // } catch (error) {
+  //   throw {
+  //     code: error?.code || undefined,
+  //     message: error.message,
+  //   };
+  // }
+
   try {
     return await next(requestContext);
   } catch (error) {
-    throw {
-      code: error?.code || undefined,
-      message: error.message,
-    };
+    console.log(error.code);
   }
 };
 
@@ -244,6 +252,7 @@ export function client(): Cosmos.CosmosClient {
   const options: Cosmos.CosmosClientOptions = {
     endpoint: endpoint() || "https://cosmos.azure.com", // CosmosClient gets upset if we pass a bad URL. This should never actually get called
     key: userContext.dataPlaneRbacEnabled ? "" : userContext.masterKey,
+    diagnosticLevel: Cosmos.CosmosDbDiagnosticLevel.debugUnsafe,
     tokenProvider,
     userAgentSuffix: "Azure Portal",
     defaultHeaders: _defaultHeaders,
--- a/src/Common/CosmosClient.ts
+++ b/src/Common/CosmosClient.ts
@@ -115,10 +115,7 @@ export const tokenProvider = async (requestInfo: Cosmos.RequestInfo) => {
 
 export const requestPlugin: Cosmos.Plugin<any> = async (requestContext, diagnosticNode, next) => {
   requestContext.endpoint = new URL(configContext.PROXY_PATH, window.location.href).href;
-  console.log(`REQUEST CONTEXT ENDPOINT: ${JSON.stringify(requestContext.endpoint)}`);
   requestContext.headers["x-ms-proxy-target"] = endpoint();
-  console.log(`REQUEST CONTEXT PROXY: ${JSON.stringify(requestContext.headers["x-ms-proxy-target"])}`);
-  console.log(`WRITE ENABLED ON ENDPOINT: ${userContext.writeEnabledInSelectedRegion}`);
   return await next(requestContext);
 };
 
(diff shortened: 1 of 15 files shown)


src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -143,6 +146,11 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
       ? LocalStorageUtility.getEntryString(StorageKey.IsGraphAutoVizDisabled)
       : "false",
   );
+  const [selectedRegionalEndpoint, setSelectedRegionalEndpoint] = useState<string>(
+    LocalStorageUtility.hasItem(StorageKey.SelectedRegionalEndpoint)
+      ? LocalStorageUtility.getEntryString(StorageKey.SelectedRegionalEndpoint)
```
languy, line 151: Is this a per cosmos db account preferences or global? If the user switches account in DE, do we expect this to stick?

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -274,6 +320,34 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
...
+      !selectedRegionIsGlobal &&
+      selectedRegionalEndpoint !== storedRegionalEndpoint
+    ) {
+      LocalStorageUtility.setEntryString(StorageKey.SelectedRegionalEndpoint, selectedRegionalEndpoint);
+      const validWriteEndpoint = userContext.databaseAccount?.properties?.writeLocations?.find(
+        (loc) => loc.documentEndpoint === selectedRegionalEndpoint,
+      );
+      updateUserContext({
+        selectedRegionalEndpoint: selectedRegionalEndpoint,
+        writeEnabledInSelectedRegion: validWriteEndpoint ? true : false,
```
languy, line 345: nit: `writeEnabledInSelectedRegion: !!validWriteEndpoint` is probably the more elegant form.

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -274,6 +320,34 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
...
+        refreshCosmosClient: true,
+      });
+      useClientWriteEnabled.setState({ clientWriteEnabled: validWriteEndpoint ? true : false });
```
languy, line 348: nit:
```suggestion
      useClientWriteEnabled.setState({ clientWriteEnabled: !!validWriteEndpoint });
```

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -583,9 +661,39 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
               </AccordionPanel>
             </AccordionItem>
           )}
+          {userContext.apiType === "SQL" && userContext.authType === AuthType.AAD && (
```
languy, line 664: For Fabric, the client needs to go a specific endpoint that the Fabric workload can intercept.
```suggestion
          {userContext.apiType === "SQL" && userContext.authType === AuthType.AAD && !isFabric() && (
```

src/hooks/useKnockoutExplorer.ts
```diff
@@ -670,6 +678,37 @@ function updateAADEndpoints(portalEnv: PortalEnv) {
...
+    const validEndpoint = userContext.databaseAccount?.properties?.readLocations?.find(
+      (loc) => loc.documentEndpoint === storedRegionalEndpoint,
+    );
+    const validWriteEndpoint = userContext.databaseAccount?.properties?.writeLocations?.find(
+      (loc) => loc.documentEndpoint === storedRegionalEndpoint,
+    );
+    if (validEndpoint) {
+      updateUserContext({
+        selectedRegionalEndpoint: storedRegionalEndpoint,
+        writeEnabledInSelectedRegion: validWriteEndpoint ? true : false,
```
languy, line 693:
```suggestion
        writeEnabledInSelectedRegion: !!validWriteEndpoint,
```

src/hooks/useKnockoutExplorer.ts
```diff
@@ -670,6 +678,37 @@ function updateAADEndpoints(portalEnv: PortalEnv) {
...
+        refreshCosmosClient: true,
+      });
+      useClientWriteEnabled.setState({ clientWriteEnabled: validWriteEndpoint ? true : false });
```
languy, line 696:
```suggestion
      useClientWriteEnabled.setState({ clientWriteEnabled: !!validWriteEndpoint });
```

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: It is a per Cosmos DB account setting.  Specific to the regions present on a given account.  If switching account, then defaults back to global endpoint.

src/hooks/useKnockoutExplorer.ts
bogercraig: Modified to !!.

src/hooks/useKnockoutExplorer.ts
bogercraig: Modified to !!.

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: Added configContext.platform !== Platform.Fabric condition.

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: Modified to !!

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: Modified to !!.

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: oh.  ok.  I see that it reverts back to global if I swap from one account to another then back to the original account.  Ideally, that setting needs to persist.

bogercraig: Migrated to local state.  Probably need better way of signaling react state when all local state is removed than what I quickly added to deleteAllStates.  

Works now to where region selection persists between account selections then resets to global when all setting preferences reset.

Would be really nice to include the retry settings also in the local state.

src/Common/CosmosClient.ts
```diff
@@ -203,6 +207,7 @@ export function client(): Cosmos.CosmosClient {
     userAgentSuffix: "Azure Portal",
     defaultHeaders: _defaultHeaders,
     connectionPolicy: {
+      enableEndpointDiscovery: userContext.selectedRegionalEndpoint ? false : true,
```
languy, line 210:
```suggestion
      enableEndpointDiscovery: !userContext.selectedRegionalEndpoint,
```

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -143,6 +153,17 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
...
+  const [selectedRegionalEndpoint, setSelectedRegionalEndpoint] = useState<string>(
+    hasState({
+      componentName: AppStateComponentNames.SelectedRegionalEndpoint,
+      globalAccountName: userContext.databaseAccount?.name,
+    })
+      ? (loadState({
+          componentName: AppStateComponentNames.SelectedRegionalEndpoint,
+          globalAccountName: userContext.databaseAccount?.name,
+        }) as string)
+      : "",
```
languy, line 165: Isn't it better to have `undefined` to clearly show that it does not exist? (rather than it exists with an invalid value)

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
```diff
@@ -583,9 +686,41 @@ export const SettingsPane: FunctionComponent<{ explorer: Explorer }> = ({
               </AccordionPanel>
             </AccordionItem>
           )}
+          {userContext.apiType === "SQL" &&
+            userContext.authType === AuthType.AAD &&
+            configContext.platform !== Platform.Fabric && (
```
languy, line 691: Let's use `isFabric()`.
```suggestion
            !isFabric() && (
```

src/Shared/AppStatePersistenceUtility.ts
```diff
@@ -216,3 +220,12 @@ export const deleteSubComponentState = (
...
 };
+
+const resetInterfaceContext = (): void => {
+  updateUserContext({
+    selectedRegionalEndpoint: undefined,
+    writeEnabledInSelectedRegion: true,
+    refreshCosmosClient: true,
+  });
+  useClientWriteEnabled.setState({ clientWriteEnabled: true });
+};
```
languy, line 231: Let's avoid adding a dependency to `UserContext` in this file which should just be responsible for app state persistence.
Let's do this reset where we call `deleteAllStates()` in the Settings pane.

languy (changes requested):

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: Adding to latest commit with import.  Will resolve when committed.

src/Common/CosmosClient.ts
bogercraig: Adding to latest commit with import. Will resolve when committed.

src/Shared/AppStatePersistenceUtility.ts
bogercraig: Moving to call in settings pane.

src/Explorer/Panes/SettingsPane/SettingsPane.tsx
bogercraig: Functionally, undefined should work.

languy (approved):

jawelton74 (approved):

PR #2047, opened 2025-02-08
Remove blocking await on sample database

sunghyunkang1111:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2047?feature.someFeatureFlagYouMightNeed=true)

This PR addresses a couple of optimizations on the Data Explorer build and runtime.

1. Removed compress:false flag from webpack configuration. This increases the build time at npm run pack:prod by ~75-100%, but it will reduce the bundle size by ~100-150%
2. Removed blocking async/await on the reading sample database call at the initialization.
- This introduces a refresh of the sidebar after the explorer is loaded, but at least they do not have to wait until this HTTP request is completed.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -248,6 +248,7 @@ module.exports = function (_env = {}, argv = {}) {
         new TerserPlugin({
           terserOptions: {
             // These options increase our initial bundle size by ~5% but the builds are significantly faster and won't run out of memory
+            // Update 2/11/202: we are removing this flag as our bundles sizes grew so that it can remove dead and unreachable code with compromise of build time
             // compress: false,
             mangle: {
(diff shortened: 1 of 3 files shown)


asier-isayas (dismissed):

webpack.config.js
```diff
@@ -248,7 +248,7 @@ module.exports = function (_env = {}, argv = {}) {
         new TerserPlugin({
           terserOptions: {
             // These options increase our initial bundle size by ~5% but the builds are significantly faster and won't run out of memory
-            compress: false,
+            // compress: false,
```
jawelton74, line 251: Please update comment wrt why we're commenting this out

jawelton74 (approved):

PR #2053, opened 2025-02-11
Emulator Settings Configurable

nishthaAhujaa:


src/Platform/Emulator/emulatorConfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
jawelton74, line 1: I think this should be placed under /configs ? Maybe as emulator.json ?

jawelton74 (approved):

PR #2086, opened 2025-03-24
Fix input validation patterns for resource ids

jawelton74:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2086?feature.someFeatureFlagYouMightNeed=true)

This change addresses errors that were preventing id input fields from validating against invalid characters (e.g. forward and backward slashes). It affects id values for the following:

- Database
- Container
- Stored procedure
- User Defined Function 
- Trigger

In cases where the input element is not a child of a form, manual validation reporting is now triggered.

image


src/Explorer/Tabs/ScriptTabBase.ts
```diff
@@ -267,22 +267,8 @@ export default abstract class ScriptTabBase extends TabsBase implements ViewMode
...
-      return false;
-    }
-
-    const invalidEndCharacters = /.*[/?#\\ ]$/;
-    if (invalidEndCharacters.test(id)) {
-      return false;
-    }
-
-    return true;
+    const validIdRegex = /[^\/?#\\]*[^\/?# \\]/;
```
vchske, line 270: I'm wondering if we should put this method in a util and write some unit tests around it catching invalid characters.

vchske (approved):

vchske (approved):

JustinKol (approved):

PR #2088, opened 2025-03-25
Self Serve text render fix

asier-isayas:
[Preview this branch](https://ms.portal.azure.com/?selfservesource=https%3A%2F%2Fdataexplorer-preview.azurewebsites.net%2Fcommit%2F2875462bc64767c24273362a65350a828016e525%2FselfServe.html%3Ffeature.pr%3Dhttps%253A%252F%252Fgithub.com%252FAzure%252Fcosmos-explorer%252Fpull%252F2088%2523users%252Faisayas%252Fself-serve-fix-3#home)

Text rendering is currently broken in non local environments. 
image

Fix
image

### DIFF
--- a/src/SelfServe/SelfServe.tsx
+++ b/src/SelfServe/SelfServe.tsx
@@ -20,7 +20,7 @@ const loadTranslationFile = async (
   selfServeType?: SelfServeType,
 ): Promise<void> => {
   const language = i18n.languages[0];
-  let namespace: string;
+  let namespace: string; // className is used as a key to retrieve the localized strings
   let fileName: string;
   if (className instanceof SelfServeBaseClass) {
     fileName = `${selfServeType}.json`;
--- a/src/SelfServe/SelfServe.tsx
+++ b/src/SelfServe/SelfServe.tsx
@@ -11,13 +11,24 @@ import { updateUserContext } from "../UserContext";
 import { isInvalidParentFrameOrigin } from "../Utils/MessageValidation";
 import "./SelfServe.less";
 import { SelfServeComponent } from "./SelfServeComponent";
-import { SelfServeDescriptor } from "./SelfServeTypes";
+import { SelfServeBaseClass, SelfServeDescriptor } from "./SelfServeTypes";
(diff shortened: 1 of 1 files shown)


src/SelfServe/SelfServe.tsx
```diff
@@ -11,13 +11,24 @@ import { updateUserContext } from "../UserContext";
...
 initializeIcons();
 
-const loadTranslationFile = async (className: string): Promise<void> => {
+const loadTranslationFile = async (
+  className: string | SelfServeBaseClass,
+  selfServeType?: SelfServeType,
+): Promise<void> => {
   const language = i18n.languages[0];
-  const fileName = `${className}.json`;
+  let namespace: string;
```
jawelton74, line 23: Could you please add a comment here regarding how the classname is used as a key to retrieve the localized strings ? We really need to get away from using this.

jawelton74 (dismissed):

src/SelfServe/SelfServe.tsx
```diff
@@ -11,13 +11,24 @@ import { updateUserContext } from "../UserContext";
...
+  let fileName: string;
+  if (className instanceof SelfServeBaseClass) {
+    fileName = `${selfServeType}.json`;
+    namespace = className.constructor.name;
```
BChoudhury-ms, line 27: can we create a mapper for the className and Localization file ? Using classNames as localization fileNames is difficult to maintain and also not a good proposal.
If we create a mapper, then based on the namespace the mapper will give the localized file.

asier-isayas: Created an improvement item.

jawelton74 (approved):

src/SelfServe/SelfServe.tsx
jawelton74: To lend some extra context here, totally agree that a deeper fix is required to move away from using class/constructor names. We've created a backlog item to do that. This PR is addressing a more immediate issue with the strings not being displayed correctly.

vchske (approved):

PR #2101, opened 2025-04-10
Implement Sample data import for Fabric Home

languy:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2101)
* Clicking on FabricHome Sample Data button opens a dialog
* The dialog will check if the "SampleData" container already exists, otherwise it will create one and bulk insert documents.
* We use the Copilot's sample data. Dynamically imported, so it doesn't take space in the main bundle.
* The bulk insert logic has been decoupled from file import, so that it can be used independently
* The "App development" button opens: https://aka.ms/cosmosdbfabricsdk in a new tab.

Additional considerations for further changes:
* add unit tests
* created container uses default values for throughput and pkey. Consider optimising.

image
image
image
image
image


sevoku (dismissed):

sevoku:
@languy the https://aka.ms/cosmosdbfabricsdk URL is not registered, do we need to set it up?

languy:
> @languy the https://aka.ms/cosmosdbfabricsdk URL is not registered, do we need to set it up?

PM team will do it.

src/Explorer/SplashScreen/FabricHome.tsx
```diff
@@ -157,17 +161,25 @@ export const FabricHomeScreen: React.FC<SplashScreenProps> = (props: SplashScree
...
+          explorer={props.explorer}
+          databaseName={userContext.fabricContext?.databaseName}
+        />
+        <div className={styles.title} role="heading" aria-label={title}>
+          {title}
+        </div>
+        {getSplashScreenButtons()}
+        <div className={styles.footer}>
+          Need help?{" "}
+          <Link href="https://cosmos.azure.com/docs" target="_blank">
```
jawelton74, line 178: This link doesn't go anywhere ?

jawelton74 (dismissed):

src/Explorer/SplashScreen/FabricHome.tsx
sevoku: Hhhm, that's a good question, it's not new code... where is it supposed to lead to?

sevoku: Maybe https://learn.microsoft.com/azure/cosmos-db/? But then we should have an aka.ms link for this!

sevoku: Oh, my bad, that's only for Fabric

sevoku: fixed 95c9230b936dc824c6b21d3f7ea3c14aac1c6984

sevoku (approved):

PR #2116, opened 2025-04-22
Open VS Code via Data Explorer button

JustinKol:
Add button to open Cosmos DB Account in Visual Studio Code.

[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2116?feature.someFeatureFlagYouMightNeed=true)

This PR will introduce a new link in the Data Explorer to open a users Cosmos DB Account in Visual Studio Code. If the user has Visual Studio Code installed, clicking the `Open in VS Code` button will open the Cosmos DB Account in Visual Studio Code. If the user does not have Visual Studio Code installed, it will need to be downloaded first.

image


src/Explorer/Explorer.tsx
```diff
@@ -282,6 +282,24 @@ export default class Explorer {
...
+    const activeTab = useTabs.getState().activeTab;
+    const baseUrl = `vscode://ms-azuretools.vscode-cosmosdb?resourceId=${userContext.databaseAccount.id}`;
+    const vscodeUrl = activeTab
+      ? `${baseUrl}&database=${activeTab.collection.databaseId}&container=${activeTab.collection?.id()}`
+      : baseUrl;
+    const vscodeInsidersUrl = vscodeUrl.replace("vscode://", "vscode-insiders://");
+
+    const linkOpened =
+      (navigator.userAgent.includes("insiders") && window.open(vscodeInsidersUrl)) || window.open(vscodeUrl);
+    const vsCodeTimeout = setTimeout(
```
sevoku, line 295: This doesn't work for me, Chrome opens a new window, briefly shows the prompt to open the link in VS Code and then redirects to the download page directly, so I can't even reach the button before the prompt disappears:
image

src/Explorer/Explorer.tsx
```diff
@@ -282,6 +282,24 @@ export default class Explorer {
...
+  public openInVsCode(): void {
+    const activeTab = useTabs.getState().activeTab;
+    const baseUrl = `vscode://ms-azuretools.vscode-cosmosdb?resourceId=${userContext.databaseAccount.id}`;
+    const vscodeUrl = activeTab
+      ? `${baseUrl}&database=${activeTab.collection.databaseId}&container=${activeTab.collection?.id()}`
+      : baseUrl;
+    const vscodeInsidersUrl = vscodeUrl.replace("vscode://", "vscode-insiders://");
+
+    const linkOpened =
+      (navigator.userAgent.includes("insiders") && window.open(vscodeInsidersUrl)) || window.open(vscodeUrl);
```
sevoku, line 294: This opens a new window/tab, shouldn't it prompt directly?

src/Explorer/Explorer.tsx
```diff
@@ -282,6 +282,24 @@ export default class Explorer {
     }
   }
 
+  public openInVsCode(): void {
+    const activeTab = useTabs.getState().activeTab;
+    const baseUrl = `vscode://ms-azuretools.vscode-cosmosdb?resourceId=${userContext.databaseAccount.id}`;
```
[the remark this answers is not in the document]
JustinKol: I have corrected this

src/Explorer/Explorer.tsx
```diff
@@ -282,6 +282,24 @@ export default class Explorer {
...
+    const vscodeUrl = activeTab
+      ? `${baseUrl}&database=${activeTab.collection.databaseId}&container=${activeTab.collection?.id()}`
```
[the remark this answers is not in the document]
JustinKol: I have corrected this

src/Explorer/Explorer.tsx
```diff
@@ -282,6 +282,42 @@ export default class Explorer {
...
+    const downloadUrl = "https://code.visualstudio.com/download";
+
+    const baseUrl = `vscod://ms-azuretools.vscode-cosmosdb?resourceId=${resourceId}`;
+    const vscodeUrl = activeTab ? `${baseUrl}&database=${database}&container=${container}` : baseUrl;
+    const vscodeInsidersUrl = vscodeUrl.replace("vscode://", "vscode-insiders://");
+
+    const startTime = Date.now();
+    let hasRedirected = false;
+
+    const iframe = document.createElement("iframe");
```
sevoku, line 299: Even with an Iframe I think this is too error prone and browser behaviour is not really predictable. I'd recommend the following:
* Open a simple Dialog with following content:
   * Title "Opening Visual Studio Code"
   * Message "If nothing happens [install Visual Studio Code]("https://code.visualstudio.com/download") and try again.
   * Buttons: "Cancel" "Download Visual Studio Code"
* When Dialog opens, redirect to the VS Code URI (use current window, don't taget blank)
* If navigation succeeds, just close the dialog. If VS Code doesn't open, user can decide to go and download or cancel.

src/Explorer/Explorer.tsx
sevoku: This doesn't work with the new iframe approach either, now it doesn't even prompt but just opens the VS Code Website.

JustinKol: Sorry, looks like I pushed 'vscod' instead of 'vscode' for the url so it wouldn't detected VS Code. Thus always resulting in the download page always appearing.

src/Explorer/Explorer.tsx
JustinKol: this has changed

JustinKol:
> Hi @JustinKol, @sevoku asked me to take a look at this PR.
> 
> From my perspective, we shouldn't use the new iframe, there is a easier way - using . Take a look at my code, a asked Copilot help me with comments and explanation
> 
> ```ts
>   public openInVsCode(): void {
>     const activeTab = useTabs.getState().activeTab;
>     const resourceId = encodeURIComponent(userContext.databaseAccount.id);
>     const database = encodeURIComponent(activeTab?.collection?.databaseId);
>     const container = encodeURIComponent(activeTab?.collection?.id());
>     const downloadUrl = "https://code.visualstudio.com/download";
> 
>     const baseUrl = `vscode://ms-azuretools.vscode-cosmosdb?resourceId=${resourceId}`;
>     const vscodeUrl = activeTab ? `${baseUrl}&database=${database}&container=${container}` : baseUrl;
>     // Do not use the insiders link, especially in way like `vscodeInsidersUrl || vscodeUrl;` since it always true and always is vscodeInsidersUrl :)
>     // This build of VSCode is experimental and usually can't be supported by production.
> 
>     // Create invisible anchor tag
>     // if display will be "none" the click will not work
>     const link = document.createElement("a");
>     link.style.position = "absolute"; 
>     link.style.left = "-9999px";
>     link.style.top = "-9999px";
>     link.href = vscodeUrl;
>     document.body.appendChild(link);
> 
>     // Track if VS Code was successfully launched
>     const startTime = Date.now();
>     let hasRedirected = false;
>     let linkRemoved = false;
> 
>     // Safe link removal helper
>     const safelyRemoveLink = () => {
>       if (!linkRemoved) {
>         try {
>           document.body.removeChild(link);
>           linkRemoved = true;
>         } catch (e) {
>           // Link might already be removed, ignore
>         }
>       }
>     };
> 
>     // Create fallback timer
>     setTimeout(() => {
>       // If very little time has passed, VS Code likely isn't installed
>       if (!hasRedirected && Date.now() - startTime < 1200) {
>         hasRedirected = true;
>         window.open(downloadUrl, "_blank");
>         logConsoleInfo("VS Code not detected. Opening download page.");
>       }
>       safelyRemoveLink();
>     }, 1000);
> 
>     // Try to open VS Code by simulating a click
>     try {
>       // The click will work only in case it is a real click by user (not programmatically called) 
>       // Between user click and this click must NOT be any async actions 
>       // overwise it is not direct real click event 
>       link.click();
>     } catch (error) {
>       if (!hasRedirected) {
>         hasRedirected = true;
>         window.open(downloadUrl, "_blank");
>         logConsoleError(`Failed to open VS Code: ${getErrorMessage(error)}`);
>       }
>       safelyRemoveLink();
>     }
>   }
> ```

src/Explorer/Explorer.tsx
sevoku: yeah, the dialog now pops up as expected and stays there, but also the download page opens in a new tab after around 1s. So I need to close the new tab or open the previous tab to get back to the message and click the button (which opens now VS Code as expected).

src/Explorer/Explorer.tsx
sevoku: with your latest fix it ([6c5e7f2](https://github.com/Azure/cosmos-explorer/pull/2116/commits/6c5e7f21d7002d31e672fd983aa0d384c7e07d67)) still doesn't work quite right unfortunately: [see above](https://github.com/Azure/cosmos-explorer/pull/2116#discussion_r2064321069)

sevoku: Hhhm, also with the increased timeout (32adf7fde66030c69fc440a3dece0c0b6b7c6701) no luck, now though the download page never opens (tried uninstalling VS Code, then even nothing happens when I click the button) and the dialog is shown twice. I looked into how GitHub is doing it, and they basically show a div that gets removed if opening succeeded. TBH, let's go with the dialog, that'd be the simplest.

sevoku (changes requested):
This works much better now! Let's double check the wording and final behaviour with the PM team.

Meanwhile we have not published the updated VS Code extension yet, hence blocking this PR from merging until everything is ready.

src/Explorer/Explorer.tsx
```diff
@@ -304,7 +304,8 @@ export default class Explorer {
       isModal: true,
       title: ` Open your Cosmos DB account in Visual Studio Code`,
       subText: `Please ensure Visual Studio Code is installed on your device.
-      If you don't have it installed, please download it from the link below.`,
+      If you don't have it installed, please download it from the link below.
+      After installation, ensure you have the Azure Databases and Azure Resources extensions installed.`,
```
sevoku, line 308: This is not really necessary, VS Code will prompt to install these if they're not. Let's check with the PM team.

JustinKol: I had someone test it and it prompt them to install

sevoku (approved):
The VS Code extension has been published with support for this last week, so we're good to go. I have tested this locally but we'Re still waiting for a Portal deployment to enable the vscode: protocol.

sevoku (approved):

jawelton74 (approved):

PR #2143, opened 2025-05-13
Add query results, export to json/csv button

JustinKol:
Added a download button to save query results to json or CSV (excel) format with a default name of query-results.

[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2143?feature.someFeatureFlagYouMightNeed=true)

Options for Json and CSV:
image
image

CSV format:
image

JSON format:
image


src/Explorer/Tabs/QueryTab/ResultsView.tsx
```diff
@@ -47,6 +48,52 @@ const ResultsTab: React.FC<ResultsViewProps> = ({ queryResults, isMongoDB, execu
     await executeQueryDocumentsPage(firstItemIndex + itemCount - 1);
   };
 
+  const ExportResults: React.FC = () => {
+    const [exportFormat, setExportFormat] = useState<"csv" | "json">("json");
+
+    const handleExport = (): void => {
+      if (exportFormat === "csv") {
+        const csvData = queryResults.documents.map((doc) => Object.values(doc).join(",")).join("\n");
+        const csvHeader = Object.keys(queryResults.documents[0]).join(",") + "\n";
```
sevoku, line 57: `,` separator doesn't work properly Excel and different locales (such as German) and even less with other apps, we should have a separator setting (see what we have done for VS Code: https://github.com/microsoft/vscode-cosmosdb/pull/2627) and use `;` as default (which is more compatible with other apps like LibreOffice and Apple Numbers) or add a separator header if we want to support Excel only (first line header line in file should be `sep=,` which is only Compatible with Excel).

src/Explorer/Tabs/QueryTab/ResultsView.tsx
```diff
@@ -47,6 +48,52 @@ const ResultsTab: React.FC<ResultsViewProps> = ({ queryResults, isMongoDB, execu
...
+    return (
+      <div>
+        <Button
+          onClick={handleExport}
+          size="small"
+          appearance="transparent"
+          icon={<ArrowDownloadRegular />}
+          title="Download Query Results"
+        ></Button>
+        <select
```
sevoku, line 84: UX wise (also for a11y) the combo box should come first and the button behind.

sevoku (changes requested):

src/Explorer/Tabs/QueryTab/ResultsView.tsx
JustinKol: Done!

sunghyunkang1111 (dismissed):

asier-isayas (dismissed):

src/Explorer/Tabs/QueryTab/ResultsView.tsx
```diff
@@ -47,6 +48,117 @@ const ResultsTab: React.FC<ResultsViewProps> = ({ queryResults, isMongoDB, execu
...
+        const allHeadersSet = new Set<string>();
+        queryResults.documents.forEach((doc) => {
+          Object.keys(doc).forEach((key) => allHeadersSet.add(key));
+        });
+        const allHeaders = Array.from(allHeadersSet);
+        const csvHeader = allHeaders.join(",");
+        // Map each document to a row using all headers, filling missing fields with empty string
+        const csvData = queryResults.documents
+          .map((doc) =>
+            allHeaders.map((header) => (doc[header] !== undefined ? JSON.stringify(doc[header]) : "")).join(","),
```
sevoku, line 66: I think we'd need to escape double quotes here, otherwise CSV parsers will get confused if the actual value contains a quote (JSON.stringify will escape quotes with a slash while valid quote escape is another quote as in `"`>`""`, see [RFC spec section 2.7 (page 2)](https://datatracker.ietf.org/doc/html/rfc4180#section-2)). For reference here is our CSV converter for VS Code: https://github.com/microsoft/vscode-cosmosdb/blob/main/src/utils/csvConverter.ts.

src/Explorer/Tabs/QueryTab/ResultsView.tsx
sevoku: Just to confirm, are we sure we want to support Excel only? I'd rather use `;` as separator by default which works for Apple Numbers and LibreOffice on other platforms, maybe let's double check with the PM team?

sevoku (approved):
Nice, this works much better now :)

src/Explorer/Tabs/QueryTab/ResultsView.tsx
sevoku: [a2a7761](https://github.com/Azure/cosmos-explorer/pull/2143/commits/a2a77617f6b99db8c8a32711bab7183c0a885a7c)

JustinKol: Fixed

PR #2158, opened 2025-05-19
Better VSCode detection

JustinKol:


sevoku (commented):
This now opens VS Code twice for me when I click the button.

src/Explorer/Explorer.tsx
```diff
@@ -292,32 +292,87 @@ export default class Explorer {
...
+            window.removeEventListener("blur", onBlur);
+            if (document.body.contains(iframe)) {
+              document.body.removeChild(iframe);
+            }
+            // Window lost focus, likely because VS Code opened
+            resolve(true);
+          };
+          window.addEventListener("blur", onBlur, { once: true });
+          // Use a test URL that won't open a specific resource but will check if the extension is installed
+          iframe.src = "vscode://ms-azuretools.vscode-cosmosdb?test=1";
```
sevoku, line 323: This can't work, VS Code will load the extension and pass the "test=1" param to our handler which will fail of course. This results in an additional VS Code prompt:
image
Which then fails:
image
Even if we would handle the test parameter and ignore it, VS Code will still prompt twice unless the user has opted in to not being asked again to handle external URLs.

PR #2176, opened 2025-06-18
Support data plane RBAC for E2E tests.

jawelton74:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2176?feature.someFeatureFlagYouMightNeed=true)

This change adds support for running the playwright E2E tests against accounts that have enabled data plane RBAC access.
This is enabled for the CI action. To support this, we acquire access tokens prior to running the tests (msal can't be used within playwright to acquire them during the test)  and pass them to Explorer via DataExplorerInputsFrame.


test/testExplorer/TestExplorer.ts
```diff
@@ -10,17 +10,40 @@ const subscriptionId = urlSearchParams.get("subscriptionId") || process.env.SUBS
...
-    authorizationToken: `bearer ${token}`,
+    authorizationToken: `bearer ${authToken}`,
   });
+
   const databaseAccount = await get(subscriptionId, resourceGroup, accountName);
+  const tags = databaseAccount?.tags;
+  const testAccountType = tags && tags["DataExplorer:TestAccountType"];
+
+  let rbacToken = "";
+  switch (testAccountType) {
```
vchske, line 31: The way testAccountType is set up, it is expected that it could be undefined. Do you need a default case here?

vchske (approved):

test/testExplorer/TestExplorer.ts
jawelton74: No, I don't believe so. In any other case we want rbacToken to remain as an empty string.

vchske (approved):

sindhuba (approved):

PR #2198, opened 2025-08-05
Enable RBAC support for MongoDB and Cassandra APIs

BChoudhury-ms:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2198?feature.someFeatureFlagYouMightNeed=true)

### Work Items
- Mongo API: https://msdata.visualstudio.com/CosmosDB/_workitems/edit/4534014
- Cassandra API: https://msdata.visualstudio.com/CosmosDB/_workitems/edit/3912589

### Summary

This PR extends Role-Based Access Control (RBAC) support to MongoDB and Cassandra APIs in Azure Cosmos DB Data Explorer, enabling Entra ID authentication for these API types alongside the existing SQL, Tables, and Gremlin APIs.

### Changes
- Adds the `x-ms-entraid-token` header for requests when Data Plane RBAC is enabled for MongoDB and Cassandra APIs.
- Updates logic in `MongoProxyClient`, `TableDataClient` and related components to conditionally include the AAD token header.
- Refactors and extends utility functions to support RBAC detection for proxy APIs.
   - Extended `isDataplaneRbacSupported` function to include MongoDB and Cassandra APIs
   - Added `hasProxyServer` utility function to identify APIs that use proxy servers
   - Created `isDataplaneRbacEnabledForProxyApi` function to determine RBAC enablement for proxy-based APIs
   - Updated `authHeaders` function to include Entra ID token when RBAC is enabled
   - Modified `setAuthorizationHeader` to include Entra ID token for Cassandra operations
- Updated token acquisition logic to use `https://cosmos.azure.com/.default` scope for proxy APIs
Modified both `acquireMsalTokenForAccount` and `configureHostedWithAAD` functions
- Optimized command bar button creation to respond to RBAC state changes and added proper dependency tracking for dataplane RBAC and AAD token updates
- Updates shell handlers and connection commands to use AAD tokens when RBAC is enabled, including changes to the Mongo shell connection logic.
- Improves status message handling and prompt detection in the terminal add-on for more robust output processing.
- Updates tests to cover new RBAC scenarios, including AAD token usage and shell command generation.
- Minor fixes and improvements, such as updating the mongosh version and test timeouts.

### Breaking Changes
- The command bar button creation is now memoized and depends on RBAC state changes for Mongo & Cassandra APIs. Also need to check for other APIs aswell

### Testing Scenarios

| Scenario | Description | Solution | Result |
|----------|-------------|----------|--------|
| 1 | RBAC Enable false | No Entra ID should be in the header. Proxy server should use keys approach. | Pass |
| 2 | RBAC Enable true, and user has no access | Entra ID should be in the header. Proxy server should use Mongo SDK with Entra ID as credential. If user lacks permission, SDK should throw an error. Proxy doesn’t decide. | Pass |
| 3 | RBAC Enable true, and user has access | Entra ID should be in the header. Proxy server should use Mongo SDK with Entra ID as credential. SDK processes the request as per user permissions. Proxy doesn’t decide. | Pass |
| 4 | RBAC Enable automatic, user may or may not have access | DE should determine RBAC enablement using `disableLocalAuth`. If true, RBAC is true; if false, RBAC is false. Entra ID should flow accordingly. | Pass |
| 5 | `disableLocalAuth` is true | Error: Timeout still occurs. | ❌ |

### Flow
image

### DIFF
--- a/src/hooks/useKnockoutExplorer.ts
+++ b/src/hooks/useKnockoutExplorer.ts
@@ -1,4 +1,5 @@
 import * as Constants from "Common/Constants";
+import { getEnvironmentScopeEndpoint } from "Common/EnvironmentUtility";
 import { createUri } from "Common/UrlUtility";
 import { DATA_EXPLORER_RPC_VERSION } from "Contracts/DataExplorerMessagesContract";
 import { FabricMessageTypes } from "Contracts/FabricMessageTypes";
@@ -334,7 +335,7 @@ async function configureHostedWithAAD(config: AAD): Promise<Explorer> {
   if (account.properties?.documentEndpoint) {
     let hrefEndpoint = "";
     if (isDataplaneRbacEnabledForProxyApi(userContext)) {
-      hrefEndpoint = new URL("https://cosmos.azure.com/").href.replace(/\/+$/, "/.default");
+      hrefEndpoint = getEnvironmentScopeEndpoint();
     } else {
       hrefEndpoint = new URL(account.properties.documentEndpoint).href.replace(/\/$/, "/.default");
     }
--- a/src/Common/Constants.ts
+++ b/src/Common/Constants.ts
@@ -138,6 +138,14 @@ export enum MongoBackendEndpointType {
   remote,
 }
 
+export class AadScopeEndpoints {
+  public static readonly Development: string = "https://cosmos.azure.com";
+  public static readonly MPAC: string = "https://cosmos.azure.com";
+  public static readonly Prod: string = "https://cosmos.azure.com";
+  public static readonly Fairfax: string = "https://cosmos.azure.us";
+  public static readonly Mooncake: string = "https://cosmos.azure.cn";
+}
+
 export class PortalBackendEndpoints {
   public static readonly Development: string = "https://localhost:7235";
   public static readonly Mpac: string = "https://cdb-ms-mpac-pbe.cosmos.azure.com";
--- a/src/Common/EnvironmentUtility.test.ts
+++ b/src/Common/EnvironmentUtility.test.ts
@@ -28,3 +28,39 @@ describe("Environment Utility Test", () => {
     expect(EnvironmentUtility.getEnvironment()).toBe(EnvironmentUtility.Environment.Development);
   });
 });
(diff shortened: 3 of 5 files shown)


src/hooks/useKnockoutExplorer.ts
```diff
@@ -331,7 +332,12 @@ async function configureHostedWithAAD(config: AAD): Promise<Explorer> {
   const resourceGroup = accountResourceId && accountResourceId.split("resourceGroups/")[1].split("/")[0];
   let aadToken;
   if (account.properties?.documentEndpoint) {
-    const hrefEndpoint = new URL(account.properties.documentEndpoint).href.replace(/\/$/, "/.default");
+    let hrefEndpoint = "";
```
jawelton74, line 336: This code is repeated in AuthorizationUtils - can we create a common function ?

src/hooks/useKnockoutExplorer.ts
```diff
@@ -331,7 +332,12 @@ async function configureHostedWithAAD(config: AAD): Promise<Explorer> {
...
+    if (isDataplaneRbacEnabledForProxyApi(userContext)) {
+      hrefEndpoint = new URL("https://cosmos.azure.com/").href.replace(/\/+$/, "/.default");
```
jawelton74: cosmos.azure.com is for public cloud. How about Mooncake/Fairfax etc ?

src/hooks/useKnockoutExplorer.ts
BChoudhury-ms: updated

src/hooks/useKnockoutExplorer.ts
BChoudhury-ms: added

src/hooks/useKnockoutExplorer.ts
BChoudhury-ms: The typescript compiler does fail in the pipeline, when we do try to create a common function for the complete logic.  For now, I have created a common function only for the new code
**Note:** _Will keep this improvement in my pending workitem book, and will work again on this when my current work plate is clear_

PR #2200, opened 2025-08-12
Added platform feature utility

sunghyunkang1111:

### DIFF
--- a/src/Explorer/Explorer.tsx
+++ b/src/Explorer/Explorer.tsx
@@ -3,7 +3,7 @@ import { Link } from "@fluentui/react/lib/Link";
 import { isPublicInternetAccessAllowed } from "Common/DatabaseAccountUtility";
 import { Environment, getEnvironment } from "Common/EnvironmentUtility";
 import { sendMessage } from "Common/MessageHandler";
-import { Platform, configContext } from "ConfigContext";
+import { configContext, Platform } from "ConfigContext";
 import { MessageTypes } from "Contracts/ExplorerContracts";
 import { useDataPlaneRbac } from "Explorer/Panes/SettingsPane/SettingsPane";
 import { getCopilotEnabled, isCopilotFeatureRegistered } from "Explorer/QueryCopilot/Shared/QueryCopilotClient";
@@ -18,6 +18,7 @@ import { LocalStorageUtility, StorageKey } from "Shared/StorageUtility";
 import { acquireMsalTokenForAccount } from "Utils/AuthorizationUtils";
 import { allowedNotebookServerUrls, validateEndpoint } from "Utils/EndpointUtils";
 import { featureRegistered } from "Utils/FeatureRegistrationUtils";
+import { isFeatureSupported, PlatformFeature } from "Utils/PlatformFeatureUtils";
 import { update } from "Utils/arm/generatedClients/cosmos/databaseAccounts";
 import { useQueryCopilot } from "hooks/useQueryCopilot";
 import * as ko from "knockout";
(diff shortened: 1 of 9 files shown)


src/Explorer/Explorer.tsx
```diff
@@ -1205,7 +1205,12 @@ export default class Explorer {
       await this.initNotebooks(userContext.databaseAccount);
     }
 
-    if (userContext.authType === AuthType.AAD && userContext.apiType === "SQL" && !isFabricNative()) {
+    if (
+      isFeatureSupported(PlatformFeature.ThroughputBucketing) &&
+      userContext.authType === AuthType.AAD &&
+      userContext.apiType === "SQL" &&
+      !isFabricNative()
```
bk201-: I think `isFabricNative` is one of the checks to understand what is supported or not. 
It is like CSS checks, instead of parsing UA everyone has to check css rules supportabilities. 

Therefore I would like to use different mappings for every platform: Fabric, Azure, etc. 

It removes any checks of platform and moves code to feature-driven way. Each platform will have their own supported feature. However, I'm not familiar with entire project code, probably this approach can't be implemented.

src/Utils/PlatformFeatureUtils.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param platform Optional platform override, defaults to current platform
+ * @returns True if the feature is enabled for the platform, false otherwise
+ */
+export const isFeatureSupported = (feature: PlatformFeature, platform?: Platform): boolean => {
+  const currentPlatform = platform ?? configContext.platform;
+
+  // Portal and Hosted platforms have all features enabled by default
+  if (currentPlatform === Platform.Portal || currentPlatform === Platform.Hosted) {
+    return true;
+  }
```
languy: Let's not assume Portal and Hosted have all features.
Fabric has specific features that don't apply (for example a different Home tab).

src/Utils/PlatformFeatureUtils.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  return PLATFORM_FEATURE_MATRIX[currentPlatform]?.has(feature) ?? false;
+};
+
+export const areScriptsSupported = (platform?: Platform): boolean => {
+  const currentPlatform = platform ?? configContext.platform;
+  if (currentPlatform === Platform.Portal || currentPlatform === Platform.Hosted) {
+    return true;
+  }
```
languy: Why not remove this and add the 3 script features to each Portal and Hosted. 
This way we have only one source of truth where the features are defined and no additional special logic that turns on/off features.

src/Common/DatabaseAccountUtility.ts
```diff
@@ -30,6 +30,8 @@ export function getWorkloadType(): WorkloadType {
 
 export function isGlobalSecondaryIndexEnabled(): boolean {
   return (
-    !isFabric() && userContext.apiType === "SQL" && userContext.databaseAccount?.properties?.enableMaterializedViews
```
languy: The Fabric changes are significant and could be done independently. Can make the changes so that the Fabric behavior stays as-is in this PR and we do all the Fabric changes in a subsequent PR.

src/Explorer/Explorer.tsx
languy: @bk201- If we define the feature mapping correctly in `PlatformFeatureUtils.ts` for a new `FabricNative` platform, we will be able to remove this `!isFabricNative()` check. Do you agree or are you suggesting something else?

src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.tsx
```diff
@@ -352,7 +354,7 @@ export function createScriptCommandButtons(selectedNodeState: SelectedNodeState)
   const buttons: CommandButtonComponentProps[] = [];
 
   const shouldEnableScriptsCommands: boolean =
-    !selectedNodeState.isDatabaseNodeOrNoneSelected() && areScriptsSupported();
+    areScriptsSupported() && !selectedNodeState.isDatabaseNodeOrNoneSelected();
```
languy: No functional change. Revert?

languy (approved):

PR #2211, opened 2025-09-08
Removed screenshot for mongo cloudshell

nishthaAhujaa:

### DIFF
--- a/src/Explorer/Tabs/ShellAdapters/BaseTerminalComponentAdapter.tsx
+++ b/src/Explorer/Tabs/ShellAdapters/BaseTerminalComponentAdapter.tsx
@@ -4,7 +4,6 @@ import { QuickstartFirewallNotification } from "Explorer/Quickstart/QuickstartFi
 import { getShellNameForDisplay } from "Explorer/Tabs/CloudShellTab/Utils/CommonUtils";
 import * as React from "react";
 import FirewallRuleScreenshot from "../../../../images/firewallRule.png";
-import VcoreFirewallRuleScreenshot from "../../../../images/vcoreMongoFirewallRule.png";
 import { ReactAdapter } from "../../../Bindings/ReactBindingHandler";
 import * as DataModels from "../../../Contracts/DataModels";
 import * as ViewModels from "../../../Contracts/ViewModels";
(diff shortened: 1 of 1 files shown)


src/Explorer/Tabs/CloudShellTab/Utils/RegionUtils.tsx
```diff
@@ -41,8 +41,8 @@ export const getNormalizedRegion = (region: string, defaultCloudshellRegion: str
   }
 
   const regionMap: Record<string, string> = {
-    centralus: "westcentralus",
-    eastus2: "eastus",
+    eastus2: "eastus2euap",
+    centralus: "centraluseuap",
```
jawelton74: This is mapping prod regions to canary regions ? Why are we changing this ?

src/Explorer/Tabs/ShellAdapters/BaseTerminalComponentAdapter.tsx
```diff
@@ -25,15 +24,15 @@ export abstract class BaseTerminalComponentAdapter implements ReactAdapter {
   ) {}
 
   public renderComponent(): JSX.Element {
+    if (this.kind === ViewModels.TerminalKind.Mongo || this.kind === ViewModels.TerminalKind.VCoreMongo) {
```
jawelton74, line 27: Shouldn't we add this clause to line 31 instead ?

Something like 

!this.isAllPublicIPAddressesEnabled() && this.kind !== ViewModels.TerminalKind.Mongo || this.kind !== ViewModels.TerminalKind.VCoreMongo

nishthaAhujaa: this has an operator precedence issue -the second part this.kind !== ViewModels.TerminalKind.VCoreMongo would be true for Mongo terminals (since Mongo RU ≠ VCoreMongo), causing them to show firewall notifications when we don't want

The early return pattern cleanly separates concerns: Mongo terminals skip all checks entirely, while other terminals follow the firewall logic.

src/Explorer/Tabs/CloudShellTab/Utils/RegionUtils.tsx
nishthaAhujaa: The existing mapping is incorrect and there are no valid pools in westcentralus, in the existing production also
centralus/westcentralus cloudshell would fail
Spoke to BE team about it, we can recheck this again
but definitely  centralus: "westcentralus", this mapping is incorrect
https://msazure.visualstudio.com/One/_git/CloudShell-UX?path=/src/components/Onboarding/UserSettings.tsx&version=GBmain&_a=contents&line=144&lineStyle=plain&lineEnd=144&lineStartColumn=4&lineEndColumn=69

jawelton74 (dismissed):

sakshigupta12feb (approved):

PR #2217, opened 2025-09-23
feat: Enhance restore container functionality and UI updates

bk201-:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2217?feature.someFeatureFlagYouMightNeed=true)

Menu will be shown only if the feature flag passed from parent iframe

image


src/Explorer/ContextMenuButtonFactory.tsx
```diff
@@ -60,6 +61,17 @@ export const createDatabaseContextMenu = (container: Explorer, databaseId: strin
     },
   ];
 
+  if (isFabric() && !userContext.fabricContext?.isReadOnly) {
```
languy: Let's check for `isFabricNative()`, because it won't work in mirroring.

bk201-: Done

sevoku (approved):

nishthaAhujaa (approved):

src/Platform/Fabric/FabricUtil.ts
```diff
@@ -105,6 +105,12 @@ const requestAndStoreAccessToken = async (): Promise<void> => {
   });
 };
 
+export const openRestoreContainerDialog = (): void => {
+  if (configContext.platform === Platform.Fabric) {
```
languy: Can/should we use `isFabricNative()` instead?

PR #2249, opened 2025-11-07
Removed broken feature to display PR url in DE console

vchske:
This removes a control which shows the link back to the PR which is supposed to show when running DE as a result of clicking on the Preview link (like the one in this PR :) ). It would look like this:
image

However, this probably doesn't look familiar as I don't know when this last work, if ever. Because the URL to github wasn't being validated, it was marked as a security issue so I took the whole feature out. I decided to remove because I don't think this would add any value to anyone.

[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2249)


preview/index.js
```diff
@@ -57,11 +56,7 @@ app.get("/pull/:pr(\\d+)", (req, res) => {
 
   fetch(`${githubApiUrl}/pulls/${pr}`)
     .then((response) => response.json())
-    .then(({ head: { ref, sha } }) => {
```
vchske: I took this part out because, if it wouldn't be used anyway, I didn't want another security case to be opened down the line. I don't think that would happen, but better to not leave cruft anyway.

PR #2251, opened 2025-11-11
Removed old, unused functionality

sakshigupta12feb:
Removed unused code


webpack.config.js
```diff
@@ -113,7 +113,6 @@ module.exports = function (_env = {}, argv = {}) {
     hostedExplorer: "./src/HostedExplorer.tsx",
     terminal: "./src/Terminal/index.ts",
     cellOutputViewer: "./src/CellOutputViewer/CellOutputViewer.tsx",
-    notebookViewer: "./src/NotebookViewer/NotebookViewer.tsx",
```
BChoudhury-ms, line 116: Thanks for identifying the old and unused code. It would be even better if we could also find and remove the dependent files (e.g., _NotebookViewer.tsx and other sibling or child components_).
**Please confirm, if this refactor is not required**

sakshigupta12feb: The refactor is NOT safe to proceed. NotebookViewer has active dependencies across the application

PR #2254, opened 2025-11-12
Fixes an issue where tab titles were not truncating when characters used 4 bytes for encoding

vchske:
Not all characters are encoded unto utf-8/utf-16 using the same number of bytes. ASCII uses just one byte, while extended character sets can be encoded using two, three, or four bytes.

Currently, when databases' or containers' names are displayed in the tab title, their names could be truncated if their names are longer than 8 characters. The truncation currently uses something like `coll.title.slice(0, 5)`. But if one of those characters are encoded using four bytes this happens:
image

This PR adds a new string util function which gets the desired number of characters regardless of the number of bytes used to encode each character. Like this:
image

[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2254)

### DIFF
--- a/src/Utils/StringUtils.ts
+++ b/src/Utils/StringUtils.ts
@@ -35,18 +35,37 @@ export const substringUtf = (text: string, startChar: number, numChars: number) 
     if (currentChar === startChar) {
       startByte = currentByte;
     }
-    switch (encoded[currentByte]) {
-      case 196:
+
+    /*
+    Unicode is utf encoded using 1, 2, 3, or 4 bytes
+    In a byte array, we know how many bytes the character is encoded based on the first byte because it
+    was developed such that the first byte's range never occurs in any other byte. Subsequent bytes are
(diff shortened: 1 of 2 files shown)


asier-isayas (dismissed):

src/Utils/StringUtils.ts
```diff
@@ -17,3 +17,39 @@ export function endsWith(stringToTest: string, suffix: string): boolean {
...
+
+  let currentChar = 0;
+  let currentByte = 0;
+  let startByte = 0;
+  for (; currentChar < startChar + numChars; ) {
+    if (currentChar === startChar) {
+      startByte = currentByte;
+    }
+    switch (encoded[currentByte]) {
+      case 196:
```
jawelton74: Please add comments to explain what these integers represent

vchske: Added

jawelton74 (approved):

PR #2257, opened 2025-11-13
Added health metrics for application load and database load

sunghyunkang1111:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2257?feature.someFeatureFlagYouMightNeed=true)

## Monitoring/Metrics on DE
The goal of this PR is to have a wireframe of the health and general monitoring and metrics of Data Explorer.
Discussion was done on this [documentation](https://microsoft-my.sharepoint.com/:w:/p/sunghyunkang/IQBpunEAA3iOQrf9otMTTzawAY_hcFo81puXLuOclrVHqvM?e=FdQPNy)

Simply, the events that will be emitted for monitoring has few parts.
- Scenario - A user's full e2e journey in the application
- Phase - A granualar broken down technical steps within the scenario that will act as a checkpoint to make progress and capture timing data
- Health Signal - Binary outcomes
- Other metadata - Captures other metadata from user's sessions.

Example Scenario configuration for "ApplicationLoad" Scenario
```
export const scenarioConfigs: Record<MetricScenario, ScenarioConfig> = {
  [MetricScenario.ApplicationLoad]: {
    requiredPhases: [ApplicationMetricPhase.ExplorerInitialized, CommonMetricPhase.Interactive],
    timeoutMs: 10000,
  },
};
```
This means that ApplicationLoad has 2 phases, `ApplicationMetricPhase.ExplorerInitialized` and  `CommonMetricPhase.Interactive` and a timeout of 10 seconds.
It is the developers responsibility to
1. Start the Scenario at the starting point
2. Set the phases completed/failed within the retrospective workflow within the React code
3. If above doesn't happen, it will timeout within the specified timeout and then emit Unhealthy metrics to the backend, then Geneva.

Example metrics snapshot object for "ApplicationLoad" Scenario
```
{
    "scenario": "ApplicationLoad",
    "startTimeISO": "2025-11-19T00:46:51.959Z",
    "endTimeISO": "2025-11-19T00:46:54.393Z",
    "durationMs": 2433.6000000089407,
    "completed": [
        "ExplorerInitialized",
        "Interactive"
    ],
    "timedOut": false,
    "vitals": {
        "ttfb": 7.5,
        "fcp": 5932
    },
    "phaseTimings": {
        "ExplorerInitialized": {
            "endTimeISO": "2025-11-19T00:46:54.363Z",
            "durationMs": 2403.7000000029802
        },
        "Interactive": {
            "endTimeISO": "2025-11-19T00:46:54.392Z",
            "durationMs": 2433
        }
    }
}
```

### Notes and potential discussions
- Currently this is configured to stop and emit failure metrics as soon as 1 of the phaes are marked fail or scenario hasn't been completed within the specificed timeout duration.
- Currently phases are NOT ordered or doesn't need to follow specific ordering, as long as they are all marked, the Scenario will emit healthy metric

### DIFF
--- a/src/Explorer/Tree/Database.test.tsx
+++ b/src/Explorer/Tree/Database.test.tsx
@@ -6,6 +6,15 @@ import { updateUserContext, userContext } from "../../UserContext";
 import Explorer from "../Explorer";
 import Database from "./Database";
 
+jest.mock("rx-jupyter", () => ({
+  sessions: {
+    create: jest.fn(),
+  },
+  contents: {
+    JupyterContentProvider: jest.fn().mockImplementation(() => ({})),
+  },
+}));
+
 const createMockContainer = (): Explorer => {
   const mockContainer = new Explorer();
   return mockContainer;


src/Metrics/MetricScenarioProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  startPhase: (scenario: MetricScenario, phase: MetricPhase) => void;
+  completePhase: (scenario: MetricScenario, phase: MetricPhase) => void;
+}
+
+const MetricScenarioContext = React.createContext<MetricScenarioContextValue | undefined>(undefined);
+
+export const MetricScenarioProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  const startScenario = useCallback((s: MetricScenario) => scenarioMonitor.start(s), []);
+  const startPhase = useCallback((s: MetricScenario, p: MetricPhase) => scenarioMonitor.startPhase(s, p), []);
+  const completePhase = useCallback((s: MetricScenario, p: MetricPhase) => scenarioMonitor.completePhase(s, p), []);
```
BChoudhury-ms: **Suggestion only**
We might not need `useCallback` here as these are simple functions and recreation of these does not give much performance penalty. Removing it will simplify the code and keep it more readable .

src/Main.tsx
```diff
@@ -79,6 +83,18 @@ const App: React.FunctionComponent = () => {
...
+  const { startScenario, completePhase } = useMetricScenario();
+  React.useEffect(() => {
+    startScenario(MetricScenario.ApplicationLoad);
+  }, [startScenario]);
+
+  React.useEffect(() => {
+    if (explorer) {
+      completePhase(MetricScenario.ApplicationLoad, ApplicationMetricPhase.ExplorerInitialized);
+    }
+  }, [explorer, completePhase]);
```
BChoudhury-ms: **Suggestion only:**
If we remove `useCallback` from context then we can avoid adding these here also as dependency.

If we don't expect these functions to trigger expensive side effects or frequent renders, removing `useCallback` and simplifying the `useEffect` dependencies could be a better choice.

src/Metrics/ScenarioMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  startPhase(scenario: MetricScenario, phase: MetricPhase) {
+    const ctx = this.contexts.get(scenario);
+    if (!ctx || ctx.emitted) {
+      return;
+    }
+    if (!ctx.config.requiredPhases.includes(phase)) {
+      return;
+    }
```
BChoudhury-ms: we can combine these `return` statements along with `ctx.phases.has(phase)`. And then we can also remove the `if(!ctx.phases.has(phase))` block outside to make more readable

src/Metrics/ScenarioMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    // Mark the phase as completed
+    const phaseCtx = ctx.phases.get(phase);
+    if (!phaseCtx) {
+      return;
+    }
```
BChoudhury-ms: same as like above ☝️ we can combine these return statements

src/Metrics/ScenarioMonitor.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (!allDone) {
+      return;
+    }
+    const finalSnapshot = this.buildSnapshot(ctx, { final: true, timedOut: false });
+    const healthy = ctx.config.validate ? ctx.config.validate(finalSnapshot) : true;
+    this.emit(ctx, healthy, false, finalSnapshot);
+  }
+
+  private getPhaseTimings(ctx: InternalScenarioContext): Record<string, PhaseTimings> {
+    const result: Record<string, PhaseTimings> = {};
```
BChoudhury-ms, line 151: **Suggestion only**
we can cache the result and do an early return logic. This can avoid multiple `performance.getEntriesByName` and `performance.measure` calls. _(But not sure about the business logic)_

src/Main.tsx
```diff
@@ -19,7 +19,7 @@ import "../externals/jquery.dataTables.min.css";
 import "../externals/jquery.typeahead.min.css";
 import "../externals/jquery.typeahead.min.js";
 // Image Dependencies
-import { Platform } from "ConfigContext";
+import { Platform } from "ConfigContext"; // configContext no longer needed directly (ScenarioMonitor handles platform)
```
vchske: This is small but I don't understand this comment. If the import is no longer needed, do we need the import? If we still need the import, I would suggest explaining that

src/Metrics/MetricEvents.test.ts
```diff
@@ -0,0 +1,104 @@
+import { configContext, Platform } from "../../ConfigContext";
```
vchske: Kind of a nit but in this codebase, unit tests aren't separated into a different folder. This should probably just be called MetricEvents.text.ts and live at the same level as MetricEvents.ts

src/Metrics/MetricScenarioProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <MetricScenarioContext.Provider value={{ startScenario, startPhase, completePhase }}>
+      {children}
+    </MetricScenarioContext.Provider>
+  );
+};
+
+export function useMetricScenario(): MetricScenarioContextValue {
```
vchske, line 23: Just curious, if <App /> is wrapped by MetricScenarioProvider, when would the context NOT be available? Wouldn't any component be under that node?

sunghyunkang1111: Right, I mean it shouldn't happen. But also you can essentially import useMetricScenario hook in any components that is not wrapped in provider, and that would cause runtime errors

PR #2258, opened 2025-11-16
Refactor Container Copy Jobs for Intra-account copy and Online operations

BChoudhury-ms:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2258?feature.someFeatureFlagYouMightNeed=true)

This PR is to implement the feedbacks given by Product Team and other engineers

**Changes implemented**
- Removed the validation screen for offline intra-account copy operations to streamline the process.
- Introduced a button to handle online operations, replacing the need for manual CLI commands.
- Implemented functionality to reset the validation cache when source account changes.
- Enhanced the logic to handle operations for the same account more effectively.
- No Job actions for Failed Jobs
- Show Complete Job action for completed jobs
- Uplift error to context level

### DIFF
--- a/src/Explorer/ContainerCopy/ContainerCopyMessages.ts
+++ b/src/Explorer/ContainerCopy/ContainerCopyMessages.ts
@@ -50,7 +50,7 @@ export default {
   assignPermissions: {
     crossAccountDescription:
       "To copy data from the source to the destination container, ensure that the managed identity of the destination account has read access to the source account by completing the following steps.",
-    intraAccountOnlieDescription: (accountName: string) =>
+    intraAccountOnlineDescription: (accountName: string) =>
       `Follow the steps below to enable online copy on your "${accountName}" account.`,
   },
   toggleBtn: {
--- a/src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/AddReadPermissionToDefaultIdentity.tsx
+++ b/src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/AddReadPermissionToDefaultIdentity.tsx
@@ -1,5 +1,6 @@
 import { Link, Stack, Text, Toggle } from "@fluentui/react";
 import React, { useCallback } from "react";
+import { logError } from "../../../../../Common/Logger";
 import { assignRole } from "../../../../../Utils/arm/RbacUtils";
 import ContainerCopyMessages from "../../../ContainerCopyMessages";
 import { useCopyJobContext } from "../../../Context/CopyJobContext";
@@ -48,8 +49,10 @@ const AddReadPermissionToDefaultIdentity: React.FC<AddReadPermissionToDefaultIde
         }));
       }
     } catch (error) {
-      console.error("Error assigning read permission to default identity:", error);
-      setContextError(error.message || "Error assigning read permission to default identity. Please try again later.");
+      const errorMessage =
+        error.message || "Error assigning read permission to default identity. Please try again later.";
+      logError(errorMessage, "CopyJob/AddReadPermissionToDefaultIdentity.handleAddReadPermission");
+      setContextError(errorMessage);
     } finally {
       setLoading(false);
     }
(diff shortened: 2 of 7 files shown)


src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/AddReadPermissionToDefaultIdentity.tsx
```diff
@@ -49,10 +49,11 @@ const AddReadPermissionToDefaultIdentity: React.FC<AddReadPermissionToDefaultIde
       }
     } catch (error) {
       console.error("Error assigning read permission to default identity:", error);
```
sunghyunkang1111: Can you use logError from Logger class unless you want to swallow the error?
It will send the error to portal telemetry.

sunghyunkang1111 (changes requested):

src/Explorer/ContainerCopy/ContainerCopyMessages.ts
```diff
@@ -48,8 +48,10 @@ export default {
 
   // Assign Permissions Screen
   assignPermissions: {
-    description:
+    crossAccountDescription:
       "To copy data from the source to the destination container, ensure that the managed identity of the destination account has read access to the source account by completing the following steps.",
+    intraAccountOnlieDescription: (accountName: string) =>
```
asier-isayas: `Online`?

src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/OnlineCopyEnabled.tsx
```diff
@@ -56,18 +60,41 @@ const OnlineCopyEnabled: React.FC = () => {
...
-  React.useEffect(() => {
-    intervalRef.current = setInterval(() => {
-      handleFetchAccount();
-    }, 30 * 1000);
+  const handleOnlineCopyEnable = async () => {
+    setLoading(true);
+    setShowRefreshButton(false);
+
+    try {
+      await updateDatabaseAccount(sourceSubscriptionId, sourceResourceGroup, sourceAccountName, {
```
vchske, line 77: I don't like this pattern of updating the account multiple times in a single user operation. What happens if they close the browser while it's awaiting the first update? Or their computer shuts down. Whatever the cause, it could lead them getting into a funky state. But, more importantly, we are turning the browser into a stateful client. The Azure pattern is the client is definitely NOT stateful (I also happen to think web apps should not be stateful anyway). In Azure, the RP/backend is the stateful one and we call it to get what state the account is in. Doing it any other way gets the portal into trouble--believe me, I've seen the tickets :)

So, we should strive for a pattern of one operation per user action. If the account/db/container couldn't do the final action before the first, and if the first action only makes sense to do in preparation for the second, or if both of these action need rto be done to prep for a third action, then this whole workflow should have been handled entirely by the RP and this just wasn't caught in the design phase. If that's not the case, we should walk them through the flow of push this button and wait, then when the RP reports that's done show them the 2nd step, etc. I'm doing exactly this in the Replication blade rewrite.

The only exception is if the save is synchronous and would not start a LRO.

src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/AddReadPermissionToDefaultIdentity.tsx
BChoudhury-ms: added `logError`, as suggested.

src/Explorer/ContainerCopy/ContainerCopyMessages.ts
BChoudhury-ms: updated

src/Explorer/ContainerCopy/CreateCopyJob/Screens/AssignPermissions/OnlineCopyEnabled.tsx
BChoudhury-ms: Highlighted this point to PM

PR #2263, opened 2025-11-24
Add Session Id

asier-isayas:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2263)

In order to better filter logs in the Portal Backend, we need to filter based on session id so that we can track the logs specific to a user's session. 

image

### DIFF
--- a/src/Explorer/Panes/SettingsPane/SettingsPane.test.tsx
+++ b/src/Explorer/Panes/SettingsPane/SettingsPane.test.tsx
@@ -5,6 +5,12 @@ import { updateUserContext } from "../../../UserContext";
 import { SettingsPane } from "./SettingsPane";
 
 describe("Settings Pane", () => {
+  beforeEach(() => {
+    updateUserContext({
+      sessionId: "1234-5678",
+    });
+  });
+
   it("should render Default properly", () => {
     const wrapper = shallow(<SettingsPane explorer={null} />);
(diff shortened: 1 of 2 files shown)


vchske (dismissed):

src/Contracts/ViewModels.ts
```diff
@@ -446,6 +446,7 @@ export interface DataExplorerInputsFrame {
   feedbackPolicies?: any;
   aadToken?: string;
   containerCopyEnabled?: boolean;
+  sessionId?: string;
```
BChoudhury-ms, line 449: **Query**
1. Since we are updating the `sessionId` to a **UUID v4** value at the very beginning of the explorer, shouldn't we define the property as required?
2. Also is there any case, where `sessionId` is not available ?

asier-isayas: 1. Changed to required. 
2. No, I moved the sessionId to init of `UserContext `in this [commit](https://github.com/Azure/cosmos-explorer/commit/d99e3a9070af30dc883e7ec2fc223380a7997143).

asier-isayas: Misread your question. I am leaving that as optional because in not all cases we are passing sessionId from Portal (for example, Postgres and vCore Mongo).

BChoudhury-ms (approved):

PR #2298, opened 2025-12-31
Fix pagination issue

sindhuba:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2298?feature.someFeatureFlagYouMightNeed=true)


test/CORSBypass.ts
```diff
@@ -2,20 +2,54 @@ import { Page } from "@playwright/test";
 
 export async function setupCORSBypass(page: Page) {
   await page.route("**/api/mongo/explorer{,/**}", async (route) => {
+    const request = route.request();
```
jawelton74, line 5: Why do we need to do this for the pagination changes ?

sindhuba: The x-ms-continuation header must be explicitly added as part of the request headers to ensure the continuation token is transmitted correctly in cross‑region requests, since certain browsers like Firefox do not allow it by default.

jawelton74: Got it, thanks !

sunghyunkang1111 (approved):

PR #2308, opened 2026-01-06
Copy jobs dark theme

sakshigupta12feb:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2308?feature.someFeatureFlagYouMightNeed=true)

https://github.com/user-attachments/assets/d89bfd54-7096-42e5-ab84-998e45972446

This PR enables dark theme support for the Container Copy Jobs feature when running in Portal mode and makes the theme toggle button always visible in the command bar.

### DIFF
--- a/src/Explorer/ContainerCopy/CommandBar/Utils.test.ts
+++ b/src/Explorer/ContainerCopy/CommandBar/Utils.test.ts
@@ -105,7 +105,7 @@ describe("CommandBar Utils", () => {
       }));
 
       const { getCommandBarButtons: getCommandBarButtonsEmulator } = await import("./Utils");
-      const buttons = getCommandBarButtonsEmulator(mockExplorer);
+      const buttons = getCommandBarButtonsEmulator(mockExplorer, false);
 
       expect(buttons.length).toBe(3);
     });


src/Explorer/ContainerCopy/MonitorCopyJobs/Components/CopyJobStatusWithIcon.tsx
```diff
@@ -35,6 +19,17 @@ const iconMap: Partial<Record<CopyJobStatusType, string>> = {
   [CopyJobStatusType.Completed]: "CompletedSolid",
 };
 
+// Icon colors for different statuses
+const statusIconColors: Partial<Record<CopyJobStatusType, string>> = {
+  [CopyJobStatusType.Failed]: "var(--colorPaletteRedForeground1)",
+  [CopyJobStatusType.Faulted]: "var(--colorPaletteRedForeground1)",
+  [CopyJobStatusType.Completed]: "#107c10", // Green color for success
```
nishthaAhujaa: can we use a css variable here like others?

src/Explorer/ContainerCopy/CommandBar/Utils.ts
```diff
@@ -11,6 +14,7 @@ import { CopyJobCommandBarBtnType } from "../Types/CopyJobTypes";
 
 function getCopyJobBtns(explorer: Explorer): CopyJobCommandBarBtnType[] {
   const monitorCopyJobsRef = MonitorCopyJobsRefState((state) => state.ref);
+  const isDarkMode = useThemeStore.getState().isDarkMode;
```
nishthaAhujaa: Maybe use a subscription based approach (e.g., the useThemeStore React hook) instead of getState(). so, the UI automatically updates when the theme changes

src/Common/Pager/index.tsx
```diff
@@ -31,6 +31,7 @@ const iconButtonStyles = {
     outline: "none",
   },
 };
+const textStyle: React.CSSProperties = { color: "var(--colorNeutralForeground1)" };
```
BChoudhury-ms: Is it possible to write a class-based code for `color: "var(--colorNeutral........)"` similar to `.themeText`?

BChoudhury-ms: It’s a best practice to centralize common styles into reusable CSS classes. 

```css
.themeText            { color: var(--colorNeutralForeground1); }
.themeTextSecondary   { color: var(--colorNeutralForeground2); }
.themeLinkText        { color: var(--colorBrandForeground1); }
.themeBackground      { background-color: var(--colorNeutralBackground1); }
.themeBackgroundSecondary { background-color: var(--colorNeutralBackground2); }
....
```

nishthaAhujaa (dismissed):

src/Common/Pager/index.tsx
BChoudhury-ms: Is this approach not possible? My suggestion is to apply it to all of the inline colors.

sakshigupta12feb: this was not working in few cases so, there I have used inline,
will try to cover that in next PR.

languy (approved):

BChoudhury-ms (approved):

PR #2312, opened 2026-01-08
Searchable dropdown

nishthaAhujaa:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2312?feature.someFeatureFlagYouMightNeed=true)

https://github.com/user-attachments/assets/ba5e70ef-4858-494a-bf9b-4b55b791c5e7

https://github.com/user-attachments/assets/6c00a5bd-1e7d-4358-9bfa-3c48b5428cb0

We have added search in the subscription and account name dropdown; in the recent change we have added the search icon and the text changes.


src/Common/SearchableDropdown.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const buttonLabel = selectedItem
+    ? getDisplayText(selectedItem)
+    : items?.length === 0
+    ? `No ${label}s Found`
+    : placeholder;
+
+  const buttonId = `${className}-button`;
+
+  return (
+    <div>
```
BChoudhury-ms: Can we use **Fluent UI** components like `Stack`, `Button`, `Text` etc. here?

src/Common/SearchableDropdown.tsx
```diff
@@ -0,0 +1,194 @@
...
+        ref={buttonRef}
+        className={className}
+        onClick={handleButtonClick}
+        style={buttonStyles}
+        disabled={disabled}
+      >
+        <span style={{ overflow: "hidden", textOverflow: "ellipsis", whiteSpace: "nowrap", display: "block" }}>
+          {buttonLabel}
+        </span>
+        <span
```
BChoudhury-ms: Instead of using inline CSS objects, we can move all styles into a separate file.

_Same suggestion applies to all the files_

src/Platform/Hosted/Components/SwitchSubscription.tsx
```diff
@@ -1,6 +1,6 @@
-import { Dropdown } from "@fluentui/react";
```
BChoudhury-ms, line 1: Because `SwitchSubscription` is shared with the Container Copy screen, we should ensure its behavior remains consistent.

src/Common/SearchableDropdown.tsx
```diff
@@ -0,0 +1,194 @@
+import { Callout, DirectionalHint, ISearchBoxStyles, Label, SearchBox } from "@fluentui/react";
```
BChoudhury-ms: Since `SearchableDropdown.tsx` is a new component, could you please add unit tests for it?

src/Platform/Hosted/Components/SwitchAccount.tsx
```diff
@@ -1,6 +1,6 @@
-import { Dropdown } from "@fluentui/react";
 import * as React from "react";
```
BChoudhury-ms, line 1: Because `SwitchAccount.tsx` is shared with the Container Copy screen, we should ensure its behavior remains consistent.

src/Common/SearchableDropdown.styles.ts
```diff
@@ -0,0 +1,78 @@
...
+import * as React from "react";
+
+export const getDropdownButtonStyles = (disabled: boolean): IButtonStyles => ({
+  root: {
+    width: "100%",
+    height: "32px",
+    padding: "0 28px 0 8px",
+    border: "1px solid #8a8886",
+    background: "#fff",
+    color: "#323130",
```
BChoudhury-ms, line 11: Color codes should be dark-theme compatible. Since these are generic components, we need to ensure they support dark theme usage.

src/Common/SearchableDropdown.tsx
```diff
@@ -0,0 +1,158 @@
...
+  className,
+  disabled = false,
+  onDismiss,
+  searchBoxStyles: customSearchBoxStyles,
+}: SearchableDropdownProps<T>): React.ReactElement => {
+  const [isOpen, setIsOpen] = useState(false);
+  const [filterText, setFilterText] = useState("");
+  const buttonRef = useRef<HTMLDivElement>(null);
+
+  const closeDropdown = useCallback(() => {
```
BChoudhury-ms: Minor: Simple events do not need to be wrapped with useCallback. The same applies to other simpler event handlers in this file, as doing so increases React runtime memoization overhead without providing any significant performance benefits.

src/Common/SearchableDropdown.styles.ts
sakshigupta12feb: We'll update the remaining hardcoded values when we do the broader Fluent UI v9 token migration.

sakshigupta12feb (approved):

PR #2471, opened 2026-04-29
Added quantizerType support in vector embedding

sunghyunkang1111:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2471?feature.someFeatureFlagYouMightNeed=true)

Adds the ability to configure a quantizer type when creating or editing vector indexes that use quantization (quantizedFlat and diskANN). Users can now choose between Product and Spherical (Preview) quantization.

image

Changes
1. New Quantizer type dropdown in the vector embedding policy editor, available wherever vector indexes are configured (New Container panel and the container's Scale & Settings tab).
2. The dropdown is enabled only for index types that support quantization, and is hidden/disabled for flat indexes.
3. The selected value is included in create and update payloads, and the existing value is read back and displayed when viewing an existing container's policy.
4. Settings tab tracks edits to vector indexes as part of dirty-state detection so saves and discards behave correctly.
Added a tooltip explaining the difference between the two quantizer options.
5. Updated localized strings for the new field and its tooltip.

### DIFF
--- a/src/Explorer/Controls/VectorSearch/VectorEmbeddingPoliciesComponent.tsx
+++ b/src/Explorer/Controls/VectorSearch/VectorEmbeddingPoliciesComponent.tsx
@@ -17,6 +17,7 @@ import {
   getDistanceFunctionOptions,
   getIndexTypeOptions,
   getQuantizerTypeOptions,
+  supportsQuantization,
 } from "Explorer/Controls/VectorSearch/VectorSearchUtils";
 import { Keys, t } from "Localization";
 import React, { FunctionComponent, useState } from "react";
@@ -159,7 +160,7 @@ export const VectorEmbeddingPoliciesComponent: FunctionComponent<IVectorEmbeddin
     vectorEmbeddings?.forEach((embedding) => {
       const matchingIndex = displayIndexes ? vectorIndexes.find((index) => index.path === embedding.path) : undefined;
       const matchingType = matchingIndex?.type;
-      const supportsQuantizer = matchingType === "quantizedFlat" || matchingType === "diskANN";
+      const supportsQuantizer = supportsQuantization(matchingType);
       mergedData.push({
         ...embedding,
         indexType: matchingType || "none",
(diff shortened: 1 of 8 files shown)


src/Explorer/Controls/VectorSearch/VectorEmbeddingPoliciesComponent.tsx
```diff
@@ -155,11 +158,14 @@ export const VectorEmbeddingPoliciesComponent: FunctionComponent<IVectorEmbeddin
     const mergedData: VectorEmbeddingPolicyData[] = [];
     vectorEmbeddings?.forEach((embedding) => {
       const matchingIndex = displayIndexes ? vectorIndexes.find((index) => index.path === embedding.path) : undefined;
+      const matchingType = matchingIndex?.type;
+      const supportsQuantizer = matchingType === "quantizedFlat" || matchingType === "diskANN";
```
jawelton74: You make this check in a few places - worth making into a function in case we need to change it later ?

src/Explorer/Controls/VectorSearch/VectorSearchUtils.ts
```diff
@@ -7,6 +7,10 @@ const indexTypes = ["none", "flat", "diskANN", "quantizedFlat"];
 export const getDataTypeOptions = (): IDropdownOption[] => createDropdownOptionsFromLiterals(dataTypes);
 export const getDistanceFunctionOptions = (): IDropdownOption[] => createDropdownOptionsFromLiterals(distanceFunctions);
 export const getIndexTypeOptions = (): IDropdownOption[] => createDropdownOptionsFromLiterals(indexTypes);
+export const getQuantizerTypeOptions = (): IDropdownOption[] => [
+  { key: "product", text: "Product" },
```
jawelton74, line 13: should these be localized ?

sunghyunkang1111: I don't think thesee should be, as they are the technical terms

jawelton74 (approved):

asier-isayas (approved):

PR #2499, opened 2026-05-26
Fix stored XSS via Cassandra column names in DataTable headers

asier-isayas:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2499)

### Problem
Cassandra column names are passed directly to DataTables `sTitle`, which renders them as HTML. An attacker can create a column named `` to execute JavaScript in any victim's browser session when they view the table.

### Fix
HTML-encode column names with `Utilities.htmlEncode()` before passing to `sTitle`, same as cell values are already encoded in `bindColumn`.

### DIFF
--- a/src/Explorer/Tables/DataTable/DataTableBindingManager.ts
+++ b/src/Explorer/Tables/DataTable/DataTableBindingManager.ts
@@ -93,7 +93,7 @@ function createDataTable(
 
   for (var i = 0; i < tableEntityListViewModel.headers.length; i++) {
     jsonColTable.push({
-      sTitle: tableEntityListViewModel.headers[i],
+      sTitle: Utilities.htmlEncode(tableEntityListViewModel.headers[i]),
       data: tableEntityListViewModel.headers[i],
       aTargets: [i],
       mRender: bindColumn,


sunghyunkang1111 (approved):

src/Explorer/Tables/DataTable/DataTableBindingManager.ts
```diff
@@ -93,7 +93,7 @@ function createDataTable(
 
   for (var i = 0; i < tableEntityListViewModel.headers.length; i++) {
     jsonColTable.push({
-      sTitle: tableEntityListViewModel.headers[i],
+      sTitle: Utilities.htmlEncode(tableEntityListViewModel.headers[i]),
       data: tableEntityListViewModel.headers[i],
```
jawelton74, line 97: Do we need to do the same for data ?

asier-isayas: No, data is not rendered as HTML. It's a column key that's used to read values from each row object.

jawelton74 (approved):

PR #2552, opened 2026-08-06
Fix casing of Cosmos DB Shell command bar button label

mkrueger:
## Summary

The Cosmos DB Shell command bar button label was rendered as "Open Cosmos DB shell" (lowercase "shell"). This corrects it to "Open Cosmos DB **Shell**". Not sure about the button name for other shells - so I leave it as is. Depending on the shell name it may look weird.

## Change

In `createOpenTerminalButtonByKind` ([CommandBarComponentButtonFactory.tsx](src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.tsx)), the label is now special-cased for `TerminalKind.CosmosDB` to use a capital "Shell", while the other shell types (Cassandra, PSQL, MongoDB (DocumentDB)) keep their existing lowercase "shell" labels (which their unit tests assert).

```ts
const label =
  terminalKind === ViewModels.TerminalKind.CosmosDB
    ? `Open ${terminalFriendlyName()} Shell`
    : `Open ${terminalFriendlyName()} shell`;
```

## Testing

- `npm run compile` — no type errors in the modified file.
- No existing test asserts the Cosmos DB button label, so no snapshot/test updates were required.


src/Explorer/Menus/CommandBar/CommandBarComponentButtonFactory.tsx
```diff
@@ -513,7 +513,10 @@ function createOpenTerminalButtonByKind(
         return "";
     }
   };
-  const label = `Open ${terminalFriendlyName()} shell`;
+  const label =
+    terminalKind === ViewModels.TerminalKind.CosmosDB
+      ? `Open ${terminalFriendlyName()} Shell`
+      : `Open ${terminalFriendlyName()} shell`;
```
sevoku: Shouldn't this always be `Shell`?

mkrueger: Maybe, but I only want to change it for my button. The upper case "Shell" is basically part of our branding.

Changing that for all shells would be an out of scope side effect.

mkrueger: ok makes sense to use Shell everywhere.

sevoku (approved):

jawelton74 (approved):

PR #2556, opened 2026-08-10
Disabled quantize byte size when the quantizer type does not support it

vchske:
[Preview this branch](https://dataexplorer-preview.azurewebsites.net/pull/2556)

This PR address 1 functional issue and 2 stylistic issues.

Functionally, as of now, when the quantizer type is anything but "Product", the quantization byte size field should be empty and raises an exception in the backend if it is not.

Stylistically, the quantization byte size field was moved below the quantizer type field since the size is dependent on the type and it is a bad user experience to change/disable a control above some selection (flow should follow the same rules as reading).

Additionally, the fields when disabled gave no visual indication they were disabled, so this change makes the label for disabled fields more distinctly disabled.

Old flow:
Selecting am index type and quantization type that supports byte size:
image

Changing to a type that does not support byte size, byte size remains:
image

It also remains if you choose an index type that does not support quantization:
image

New flow:
No index type is chosen so no properties are enabled and are more visually telling of being disabled:
image

Choosing an index type that supports quantization, the type defaults to "Spherical" which does not support byte size:
image

Choosing "Product" enables the byte size field:
image

Choosing a different index type which also supports quantization keeps the quantization fields intact:
image

Choosing "Spherical" disables byte size and clears the value:
image

Alternatively, choosing an index type that does not support quantization, disables both quantization fields:
image

### DIFF
--- a/src/Explorer/Controls/VectorSearch/VectorEmbeddingPoliciesComponent.tsx
+++ b/src/Explorer/Controls/VectorSearch/VectorEmbeddingPoliciesComponent.tsx
@@ -2,6 +2,7 @@ import {
   DefaultButton,
   Dropdown,
   IDropdownOption,
+  ILabelStyleProps,
   IStyleFunctionOrObject,
   ITextFieldStyleProps,
   ITextFieldStyles,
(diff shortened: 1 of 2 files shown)


sunghyunkang1111 (approved):
Nice!

jawelton74 (approved):

PR #2559, opened 2026-08-11
Add client-side connection-string login for SQL, Tables, and Gremlin

asier-isayas:
## Summary

Connection‑string login for **SQL, Tables, and Gremlin** now authenticates **entirely in the browser** using the account key, instead of routing through the Portal Backend. Mongo and Cassandra are unchanged and continue to use the existing encrypted‑token proxy path.

## Why

Previously, every connection‑string login went to the Portal Backend to exchange the connection string for an encrypted token (`/generatetoken`), fetch account metadata (`/accessinputmetadata`), and sign each data‑plane request (`/authorizationtokens`). For SQL, Tables, and Gremlin this round trip is unnecessary — the Cosmos client can sign requests locally with the account key. Removing it means fewer backend dependencies, fewer network hops, and faster sign‑in for those APIs.

## What changed

- The connection string is parsed **in the browser** to derive the account endpoints, so we no longer call the backend's `accessinputmetadata` API.
- For SQL and Gremlin the document endpoint is taken directly from the `AccountEndpoint` in the connection string. Table connection strings only carry the table endpoint, so the document endpoint is derived from the account name.
- The Cosmos client signs each request locally with the master key (via its existing `tokenProvider`), so `generatetoken` and per‑request `authorizationtokens` calls are skipped.
- Mongo and Cassandra still flow through the encrypted‑token proxy exactly as before.

## User‑visible behavior

- **SQL / Tables / Gremlin:** Sign in with a connection string with no backend calls.
- The connection string is accepted as given and Data Explorer opens immediately. A bad key or an account that blocks the user's IP is reported by the operation that hits it once Data Explorer is open.

## Testing

- Unit tests updated/added across `HostedExplorer`, `ConnectExplorer`, `ConnectionStringParser`, and `HostedUtils`, including:
  - SQL/Tables/Gremlin sign in client‑side without calling the backend.
  - The document endpoint is taken from the connection string rather than reconstructed.
- E2E tests were added for SQL, Tables, and Gremlin connection string login, plus a case covering login with an incorrect account key, which opens Data Explorer rather than failing at the connect form.


test/fx.ts
```diff
@@ -1,8 +1,15 @@
...
 import { Frame, Locator, Page, expect } from "@playwright/test";
-import crypto from "crypto";
+import crypto, { webcrypto } from "crypto";
 import { TestContainerContext } from "./testData";
 
+// The @azure/cosmos client signs requests with globalThis.crypto (Web Crypto API).
+// In Node.js >= 19 it's already available; only assign the polyfill for older versions.
+// This lives in fx.ts (imported by every spec) so the polyfill always runs.
+if (!globalThis.crypto) {
+  Object.defineProperty(globalThis, "crypto", { value: webcrypto, writable: true, configurable: true });
```
asier-isayas, line 10: This was moved from `testData.ts` to here.

src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -41,12 +52,43 @@ export const isAccountRestrictedForConnectionStringLogin = async (connectionStri
   return (await response.text()).toLowerCase() === "true";
 };
 
+// Verifies the account key can actually authenticate against the account by making a lightweight read of
+// the database account through the Cosmos client. Returns undefined on success, or an error message when
+// the client cannot connect.
+export const validateDirectConnectionStringConnectivity = async (
+  connectionString: string,
+  metadata: AccessInputMetadata,
+): Promise<string | undefined> => {
```
vchske: Been thinking about this kind of scenario lately as I think I've made quite a few mistakes with them. Now I think that returning undefined as a success code is probably not the right semantic. I'm leaning these days to returning a tuple: { isValid: boolean; errorMessage?: string }
If you don't like that, maybe instead rename the method to better indicate intent like checkDirectConnectionStringForError or something.
Whichever thing to do depends on how you expect possible other clients to call this method.
What do you think?

src/hooks/useKnockoutExplorer.ts
```diff
@@ -467,13 +467,25 @@ function configureHostedWithConnectionString(config: ConnectionString): Explorer
...
+    // Direct client-side signing path (SQL, Tables, Gremlin). Requests are signed locally with the
+    // account key via the Cosmos client's tokenProvider, so no Portal Backend proxy or encrypted token
+    // is required.
+    updateUserContext({
+      authType: AuthType.ConnectionString,
+      databaseAccount,
+      masterKey: config.masterKey,
+    });
+  } else {
+    // Legacy encrypted-token proxy path (Mongo, Cassandra).
```
vchske, line 480: Nit: can you gel this comment a bit better with the next line of comment?

src/Platform/Hosted/HostedUtils.ts
```diff
@@ -45,3 +46,114 @@ export function extractMasterKeyfromConnectionString(connectionString: string):
...
+
+  return undefined;
+}
+
+// Client-side equivalent of the Portal Backend's ConnectionStringAccessProvider.ValidateHostAndAccount,
+// scoped to the direct-login APIs (SQL, Tables, Gremlin). Ensures the connection string parses to an
+// account name, exposes an endpoint host in an allowlisted DNS zone, that the account name matches the
+// first DNS label of that host, and that an account key is present. Returns an error message when
+// invalid, or undefined when the connection string is valid for direct login.
+export function validateDirectConnectionStringLogin(
```
vchske: Same comment here as my previous one

src/Platform/Hosted/Components/ConnectExplorer.tsx
asier-isayas: Changed to `void` and throws on error.

src/hooks/useKnockoutExplorer.ts
asier-isayas: Removed next line of comment

src/Platform/Hosted/HostedUtils.ts
asier-isayas: Changed to `void` and throws on error.

sunghyunkang1111 (dismissed):

vchske (approved):

sunghyunkang1111 (approved):

PR #2568, opened 2026-08-26
Support Sovereign/PPE endpoint zones for connection string login + E2E test for SQL account with public network access disabled

asier-isayas:
## Support sovereign and PPE endpoint zones for connection string login

Client-side connection string login for SQL/Table/Gremlin (#2559) hardcoded the public-cloud DNS zones, so it only worked against `*.azure.com` accounts.

### DNS zones moved into `ConfigContext`

`EndpointsRegex` in `Constants.ts` is replaced by seven configurable zone lists: `SQL_DNS_ZONES`, `MONGO_DNS_ZONES`, `MONGO_COMPUTE_DNS_ZONES`, `CASSANDRA_DNS_ZONES`, `TABLE_DNS_ZONES`, plus `DOCUMENT_ENDPOINT_ZONES` and `GREMLIN_ENDPOINT_ZONES` for endpoints the parser constructs. Values come from the Cosmos DB's `Environment.*.json` settings.

Because the zone list decides where an account key is sent, the matching was tightened:

- Zones must run to the end of the host. `(.*)` previously let `Test.documents.azure.com.attacker.example` pass as a `documents.azure.com` account.
- Constructed endpoints must match the kind of zone that matched. A PPE account gets PPE endpoints; if config has no zone of that kind, the connection string is rejected rather than building an endpoint the account doesn't own.

`documentEndpoint` is now derived for Mongo, and Cassandra, not just Table. SQL and Gremlin still take it verbatim from the connection string.

### E2E test for public network access disabled

- Adds `TestAccount.SQLConnectionStringPublicNetworkAccessDisabled` and a test in `test/sql/connectionStringLogin.spec.ts`.
- The account has `publicNetworkAccess` disabled, so connection-string login succeeds but database access fails.
- Asserts the connect screen closes, no inline error appears, the notification console shows `Error while refreshing databases`, and only the `Home` node appears in the resource tree.

#### Error handling

If the connection string is not recognized, an inline error is displayed.

image <br>

The Mongo/Cassandra path previously had no `catch`, so a `fetchEncryptedToken` failure resulted in an unhandled rejection with no message. It now reads the rejected response and displays an inline error. A `403` means the account firewall dropped the request. A link directs users where they can add Portal middleware IPs to their firewall rules. _(This change will not show until Portal Backend auth changes for Mongo and Cassandra are released)_

image

<br>

#### Connect button

While connecting, the **Connect** button changes to **Connecting...** and is disabled to indicate progress and prevent duplicate submissions.

image


src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -43,39 +46,70 @@ export const ConnectExplorer: React.FunctionComponent<Props> = ({
...
+                  return;
+                }
+
                 setErrorMessage("");
+                setIsBlockedByFirewall(false);
+                setIsConnecting(true);
 
                 try {
-                  if (await isAccountRestrictedForConnectionStringLogin(connectionString)) {
+                  if (isResourceTokenConnectionString(connectionString)) {
```
jawelton74, line 75: this changes the validation so that a resource token cxn string can pass before we check if the account is restricted.

vchske (dismissed):

src/Platform/Hosted/Components/ConnectExplorer.tsx
```diff
@@ -58,6 +58,18 @@ export const ConnectExplorer: React.FunctionComponent<Props> = ({
                 setIsConnecting(true);
 
                 try {
+                  try {
```
jawelton74: having nested try clauses is pretty messy. can we move the account restrictions check to before the first try statement ?

jawelton74 (approved):

languy (approved):

PR #2590, opened 2026-09-10
Fix Data Explorer health scenarios reporting false unhealthy loads

bk201-:
`ScenarioMonitor.completePhase()` silently no-oped when the phase had not been started yet, and the production call ordering in `Explorer.tsx` / `ResourceTree.tsx` always hits that case. `DataExplorerHealthV2` therefore reported loads as unhealthy that had in fact completed.

## Evidence

Over three days, `scenario_timeout` events:

| Scenario | Timeouts | Dominant missing phase | Share |
|---|---|---|---|
| `DatabaseLoad` | 1627 | `DatabaseTreeRendered` | **1624 (99.8%)** |
| `ApplicationLoad` | 649 | `PlatformConfigured` + `ExplorerInitialized` | **649 (100%)** |

`documentHidden` was true in 692 of 2276 (30%) — a contributing factor, not the cause; the same pattern dominates in foreground tabs.

## Root cause

**`DatabaseTreeRendered`** is started at `Explorer.tsx:451`, immediately after `completePhase(CollectionsLoaded)`. The effect that completes it (`useMetricPhases.ts:55-59`) re-fires only on a `databaseTreeNodes` change, and `databaseTreeNodes` is memoised (`ResourceTree.tsx:54`). The last change happens *before* line 451, so the effect never re-fires and the phase stays open until the 10s timeout. This is a deterministic ordering inversion, not a race — which is why it is 99.8% and not 50%.

**`Interactive`** is completed by a one-shot effect whose deps (`[scenario, enabled]`) are stable. If `ResourceTree` mounts before `Explorer.tsx:577` starts the scenario, the completion is dropped and never retried. Missing in 588 of 1627.

**`ApplicationLoad`** never got past `configurePortal()`, which resolves only from the iframe `message` handler with no timeout and no reject. When the portal does not post the init message the promise stays pending forever — the user sits on `<LoadingExplorer />` and the scenario times out with no diagnostic. Both phases hang off the same `await`, which is why they are always missing together.

## Changes

- `completePhase` self-starts a required phase that is completed before `startPhase`, instead of dropping the completion.
- Completions reported before the scenario exists are buffered and replayed on `start()`.
- A timeout while the tab is backgrounded no longer counts against health. Browsers throttle timers and suspend rAF there, so phase completion is unreliable and the elapsed time is not the user's experience. `documentHidden` is still reported, so these remain sliceable in telemetry.
- `configurePortal()` rejects after 30s with a `ConfigurePortal` failure trace — deliberately well above the 10s scenario budget so only genuinely stuck handshakes fail. The caller now handles the rejection instead of leaving it unhandled.

## Tests

`src/Metrics/ScenarioMonitor.phaseOrdering.test.ts` replays the exact production call ordering and asserts it now completes healthy, plus: an early completion before `start()`, a genuine stall still reporting unhealthy, and a backgrounded-tab timeout not counting against health.

Locally `src/Metrics/` is 53 passed (4 new + 49 existing), and `tsc --noEmit` reports no errors in `src/`.

## Notes for reviewers

- **CI may be red for unrelated reasons.** master bumped `cheerio` from `1.0.0-rc.12` to `1.2.0`, and `enzyme@3.11.0` requires `cheerio/lib/utils`, which no longer exists. Since `jest.config.js:133` loads `enzyme-to-json/serializer` for every suite, no test in the repo currently runs on a clean install. I validated with a temporary config that drops the enzyme serializer; that config is not part of this PR.
- On npm 12 install scripts are blocked by default, so `canvas` does not build and `jsdom` fails to load. The repo has no `allowScripts` field in `package.json` yet — worth adding separately.
- Not included: an error state instead of the indefinite `<LoadingExplorer />`. That needs new localized strings and is better as its own change.

The sending half of this handshake is fixed separately in `CosmosDB-portal` (PR 2294596): `_fetchMasterKey` / `_getAuthorizationToken` had no rejection handler, so the init message was never posted.

Related: IcM 865096261

### DIFF
--- a/src/Explorer/ContainerCopy/containerCopyStyles.less
+++ b/src/Explorer/ContainerCopy/containerCopyStyles.less
@@ -125,6 +125,11 @@
       opacity: 0.8;
     }
   }
+  .accordionPanel {
+    // Keep confirmation popovers above neighboring accordion headers.
+    position: relative;
+    z-index: 1;
+  }
   .foreground {
     z-index: 10;
     background-color: var(--colorNeutralBackground2);
--- a/src/Metrics/ScenarioMonitor.phaseOrdering.test.ts
+++ b/src/Metrics/ScenarioMonitor.phaseOrdering.test.ts
@@ -1,0 +1,152 @@
+/**
+ * @jest-environment jsdom
+ *
+ * Regression tests for IcM 865096261 — DataExplorerHealthV2 "High Unhealthy Percentage".
+ *
+ * Before the fix, ScenarioMonitor.completePhase() silently no-oped when the phase had not
+ * been started yet. The production call ordering in Explorer.tsx / ResourceTree.tsx always
+ * hits that case:
+ *
+ *   - DatabaseTreeRendered is started at Explorer.tsx:451, but the effect that completes it
+ *     (useMetricPhases.ts:55-59) last fires on the databaseTreeNodes change that happens
+ *     *before* line 451. databaseTreeNodes is useMemo'd (ResourceTree.tsx:54), so the effect
+ *     never re-fires and the phase stayed open until the 10 s timeout.
+ *   - Interactive is completed by a one-shot effect (deps [scenario, enabled]) that can run
+ *     before Explorer.tsx:577 starts the scenario at all.
+ *
+ * Telemetry over 3 days: DatabaseTreeRendered missing in 1624 of 1627 DatabaseLoad timeouts
+ * (99.8%), Interactive missing in 588.
+ *
+ * Line references are against 1f04f0ae.
+ */
+
+import { updateUserContext } from "../UserContext";
+import MetricScenario, { reportMetric } from "./MetricEvents";
+import { ApplicationMetricPhase, CommonMetricPhase } from "./ScenarioConfig";
+import { scenarioMonitor } from "./ScenarioMonitor";
+
+jest.mock("./MetricEvents", () => ({
+  __esModule: true,
+  default: {
+    ApplicationLoad: "ApplicationLoad",
+    DatabaseLoad: "DatabaseLoad",
+  },
+  reportMetric: jest.fn().mockResolvedValue({ ok: true }),
+}));
+
+jest.mock("../ConfigContext", () => ({
+  configContext: {
+    platform: "Portal",
+    PORTAL_BACKEND_ENDPOINT: "https://test.portal.azure.com",
+  },
+  Platform: { Portal: "Portal", Hosted: "Hosted", Emulator: "Emulator", Fabric: "Fabric" },
+}));
+
+/** Stands in for the useDatabaseLoadScenario effect firing on a databaseTreeNodes change. */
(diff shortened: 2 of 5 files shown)


sunghyunkang1111:
### 1. Overview

Reducing false-unhealthy emissions is the right goal, and neither this PR nor the companion portal handshake fix needs to eliminate every transient failure. However, the reviewed patch also permits false-healthy outcomes and can prevent late initialization from recovering. Those are code-correctness issues to address before merge. Deciding which remaining failures warrant an IcM is separate monitoring work.

The concern is that some changes make the observed timeout patterns disappear without establishing that the load finished. A lower unhealthy percentage alone would not demonstrate that the fix worked.

### 2. What needs to be fixed in this PR

1. **Preserve late initialization recovery.** If a valid init message arrives but key retrieval takes longer than 30 seconds, `Promise.race` rejects the caller. The message handler remains active: when the dependency completes, it can still create an Explorer and emit `ConfigurePortal` success, but the rejected caller never resumes its normal `setExplorer` path. Previously, the pending initialization could recover. This is a functional regression path, not residual alert noise. Catch errors inside the async handler and give initialization a consistent lifecycle: either allow recovery through the waiting caller, or cancel/ignore late work and expose an error/retry state.

2. **Do not accept an earlier render as proof that the loaded tree is ready.** The early-completion buffer and phase auto-start can accept a tree render before collections finish. Completing the remaining phases then reports healthy with a zero-duration `DatabaseTreeRendered`, without waiting for a render of the collection-bearing tree. That removes the missing-phase signature by accepting an insufficient observation. Fix the producer ordering and acknowledge the committed tree for the current load's ready revision. Cover stale callbacks, empty accounts and unchanged trees; checking only that a healthy event was emitted is insufficient.

3. **Do not turn unfinished background loads into successes.** A load with no completed phases can time out while hidden and still emit `healthy=true` under this patch. Background timer throttling can justify different alert eligibility, but it does not establish successful initialization. Preserve the actual timeout/completion outcome and visibility context. Let the monitor apply a consistent background policy rather than lowering the failure rate by relabeling incomplete loads.

These are source-level reproductions, and the relevant behavior is unchanged at reviewed head `75d2fc20`. Their production frequency is not established. The first breaks recovery; the other two weaken what a healthy outcome proves.

### 3. What belongs in monitor and alerting changes

- Decide which recorded failures are actionable. The observed 88.4 ms fetch failure can remain an error without necessarily opening an IcM. Use persistence, failure counts and appropriate volume safeguards to reduce incidents from isolated failures.
- Apply background eligibility consistently to successful and failed outcomes. Retain raw failures and deadline misses even when alert policy excludes them.
- Avoid a blanket sample floor: 72.9% of five-minute windows in the combined-scenario Stage log replay had fewer than 20 records. Preserve low-traffic outage and no-data detection.
- Retrieve the actual Stage V2 rule and explain MDM's original **2/19** versus the backend logs' **1/9** before calibrating thresholds. That discrepancy is monitoring follow-up, not a blanket blocker for a scoped, correct code improvement.

